PR #27, opened 2019-03-12
Task/localization

jobala:
## Overview

Adds localization

### Demo

image


src/app/views/query-runner/QueryInput.tsx
```diff
@@ -28,7 +30,9 @@ export const QueryInputControl = ({
       <PrimaryButton
         onClick={() => handleOnRunQuery()}
       >
-        Run Query
+        <FormattedMessage
```
MIchaelMainer: I really like how a localized string resource is referenced like this

MIchaelMainer:
@microsoftgraph/openlocalization - adding localized resources to a new version of Graph Explorer. Any guidance for new projects is appreciated.

src/app/views/query-runner/QueryInput.tsx
jobala: Me too.

PR #28, opened 2019-03-22
Authentication component

thewahome:
## Overview

Allows authentication of users so that they are able to send authenticated requests to the graph 

### Notes

Current implementation is done via popup, redirect for now causes loss of all data including the access token provided on the url after login is successful

## Testing Instructions

* Spin up the app
* Send requests as Megan Bowen
* Hit the sign in button and authenticate
* Send requests using your logged in account


package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@babel/core": "7.2.2",
     "@svgr/webpack": "4.1.0",
+    "@types/hellojs": "^1.16.2",
```
darrelmiller: Why are we pulling in hellojs?  We should be using MSAL.js here.

src/app/services/actions/query-action-creators.ts
```diff
@@ -16,21 +15,24 @@ export function queryResponseError(response: object): IAction {
...
-
-    return fetch(`https://proxy.apisandbox.msdn.microsoft.com/svc?url=${url}`, { headers })
+    let authToken = '{token:https://graph.microsoft.com/}';
+    let graphUrl = `https://proxy.apisandbox.msdn.microsoft.com/svc?url=${url}`;
+    const authenticated = localStorage.getItem('authenticated');
+    if (authenticated) {
+      authToken = JSON.parse(authenticated).token;
+      graphUrl = url;
+    }
+    const headers = { Authorization: `Bearer ${authToken}` };
```
darrelmiller: We are supposed to be using the Graph Javascript SDK for making calls.  It has middleware to do this auth work.  We don't need to deal with tokens manually.

darrelmiller: See this sample here https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/dev/samples/browser/src/main.js#L13  (There is an issue with the latest release where this sample only works on classic Edge. This should be fixed soon.)

package.json
jobala: When pulling in MSAL.js, pull in an exact version. That will help us track changes to our dependencies. The caret before package version means npm will install the latest major change. For example if package version is ^1.16.2 in package.json and there's 1.17.0 availiable, during npm install version 1.17.0 of the package will be installed.

Use `npm i <package-name> --exact` to pull an exact version of a dependency.

src/app/services/actions/query-action-creators.ts
```diff
@@ -16,21 +15,24 @@ export function queryResponseError(response: object): IAction {
...
 export function runQuery(url: string): Function {
-  const headers = { Authorization: 'Bearer {token:https://graph.microsoft.com/}' };
-
   return (dispatch: Function) => {
     const respHeaders: any = {};
-
-    return fetch(`https://proxy.apisandbox.msdn.microsoft.com/svc?url=${url}`, { headers })
+    let authToken = '{token:https://graph.microsoft.com/}';
+    let graphUrl = `https://proxy.apisandbox.msdn.microsoft.com/svc?url=${url}`;
+    const authenticated = localStorage.getItem('authenticated');
```
jobala: What we are getting from local storage is the access token. We should call localStorage.getItem('accessToken')

src/app/services/reducers/auth-reducers.ts
```diff
@@ -0,0 +1,12 @@
+import { IAction } from '../../../types/action';
+import { AUTHENTICATE_USER } from '../constants';
+
+export function authResponse(state = {}, action: IAction) {
```
jobala: Fix the formating to two spaces.

src/app/services/reducers/auth-reducers.ts
```diff
@@ -0,0 +1,12 @@
...
+    switch (action.type) {
+        case AUTHENTICATE_USER:
+        localStorage.setItem('authenticated', JSON.stringify(action.response));
```
jobala: Indent the statements inside case.

Reducers are supposed to be pure functions. A pure function is a function that doesn't have side effects. Example of side effects are network requests, querying a database etc. Accessing local storage here is a side effect. You shouldn't access local storage here, you can do it in an action creator.

src/app/services/reducers/index.ts
```diff
@@ -1,7 +1,8 @@
 import { combineReducers } from 'redux';
 
+import { authResponse } from './auth-reducers';
 import { graphResponse } from './query-runner-reducers';
 
 export default combineReducers({
-  graphResponse,
+  graphResponse, authResponse,
```
jobala: To increase readability add `authResponse` under `graphResponse`. 
```
{
  graphResponse,
  authResponse
}
```

src/app/views/authentication/Authentication.tsx
```diff
@@ -0,0 +1,142 @@
...
+          <span className='user-email'>{authenticated.user.emailAddress}</span>
+        </div>
+      </div>
+    );
+  }
+}
+
+function mapDispatchToProps(dispatch: Dispatch): object {
+  return {
+    actions: bindActionCreators(authActionCreators, dispatch),
```
jobala: You don't have to call `bindActionCreators` twice. You can do the following
```
bindActionCreators({ queryActions, authActions }, dispatch)
```

src/app/views/authentication/Authentication.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const authenticated = localStorage.getItem('authenticated');
+    if (authenticated && this.props.actions) {
+      this.props.actions.authenticateUser(JSON.parse(authenticated));
+      this.setState({
+        authenticated: JSON.parse(authenticated),
+      });
+    }
+  }
+
+  public signIn = () => {
```
jobala: You should use abstract the implementation of authentication. This component should depend on an abstraction(interface) and not an implementation(hello.js/msal.js)

src/app/views/authentication/Authentication.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { bindActionCreators, Dispatch } from 'redux';
+
+import { IAuthenticationProps, IAuthenticationState } from '../../../types/authentication';
+import * as authActionCreators from '../../services/actions/auth-action-creators';
+import * as queryActionCreators from '../../services/actions/query-action-creators';
+import { ADMIN_AUTH_URL, AUTH_URL, DEFAULT_USER_SCOPES , TOKEN_URL, USER_INFO_URL } from '../../services/constants';
+import './authentication.scss';
+
+export class Authentication extends Component<IAuthenticationProps,  IAuthenticationState> {
+  public state = {
```
jobala: I prefer using a constructor because it aligns with the OOP model as compared to a class variable.

src/app/views/authentication/Authentication.tsx
```diff
@@ -0,0 +1,142 @@
+import hello from 'hellojs';
+import { PrimaryButton } from 'office-ui-fabric-react';
+import React, { Component } from 'react';
+
```
jobala: Remove this newline

src/app/views/authentication/Authentication.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { IAuthenticationProps, IAuthenticationState } from '../../../types/authentication';
+import * as authActionCreators from '../../services/actions/auth-action-creators';
+import * as queryActionCreators from '../../services/actions/query-action-creators';
+import { ADMIN_AUTH_URL, AUTH_URL, DEFAULT_USER_SCOPES , TOKEN_URL, USER_INFO_URL } from '../../services/constants';
+import './authentication.scss';
+
+export class Authentication extends Component<IAuthenticationProps,  IAuthenticationState> {
+  public state = {
+    authenticated: {
```
jobala: This information is already available in the store, you don't need this local state.

src/app/views/authentication/Authentication.tsx
```diff
@@ -0,0 +1,142 @@
...
+      user: {
+        displayName: '',
+        emailAddress: '',
+      },
+      token: '',
+    },
+  };
+
+  public componentDidMount = () => {
+    const authenticated = localStorage.getItem('authenticated');
```
jobala: You should get this information from the store instead of the local storage.

src/app/views/authentication/authentication.scss
```diff
@@ -0,0 +1,20 @@
...
+    margin-bottom: $gutter;
+  }
+  
+  .authentication-details { 
+    margin-bottom: $gutter;
+    justify-content: space-between;
+  }
+
+  .user-email {
+    font-size: .8em;
```
jobala: There's a font size variable in `_variables.scss`

jobala: Using variables makes the styling uniform.

jobala:
The sign in button still shows after signing in.

src/app/services/actions/query-action-creators.ts
thewahome: access token is in the authenticated object... as per the design spec

src/app/services/reducers/auth-reducers.ts
thewahome: makes sense. adding a new action creator

src/app/views/authentication/Authentication.tsx
thewahome: the store looses data when the page reloads. Thus on successful authentication, the details are saved in the store and the status also saved in the local storage. The componentDidMount checks whether the user is authenticated and passes the localstorage data to the state so that the user data is still known

src/app/views/authentication/Authentication.tsx
thewahome: awesome

src/app/views/authentication/Authentication.tsx
jobala: You should hydrate the store with this information whenever the application loads.

src/app/services/actions/query-action-creators.ts
jobala: What does the authenticated object contain? I suppose an authenticated user information? I think we should rename it to authenticatedUser.

src/app/views/authentication/Authentication.tsx
thewahome: done

PR #29, opened 2019-03-25
Sign in action

thewahome:
## Overview

Moves the sign-in action and profile to their own components
Adds a reusable button that accepts the progress status and displays a spinner to show that there is action happening


darrelmiller:
Why are we using hello.js?  Why are we not using the Javascript Graph SDK?

jobala:
Replace the sign in button with a  loading spinner when signing in.

src/app/views/authentication/Authentication.tsx
```diff
@@ -107,17 +138,16 @@ export class Authentication extends Component<IAuthenticationProps,  IAuthentica
   }
 
   public render() {
-    const { authenticated } = this.state;
+    const { authenticated, loading } = this.state;
+    let buttonText = 'Sign In';
+    if (authenticated.status) {
+      buttonText = 'Sign Out';
+    }
```
jobala: You can do the following

```
const buttonLabel = authenticated.status ? 'Sign In' : 'Sign Out'
```

PR #31, opened 2019-03-26
Adds support for iframe message passing

jobala:
## Overview

Adds support to receive message from a host document.

### Demo

N/A

### Notes

`WindowMessenger` uses iframe message passing to send data to an iframe. This PR listens for a message event and extracts the data.
The data should be an object with the following type 
```
{
    sampleUrl: string;
    sampleVerb: string;
    sampleBody: object;
    sampleHeaders: object;
{
```
## Testing Instructions

Embed GE in an iframe and send a message to the iframe.

### DIFF
--- a/src/app/views/query-runner/QueryInput.tsx
+++ b/src/app/views/query-runner/QueryInput.tsx
@@ -10,7 +10,7 @@ export const QueryInputControl = ({
   handleOnUrlChange,
   httpMethods,
   selectedVerb,
-  sampleURL,
+  sampleUrl,
 }: IQueryInputControl) => {
 
   return (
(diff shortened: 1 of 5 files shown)


Kachulio1:
@jobala where is this `iframe message`?  I'm new here want to understand the code base better.

jobala:
@Kachulio1 thank you for your questions.

> Why are we ignoring Typescript here?

We are ignoring a Typescript linting error, the other alternative is to specify the type of `event.origin` as `any`


> @jobala where is this iframe message? I'm new here want to understand the code base better.
The iframe message will be passed from the host document. This application only receives the data passed and updates the state.

darrelmiller:
I approved this because we are going to need to accept a message.  However, that message will not be passed in to us as separate url, body, method parameters.  We are going to need to parse that ourselves.  I will create a new issue for that work.

PR #32, opened 2019-03-27
adds microsoft graph sdk authentication

thewahome:
## Overview

Changes implementation from hellojs to microsoft graph javascript sdk


src/app/services/reducers/auth-reducers.ts
```diff
@@ -4,8 +4,7 @@ import { AUTHENTICATE_USER } from '../constants';
 export function authResponse(state = {}, action: IAction) {
```
MIchaelMainer: Specify the return type

src/app/views/authentication/AuthService.ts
```diff
@@ -0,0 +1,23 @@
+import { MSALAuthenticationProvider } from '@microsoft/microsoft-graph-client/lib/src/MSALAuthenticationProvider';
+import { DEFAULT_USER_SCOPES } from '../../services/constants';
+
+const clientId = 'cb2d7367-7429-41c6-ab18-6ecb336139a6';
+const graphScopes = [DEFAULT_USER_SCOPES];
+const options = { redirectUri: window.location };
+const authProvider = new MSALAuthenticationProvider(clientId, graphScopes, options);
+
+export async function getAccessToken() {
```
MIchaelMainer: Specify the return type, including void. Please apply to all functions.

src/app/views/authentication/Authentication.tsx
```diff
@@ -1,182 +1,152 @@
...
-  public state = {
-    authenticated: {
-      status: false,
-      user: {
-        displayName: '',
-        emailAddress: '',
-        profileImageUrl: '',
+  constructor(props: any) {
+    super(props);
+    this.state = {
```
MIchaelMainer: `state` property definition, let's explicitly define it as IAuthenticationState

src/app/views/authentication/Authentication.tsx
```diff
@@ -1,182 +1,152 @@
...
 export class Authentication extends Component<IAuthenticationProps,  IAuthenticationState> {
-  public state = {
-    authenticated: {
-      status: false,
-      user: {
-        displayName: '',
-        emailAddress: '',
-        profileImageUrl: '',
+  constructor(props: any) {
+    super(props);
```
MIchaelMainer: What is `props`?

src/app/views/authentication/Authentication.tsx
```diff
@@ -1,182 +1,152 @@
...
-                });
-              }
-            }
-          } catch (e) {
-            // tslint:disable-next-line:no-console
-            console.log(e);
-          }
-        }
-      });
+  private async getUserInfo(queryActions: any) {
```
muthurathinam: Can we move these funtion (getUserInfo and getImageUrl) to some other file.. its doesn't look like an appropriate fit inside Authentication.tsx file or do we need to rename this file ?

src/app/views/authentication/Authentication.tsx
```diff
@@ -1,182 +1,152 @@
...
+    const jsonUserInfo = userInfo.response.body;
+    return jsonUserInfo;
+  }
+
+  private async getImageUrl(queryActions: any) {
+    const userPicture = (queryActions) ? await queryActions.runQuery({
+      sampleURL: USER_PICTURE_URL,
+    }) : null;
+    const buffer = await userPicture.response.body.arrayBuffer();
+    const blob = new Blob([buffer], { type: 'image/jpeg' });
```
muthurathinam: Have we tested that type "image/jpeg" would work for all types of images ?

src/app/views/authentication/Authentication.tsx
```diff
@@ -1,182 +1,152 @@
...
+    const imageUrl = URL.createObjectURL(blob);
+    return imageUrl;
   }
 
   public render() {
-    const { authenticated, loading } = this.state;
-    const buttonLabel = authenticated.status ? 'sign out' : 'sign in';
+    const { authenticatedUser, loading } = this.state;
+    const buttonLabel = authenticatedUser.status ? 'sign out' : 'sign in';
```
muthurathinam: Are we not concerned about i18n as of now ? (sign out and sign in)

thewahome: It's factored in in the component level. the words are passed in to the component and the component formats the strings in the required language

src/app/views/authentication/Authentication.tsx
```diff
@@ -1,182 +1,152 @@
...
   public componentDidMount = () => {
-    const authenticated = localStorage.getItem('authenticated');
-    if (authenticated && this.props.actions && JSON.parse(authenticated).status) {
-      this.props.actions.authenticateUser(JSON.parse(authenticated));
+    const authenticatedUser = localStorage.getItem('authenticatedUser');
+    if (authenticatedUser && this.props.actions && JSON.parse(authenticatedUser).status) {
+      this.props.actions.authenticateUser(JSON.parse(authenticatedUser));
       this.setState({
-        authenticated: JSON.parse(authenticated),
+        authenticatedUser: JSON.parse(authenticatedUser),
```
jobala: JSON.parse is being called multiple times in this function. If you make the return value a variable you will only have to call it ones.

```
const authUser = JSON.parse(authenticatedUser)
```

jobala: JSON.parse is being called multiple times in this function. If you make the return value a variable you will only have to call it ones.

```
const authUser = JSON.parse(authenticatedUser)
```

src/app/views/authentication/Authentication.tsx
thewahome: We carried this over from classic. As there were no noted problems with the image before, its highly unlikely for there to be a problem now.

src/app/services/actions/query-action-creators.ts
```diff
@@ -20,11 +20,11 @@ export function runQuery(query: IQuery): Function {
   let authToken = '{token:https://graph.microsoft.com/}';
   let graphUrl = `https://proxy.apisandbox.msdn.microsoft.com/svc?url=${query.sampleURL}`;
   const respHeaders: any = {};
-  const authenticated = localStorage.getItem('authenticated');
+  const authenticatedUser = localStorage.getItem('authenticatedUser');
```
jobala: We are querying local storage everytime we make a request and in other places. This is not efficient. We can reduce the number of times we are fetching data from the local storage by loading authenticatedUser into the store when the application loads.

You can do this by passing `authenticatedUser` as the initial state when creating the store in `index.tsx`

src/app/services/actions/query-action-creators.ts
```diff
@@ -20,11 +20,11 @@ export function runQuery(query: IQuery): Function {
...
 
   return (dispatch: Function) => {
-    if (authenticated && JSON.parse(authenticated).token) {
-      authToken = JSON.parse(authenticated).token;
+    if (authenticatedUser && JSON.parse(authenticatedUser).token) {
+      authToken = JSON.parse(authenticatedUser).token;
```
jobala: JSON.parse is called twice in this function. You can make it's return value a variable to avoid repeating yourself.

src/app/views/authentication/Authentication.tsx
```diff
@@ -1,182 +1,134 @@
...
+    };
+  }
 
   public componentDidMount = () => {
-    const authenticated = localStorage.getItem('authenticated');
-    if (authenticated && this.props.actions && JSON.parse(authenticated).status) {
-      this.props.actions.authenticateUser(JSON.parse(authenticated));
+    const authenticatedUser = localStorage.getItem('authenticatedUser');
+    const authUser = (authenticatedUser) ? JSON.parse(authenticatedUser) : null;
+    if (authenticatedUser && this.props.actions && authUser.status) {
```
jobala: Add a new line between the variable declaration and the if statements

jobala:
Rename AuthService to auth-util. Because it contains authentication utility methods.

src/app/services/actions/query-action-creators.ts
thewahome: I used learnings from this article.  Local storage actually assists us and is cleared after use

https://hackernoon.com/how-to-take-advantage-of-local-storage-in-your-react-projects-a895f2b2d3f2

src/app/views/authentication/Authentication.tsx
thewahome: lines added. However, aesthetics are enforced by the linter. We should consider modifying the linter as it disapproves the lines and forces me to add two lines, and then remove one for the linter

src/app/services/actions/query-action-creators.ts
jobala: What do you mean it is cleared after use? 

We are making calls to local storage in multiple places. We can reduce this by loading the data from local storage to the store when the page loads. This can be a maintanance nightmare especially when we decided to change the key with which we are storing the values.

It's also a perfomance overhead.

PR #33, opened 2019-03-29
Decodes data encoded as a base64 token

jobala:
## Overview

Passes transient state to Graph Explorer as a base64 encoded token.

### Demo

N/A

### Notes

N/A

## Testing Instructions

* Launch Graph Explorer
* Copy paste the url below and reload
```
http://localhost:3000/?query=eyJzYW1wbGVVcmwiOiJodHRwczovL2dyYXBoLm1pY3Jvc29mdC5jb20vdjEuMC9tZS9tZXNzYWdlcyIsInNhbXBsZVZlcmIiOiJwb3N0Iiwic2FtcGxlQm9keSI6eyJuYW1lIjoiU29tZW9uZSBJbXBvcnRhbnQifSwic2FtcGxlSGVhZGVycyI6eyJjb250ZW50LXR5cGUiOiJhcHBsaWNhdGlvbi9qc29uIn19

The base64 token above is generated from 
{
     'sampleUrl': 'https://graph.microsoft.com/v1.0/me/messages',
     'sampleVerb': 'post',
     'sampleBody': {
         'name': 'Someone Important',
     },
     'sampleHeaders': {
         'content-type': 'application/json'
     }
 }
```
* Observe that the sample is `https://graph.microsoft.com/v1.0/me/messages`


jobala:
@MIchaelMainer 

Graph Explore only decodes the string. The encoding should be done by ApiDoctor. @darrelmiller is aware of this limitation.

darrelmiller:
@jobala When we implement sharing a query then GE is going to need to encode as well as decode.  It's not a priority for the moment though.  I'm not too concerned about the browser limit.  There won't likely be many scenarios where we are going to hit it.

PR #36, opened 2019-04-08
enable request options

thewahome:
## Overview

Allows the user to query the graph using the different request options

### Notes

There are a few things that are not covered in this PR.
* User feedback when the Run Query button is clicked and the request has been created and awaiting the response
* Clearing the request body automatically
* Notifying the user when the request has successfully/unsuccessfully completed


src/app/services/actions/query-action-creators.ts
```diff
@@ -31,9 +32,20 @@ export function runQuery(query: IQuery): Function {
...
+    const headers = {
+      'Authorization': `Bearer ${authToken}`,
+      'Content-Type': 'application/json',
+    };
 
-    return fetch(graphUrl, { headers })
+    let options: IRequestOptions = {};
+    options = { method: query.selectedVerb, headers};
+
+    if (query.sampleBody && query.selectedVerb !== 'GET') {
```
jobala: To improve readability you can make the expressions in the condition statement variables.
```
const hasBody = !!query.sampleBody // !! forces it to be true or false
const isGET = query.selectedVerb

if (hasBody && !isGET) {
  ...
}
```

src/app/services/actions/query-action-creators.ts
```diff
@@ -31,9 +32,20 @@ export function runQuery(query: IQuery): Function {
...
+      const body = JSON.stringify(query.sampleBody);
+      options.body = body;
```
jobala: Mutating objects like this can be a source of hard to debug bugs. You can instead create a new object and assign it to options.

options = Object.assign({}, options, { body: body })

src/app/services/actions/query-action-creators.ts
jobala: Should this work for `DELETE` requests

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -22,6 +22,10 @@ export class QueryRunner extends Component<IQueryRunnerProps, IQueryRunnerState>
       ],
       selectedVerb: 'GET',
       sampleURL: 'https://graph.microsoft.com/v1.0/me/',
+      sampleBody: undefined,
+      headers: [{ name: '', value: '' }],
+      headerName: '',
```
jobala: The headers state can be derived from headerName and headerValue. If part of the state can be derived it is not necessary to explicitly define it.

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -37,14 +41,74 @@ export class QueryRunner extends Component<IQueryRunnerProps, IQueryRunnerState>
...
+    if (value) {
+      this.setState({
+        headerValue: value,
+      });
+    }
+  };
+
+  private handleOnHeaderDelete = (headerIndex: any) => {
+    const { headers } = this.state;
+    const headersToDelete = [...headers];
```
jobala: Add a new line after variable declarations.

Here we are destructuring headers then inititializing `headersToDelete` with `headers`. The same can be achieved succinctly by doing the following

```
  const { headers: headersToDelete } = this.state

```
This will destructure and inititialize

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -37,14 +41,74 @@ export class QueryRunner extends Component<IQueryRunnerProps, IQueryRunnerState>
...
+    headersToDelete.splice(headerIndex, 1);
+    this.setState({
+      headers: headersToDelete,
+    });
+    const listOfHeaders = headers;
```
jobala: Add a new line after variable declaration

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -37,14 +41,74 @@ export class QueryRunner extends Component<IQueryRunnerProps, IQueryRunnerState>
...
+      this.setState({
+        headers: newHeaders,
+        headerName: '',
+        headerValue: '',
+      });
+    }
+  };
+
+  public getLastHeader() {
+    const headersLength = this.state.headers.length;
```
jobala: Using `pop()` will make the code cleaner

```
return headers.pop()
```

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -37,14 +41,74 @@ export class QueryRunner extends Component<IQueryRunnerProps, IQueryRunnerState>
...
+    return this.state.headers[headersLength - 1];
+    }
+
   private handleOnRunQuery = () => {
-    const { sampleURL } = this.state;
+    const { sampleURL, selectedVerb, sampleBody } = this.state;
```
jobala: How are we handling queries that have headers?

src/app/views/query-runner/request/Request.tsx
```diff
@@ -1,91 +1,38 @@
...
+              handleOnHeaderNameChange={(event: any, name: any) => handleOnHeaderNameChange(name)}
+              handleOnHeaderValueChange={(event: any, value: any) => handleOnHeaderValueChange(value)}
+              handleOnHeaderValueBlur={(event: any, header: any) => handleOnHeaderValueBlur(header)}
+              headers={headers}
+            />
+          </PivotItem>
+        </Pivot>
+      </div>
+    );
+  };
```
jobala: Good stuff on making this a function.

jobala:
@thewahome  regarding the things that are not covered in this PR, is there anything I can do to unblock you?

src/app/views/common/monaco/Monaco.tsx
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
-import MonacoEditor from 'react-monaco-editor';
+import MonacoEditor, { ChangeHandler } from 'react-monaco-editor';
 
 import './monaco.scss';
 
 interface IMonaco {
   body: object|undefined;
+  onChange?: ChangeHandler | undefined;
```
Kachulio1: I don't think this should be optional, no?

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -22,6 +22,10 @@ export class QueryRunner extends Component<IQueryRunnerProps, IQueryRunnerState>
       ],
       selectedVerb: 'GET',
       sampleURL: 'https://graph.microsoft.com/v1.0/me/',
+      sampleBody: undefined,
```
Kachulio1: can we use `null` or `''` instead of `undefined`  so that it can be easy to debug.

src/app/views/common/monaco/Monaco.tsx
jobala: It is optional because there are instances where we only want to hydrate the editor and don't expect a user to modify the content.

src/app/views/query-runner/QueryRunner.tsx
jobala: I prefer an empty string.

src/app/views/common/monaco/Monaco.tsx
Kachulio1: gotcha

PR #37, opened 2019-04-14
fix infinite loading when popup window is closed

Kachulio1:
## Overview

When the sign in with Microsoft button is clicked and a user closes the signup/sign in popup window the button deosn't go back to the initial state.
### Demo
Screen Shot 2019-04-15 at 01 02 58
You get this error in the console.
Screen Shot 2019-04-15 at 01 02 35

### DIFF
--- a/src/app/views/authentication/Authentication.tsx
+++ b/src/app/views/authentication/Authentication.tsx
@@ -41,10 +41,10 @@ export class Authentication extends Component<IAuthenticationProps,  IAuthentica
         accessToken = await getAccessToken();
       } catch (e) {
         this.setState({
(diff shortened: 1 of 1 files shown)


darrelmiller:
Hey @Kachulio1 , we appreciate the fix.  We need to actually change the way the sign-in works so that it is no longer a popup as we need to host this in an iFrame.  I'm not sure how your fix will be impacted by this change /cc @thewahome

Kachulio1:
@thewahome  should I close this PR?

thewahome:
Yeah. @Kachulio1, you can close it

PR #44, opened 2019-04-18
Feat/theme

jobala:
## Overview

Adds theming support for Graph Explorer

#### Light
light theme

#### Dark
dark theme

### Notes
This PR introduces a CSS-in-JS approach. From this point on, all styling should use this approach as the standard technique for styling. If you touch a scss file migrate it to CSS-in-JS.

Theme switching is manual right now, it can be automated once we find out how to detect the host document theme from the docs team.
## Testing Instructions

* Launch Graph Explorer
* Navigate to `src/index.tsx`
* In the `loadTheme` function change the argument to either dark or light
* Observe that GE's theme changes


package.json
```diff
@@ -5,6 +5,8 @@
   "dependencies": {
     "@babel/core": "7.2.2",
     "@microsoft/microsoft-graph-client": "^1.5.2",
+    "@svgr/webpack": "4.1.0",
+    "@uifabric/styling": "^6.46.0",
```
thewahome: should we call an exact package here?

jobala: Yes we should. Thanks for pointing this out.

PR #57, opened 2019-05-10
disables run query button when query is running

thewahome:
## Overview

disables run query button when query is running

### Demo

### Notes

## Testing Instructions

* Click run query button
* It should disable and show a loading spinner until the query is complete


src/app/services/actions/query-action-creator-util.ts
```diff
@@ -20,7 +20,15 @@ export function queryResponse(response: object): IAction {
...
 
+export function queryRunningStatus(response: object): IAction {
+  return {
+    type: QUERY_GRAPH_RUNNING,
+    response,
+  };
+}
+
 export function anonymousRequest(dispatch: Function, query: IQuery) {
+
```
jobala: Remove new line.

src/app/services/actions/query-action-creator-util.ts
```diff
@@ -30,28 +38,35 @@ export function anonymousRequest(dispatch: Function, query: IQuery) {
...
       throw new Error('The request was not completed');
     })
     .then((json) =>
-
       dispatch(
         queryResponse({
           body: json,
           headers: respHeaders,
         }),
+        queryRunningStatus({ isLoadingData: false })
```
jobala: By the time this `queryRunningStatus` is dispatched the value of `isLoadingData` is false because of the dispatch in `line 47`. It is safe to not dispatch `queryRunningStatus` here.

src/app/services/actions/query-action-creator-util.ts
```diff
@@ -30,28 +38,35 @@ export function anonymousRequest(dispatch: Function, query: IQuery) {
...
     )
-    .catch((error) => dispatch(queryResponseError(error)));
+    .catch((error) => dispatch(
+      queryResponseError(error),
+      queryRunningStatus({ isLoadingData: false }))
+    );
 }
 
 export function authenticatedRequest(dispatch: Function, query: IQuery) {
+  dispatch(queryRunningStatus({ isLoadingData: true }));
```
jobala: This dispatch should happen inside the `makeRequest` function. Since a dispatch with `isLoadingData` to true should have a complementing dispatch with `isLoadingData` false. Having both complementary dispatches inside the `makeRequest` function improves readability and also prevents opportunities for errors.

Move this to line 108

src/app/services/actions/query-action-creator-util.ts
thewahome: At what point is it safer?

src/app/services/reducers/query-runner-reducers.ts
```diff
@@ -1,12 +1,23 @@
...
+import { QUERY_GRAPH_ERROR, QUERY_GRAPH_RUNNING, QUERY_GRAPH_SUCCESS } from '../redux-constants';
 
 export function graphResponse(state = {}, action: IAction): object {
   switch (action.type) {
     case QUERY_GRAPH_SUCCESS:
       if (typeof action.response !== 'string') {
         return action.response;
       }
+    case QUERY_GRAPH_ERROR:
+      if (typeof action.response !== 'string') {
```
Kachulio1: whay are we checking if payload  is a not of type string

jobala: Payloads can either be string or objects. 

The other alternative is for the return type to be `object|string` instead of  `object`

src/app/services/actions/query-action-creator-util.ts
jobala: I don't understand the question

jobala: It is safe to remove line `59`

src/app/services/reducers/query-runner-reducers.ts
Kachulio1: @jobala but the action creators have response type as `object`  `queryResponseError(response: object)`, `queryRunningStatus(response: object)` so we can be sure  if the case is `QUERY_GRAPH_ERROR` then we have an object, but you are saying that `response.body` `queryResponseError(response.body)` might be a string or object.

src/app/services/reducers/query-runner-reducers.ts
```diff
@@ -1,12 +1,23 @@
...
+        return action.response;
+      }
+    case QUERY_GRAPH_RUNNING:
+    if (action.response) {
+        return {
+          ...state,
+          isLoadingData: !!action.response
```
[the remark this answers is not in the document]
thewahome: running a bang twice determines the opposite of a value, and then returns the opposite of that. Hence 
`!!true // Evaluates to true.
!!false // Evaluates to false.`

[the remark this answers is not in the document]
Kachulio1: yeah, that's if you have something like `!!age` which is a Number and you want to convert it to a` boolean`  but here you are explicitly passing `queryRunningStatus({ isLoadingData: false })` or `true` which is already a `boolean`.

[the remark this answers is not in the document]
thewahome: Sometimes it returns an undefined. It's more of a defensive move

src/app/services/reducers/query-runner-reducers.ts
jobala: Action creators are of the type `IAction`. If you look at `src/types/action.ts`, you will see that the `response` property is either an object or string.

src/app/services/reducers/query-runner-reducers.ts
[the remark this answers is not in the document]
Kachulio1: I think it should be `action.response.isLoadingData`

src/index.tsx
```diff
@@ -25,8 +25,10 @@ initializeIcons();
 
 const appState = store({
   authToken: '',
+  isLoadingData: false
 });
 
+
```
jobala: Unecessary new line

PR #65, opened 2019-05-27
enable sending of request headers with query

thewahome:
## Overview

Enables the sending of request headers with the queries

### Demo

### Notes

Does not remove the header from the display but removes it from the 'behind the scenes' array

## Testing Instructions

* Head over to the request headers tab
* Put in the header name in the key field and the header value in the key field
* hit run query
* Investigate using the developer console the request headers that you have sent out. The ones you added will appear


src/app/services/reducers/request-headers-reducers.ts
```diff
@@ -0,0 +1,26 @@
...
+        if (element !== header) {
+          headers.push(element);
+        }
+      });
+
+      if (headers.length === 0) {
+        headers.push({name: '', value: ''});
+      }
+
+      return headers;
```
jobala: You can make this more succinct with `Array.filter`.

```
state.filter(header => header !== action.response)
```

src/app/services/redux-constants.ts
```diff
@@ -4,3 +4,5 @@ export const CLEAR_QUERY_ERROR = 'CLEAR_QUERY_ERROR';
 export const GET_AUTH_TOKEN_SUCCESS = 'GET_AUTH_TOKEN_SUCCESS';
 export const LOGOUT_SUCCESS = 'LOGOUT_SUCCESS';
 export const QUERY_GRAPH_RUNNING = 'QUERY_GRAPH_RUNNING';
+export const HEADER_ADD = 'HEADER_ADD';
```
jobala: There are two groups of constants, SUCCESS and ERROR

Rename these to either HEADER_ADD_SUCCESS/ERROR for consistency.

src/app/views/query-runner/request/HeadersList.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    />;
+
+                default:
+                    return <span className='field-content'>{fieldContent}</span>;
+            }
+        }
+    };
+
+    const columns = [
+        { key: 'key', name: 'Key', fieldName: 'name', minWidth: 100, maxWidth: 200, isResizable: true },
```
jobala: Why are the columns resizable?

src/app/views/query-runner/request/RequestHeaders.tsx
```diff
@@ -1,55 +1,120 @@
...
+import { PrimaryButton, TextField } from 'office-ui-fabric-react';
+import React, { Component } from 'react';
+import { FormattedMessage, injectIntl } from 'react-intl';
+import { connect } from 'react-redux';
+import { bindActionCreators, Dispatch } from 'redux';
+import { IRequestHeadersProps } from '../../../../types/request';
+import * as headersActionCreators from '../../../services/actions/request-headers-action-creators';
+import HeadersList from './HeadersList';
+
+class RequestHeaders extends Component<IRequestHeadersProps, any> {
```
jobala: Formatting in this file is off, for consistency use two spaces.

src/types/request.ts
```diff
@@ -1,25 +1,29 @@
-export interface IRequestHeadersControl {
+export interface IHeadersListControl {
```
jobala: Formatting in this file is off as well.

PR #82, opened 2019-07-04
increase key and value width

### LINKED ISSUE #70: Headers Key and Value are not evenly placed in their container.
The Key and Value should each occupy a width of 45% with respect to the container and the delete button 10%.

headers ui enhancement

Kachulio1:
### Demo
Screen Shot 2019-07-04 at 23 15 03

Fix #70 

### Question 
I have increased the minWidth, maxWidth for the columns to get the current results ☝️ 
But I don't think that will solve the issue 100% 

My second approach was to add a class name as seen below 👇 
Screen Shot 2019-07-04 at 23 13 47

and set the width to 45% like what the issue #70 description suggested 
Screen Shot 2019-07-04 at 23 28 39 I know, css in js 😨 

this is what I get 👇 
Screen Shot 2019-07-04 at 23 10 19

because the parent div doesn't have a width property. 
could someone point me in the right direction?

### DIFF
--- a/src/app/views/query-runner/request/HeadersList.tsx
+++ b/src/app/views/query-runner/request/HeadersList.tsx
@@ -26,9 +26,9 @@ const renderItemColumn = (item: any, index: number | undefined, column: IColumn 
 };
 
 const columns = [
-  { key: 'key', name: 'Key', fieldName: 'name', minWidth: 100, maxWidth: 200 },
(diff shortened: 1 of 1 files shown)


Kachulio1:
@jobala ☝️

jobala:
@Kachulio1  We can work with the first approach for now.

Kachulio1:
👌

Kachulio1:
@jobala It's ready then

PR #127, opened 2019-07-17
Dockerize graph explorer 

Kachulio1:
## Overview

Docker will help us to avoid unexpected errors in different environments.

### Demo

run docker-compose up --build

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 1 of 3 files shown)


Kachulio1:
@darrelmiller Yes it would be useful, the reason for creating the PR was I was getting an error specific to my environment. 
Screen Shot 2019-07-13 at 18 15 14

The reason why I closed it, is that I didn't create an issue to ask the team if it was okay to dockerize GE. I did that so that I could create the issue first but then I got tied up with work and forgot.

Kachulio1:
Should I reopen the PR?

PR #134, opened 2019-08-05
Task/simpler build files

### LINKED ISSUE #100: Package Graph Explorer as a JS component for deployment

jobala:
## Overview

Modifies Webpack to produce a simplified build output

### Demo

N/A

### Notes
This PR produces single .js and .css files instead of chunked files. While the single file output is better for hosting in the portal it degrades perfomance of Graph Explorer v2.

## Testing Instructions

N/A

Fixes #100


MIchaelMainer:
@jobala what are the *.worker files? Are those all compiled into the files under /static?

jobala:
Those are web/service workers. No they are not compiled into the files under /static.

config/webpack.config.js
```diff
@@ -540,8 +460,7 @@ module.exports = function(webpackEnv) {
         new MiniCssExtractPlugin({
           // Options similar to the same options in webpackOptions.output
           // both options are optional
-          filename: 'static/css/[name].[contenthash:8].css',
```
cwahome: if we stop giving hashes to the css name, how will the browser know not to pick the css file in the cache but pick the one  in the server?

jobala: The browser will have to hit the server for the .css & .js files. I mentioned under `notes` that we are optimizing for ease of integration into portal.

PR #145, opened 2019-08-21
adds the history component

cwahome:
## Overview

Adds a history component to the sidebar

### Notes

One particular test is failing, regarding the run query. May need help changing it to run with the current architecture and pass.

image


src/index.tsx
```diff
@@ -80,6 +82,14 @@ if (hostDocumentLocale) {
   appState.dispatch(setGraphExplorerMode(Mode.TryIt));
 }
 
+readData().then((data: any) => {
+  if (data.length > 0) {
```
jobala: This will break when data is null.

src/store/cache.ts
```diff
@@ -0,0 +1,16 @@
+import localforage from 'localforage';
+
+const key = 'history';
+
+export function writeData  (data: object) {
```
jobala: Specify the type of data

src/store/cache.ts
```diff
@@ -0,0 +1,16 @@
...
+  localforage.setItem(key, data);
+}
+
+export async function readData  () {
```
jobala: Specify the return type

jobala:
#### Bugs
1. GE breaks if there are no initial history items

#### Questions
2. Why are we using `localForage` instead of the IndexedDb native api?

#### Suggestions
1. Add screenshots to the PR description

src/app/services/reducers/index.ts
```diff
@@ -6,13 +6,15 @@ import { isLoadingData } from './query-loading-reducers';
 import { queryRunnerError } from './query-runner-error';
 import { graphResponse } from './query-runner-reducers';
 import { headersAdded } from './request-headers-reducers';
+import { requestHistory } from './request-history-reducers';
```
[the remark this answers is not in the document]
cwahome: makes sense

src/app/services/reducers/request-history-reducers.ts
```diff
@@ -0,0 +1,20 @@
...
+  switch (action.type) {
+    case ADD_HISTORY_ITEM_SUCCESS:
+      let history = [...state, action.response];
+      history = history.reduce((current, compare) => {
+        return current.findIndex((historyItem: IHistoryItem) => {
+          return historyItem.runTime === compare.runTime;
+        }) < 0 ? [...current, compare] : current;
+      }, []);
+
+      writeData(history);
```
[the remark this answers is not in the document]
cwahome: Moving this to the action creator instead

cwahome:
> #### Questions
> . Why are we using `localForage` instead of the IndexedDb native api?
> 

I used this because it simplifies and abstracts the native api in a very easy to understand way like localStorage.

jobala:
>I used this because it simplifies and abstracts the native api in a very easy to understand way like localStorage.

But it adds a bunch of other dependencies that we don't use. I would prefer we use the native api. Maybe @MIchaelMainer  can advise here.

src/app/services/actions/query-action-creators.ts
```diff
@@ -1,28 +1,80 @@
...
+    body: query.sampleBody,
+    responseHeaders: respHeaders,
+    createdAt,
+    status,
+    statusText,
+    result,
+    duration,
+    har: ''
   };
+
```
jobala: Should we create the HAR payload if we are not exporting/downloading history items?

cwahome: The HAR payload was created but the export function did not work. It was removed from the UI to make it work as another work item.

cwahome: > I don't understand why click on view in the history component dispatches QUERY_GRAPH_SUCCESS and it's not obvious from the code. Please respond on how/where this is happening.

> 

View dispatches an action creator called viewHistoryItem which dispatches the QUERY_GRAPH_SUCCESS action.

jobala: I see, this is misleading and has an element of surprise to it. The action type `QUERY_GRAPH_SUCCESS` should only be dispatched when making a network request, which is not the case for viewing history items.

I suggest that you dispatch a `VIEW_HISTORY_ITEM_SUCCESS` action type and handle the type in the graphResponse reducer.

PR #173, opened 2019-09-10
Disables the http version selector for TryIt

### LINKED ISSUE #164: Hide all the verbs except GET for try it
Because we do not support anything but GET in Try It. Can we disable the drop down or just simply have it as a label rather than a drop down for now?

jobala:
## Overview

Disables the http version selector for customers using Graph Explorer in TryIt mode

### Demo

Screenshot 2019-09-10 at 11 11 06 AM

### Notes

At the moment, customers using graph explorer in TryIt mode only have access to a limited feature set. One of these limitations is that they don't need to change http verbs, therefore, this PR disables the selector control.

## Testing Instructions

* Launch Graph Explorer
* Add the query parameter `?locale=en-us` to activate the TryIt mode
* Observe that the http verb selector is disabled.

Fixes #164

### DIFF
--- a/src/app/views/query-runner/QueryInput.tsx
+++ b/src/app/views/query-runner/QueryInput.tsx
@@ -1,7 +1,8 @@
-import { Dropdown, styled, TextField } from 'office-ui-fabric-react';
+import { Dropdown, TextField } from 'office-ui-fabric-react';
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 
+import { Mode } from '../../../types/action';
 import { IQueryInputProps } from '../../../types/query-runner';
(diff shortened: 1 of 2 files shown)


src/app/views/query-runner/QueryInput.tsx
```diff
@@ -29,6 +31,7 @@ export class QueryInput extends Component<IQueryInputProps, any> {
             role='listbox'
             selectedKey={selectedVerb}
             options={httpMethods}
+            disabled={mode === Mode.TryIt}
```
MIchaelMainer: TryIt mode will eventually support methods other than GET so perhaps we want to add a new mode called GetOnly to support this state.

PR #177, opened 2019-09-12
Feature/AdaptiveCards

### LINKED ISSUE #92: Integrate Adaptive Cards to Graph Explorer
Adaptive Cards are an open card exchange format enabling developers to exchange UI content in a common and consistent way.  Adaptive cards will create a beautiful UI to render JSON responses once a query runs and this in turn be useful especially for citizen developers with not much coding experience. 

The plan is to add  a new tab for Adaptive cards next to the Response Headers tab as shown on the screenshot below. 

image

There are only two templates available for now and we are going to prototype with those for now as we wait for the Adaptive cards team to work on more templates. If the query does not have an adaptive card, the adaptive card tab will not show.

However, we might need to brainstorm on how to tell the users the queries that have adaptive cards. 

Link to adaptive cards site: https://adaptivecards.io/

andrueastman:
## Overview

This PR add support for visual response for some of the responses for the graph by the use of [Adaptive Cards](https://adaptivecards.io/).

The PR currently supports a visual response for calls to calls to /me, /me/manager and /me/drive/root/children and more will be added as more adaptive cards templates are created by the adaptive cards team.

Tests have also been added. 😄 

This will close #92 

### Demo

image

image

### Notes

## Testing Instructions

* Make a request https://graph.microsoft.com/v1.0/me/ and switch to the Response Visual tab to view the card.
* If a response is not supported the tab will have the message "The Adaptive Card for this response is not available"


package.json
```diff
@@ -8,6 +8,7 @@
     "@svgr/webpack": "4.1.0",
     "@types/hellojs": "1.16.2",
     "@uifabric/styling": "6.46.0",
+    "adaptivecards": "^1.1.0",
```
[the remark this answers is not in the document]
jobala: +1.

This way we always have exact version of dependecies running on all machines. Prevents the it works on my machine syndrom. :-)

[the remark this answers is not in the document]
Kachulio1: @jobala  I think having a `yarn.lock` or `package-lock` have the exact version, so when you run `npm i` it will try to install the latest changes but if there is a `lock file` in the dir then it will install the exact version specified in the `lock file` unless someone runs an upgrade command.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+class AdaptiveCard extends Component<IAdaptiveCardProps> {
+  private adaptiveCard: AdaptiveCardsAPI.AdaptiveCard;
+  constructor(props: IAdaptiveCardProps) {
+    super(props);
+    this.adaptiveCard = new AdaptiveCardsAPI.AdaptiveCard();
+  }
+
+  public componentWillMount() {
+    const { body , sampleQuery, hostConfig } = this.props;
+    this.props.actions!.getAdaptiveCard(body, sampleQuery);
```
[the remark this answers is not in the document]
andrueastman: Thanks for pointing this out.

public/index.html
```diff
@@ -84,5 +84,5 @@
   <noscript>You need to enable JavaScript to run this app.</noscript>
   <div id="root"></div>
 </body>
-
+  <script type="text/javascript" src="https://unpkg.com/markdown-it/dist/markdown-it.min.js"></script>
```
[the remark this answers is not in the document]
andrueastman: This is now resolved.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (hostConfig) {
```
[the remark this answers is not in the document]
andrueastman: Now passing an actual config for the light theme. The check is now left for possible reuse of the component.

package.json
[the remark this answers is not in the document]
andrueastman: This is now resolved to align to convention.

src/app/services/actions/adaptive-cards-action-creator.ts
```diff
@@ -0,0 +1,82 @@
...
+    response: ''
+  };
+}
+
+export function getAdaptiveCard(payload: string, sampleQuery: IQuery): Function {
+  return async (dispatch: Function) => {
+
+    if (!payload) {
+      // no payload so return empty result
+      return dispatch(getAdaptiveCardSuccess(''));
```
jobala: Instead of calling `getAdaptiveCardSuccess` with an empty string as an argument, you can take advantage of default parameters.

```js
function getAdaptiveCardSuccess(result = '') {}
```

Then when there's not payload you just call `getAdaptiveCardSuccess()`

src/app/services/actions/adaptive-cards-action-creator.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (template === '' ) {
+      // we dont support this card yet
+      return dispatch(getAdaptiveCardError('No template available'));
+    }
+
+    const url = `https://templates.adaptivecards.io/graph.microsoft.com/${template}`;
+    const body = JSON.stringify(payload);
+    dispatch(getAdaptiveCardPending());
+
+    return fetch(url, {
```
jobala: Does this work in Edge?

src/app/services/actions/adaptive-cards-action-creator.ts
```diff
@@ -0,0 +1,82 @@
...
+function lookupTemplate(sampleQuery: IQuery) {
+  if (sampleQuery) {
+    const urlObject: URL = new URL(sampleQuery.sampleUrl);
+    // remove the prefix i.e. beta or v1.0 and any possible extra whack character at the end'/'
+    const query = urlObject.pathname.substr(6).replace(/\/$/, '') + urlObject.search;
+    const template = templateTable[query];
+    if (template) {
+      return template;
+    }
+  }
```
jobala: Do you think it would be better to use the return first approach here?
```js
if (!sampleQuery) { return }

...
```
Also you can specify the return type.

src/app/services/actions/adaptive-cards-action-creator.ts
```diff
@@ -0,0 +1,82 @@
...
+  return '';
+}
+
+const templateTable: any = {
```
jobala: More of `templateMap?` or `templates`

src/app/services/actions/adaptive-cards-action-creator.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+function lookupTemplate(sampleQuery: IQuery) {
+  if (sampleQuery) {
+    const urlObject: URL = new URL(sampleQuery.sampleUrl);
+    // remove the prefix i.e. beta or v1.0 and any possible extra whack character at the end'/'
+    const query = urlObject.pathname.substr(6).replace(/\/$/, '') + urlObject.search;
+    const template = templateTable[query];
+    if (template) {
+      return template;
```
jobala: What do you think of `return templateTable[query]` ?

src/app/services/reducers/adaptive-cards-reducer.ts
```diff
@@ -0,0 +1,29 @@
+import { IAction } from '../../../types/action';
```
jobala: Indentation is off in this file. It uses 4 instead of 2 spaces.

src/app/services/redux-constants.ts
```diff
@@ -20,3 +20,6 @@ export const SAMPLES_FETCH_SUCCESS = 'SAMPLES_FETCH_SUCCESS';
 export const SAMPLES_FETCH_ERROR = 'SAMPLES_FETCH_ERROR';
 export const SAMPLES_FETCH_PENDING = 'SAMPLES_FETCH_PENDING';
 export const CHANGE_THEME_SUCCESS = 'CHANGE_THEME_SUCCESS';
+export const ADAPTIVE_FETCH_SUCCESS = 'ADAPTIVE_FETCH_SUCCESS';
```
jobala: Action constants sound better when you start with the verb ie `FETCH_ADAPTIVE_CARD_SUCCESS` :)

src/app/services/actions/adaptive-cards-action-creator.ts
andrueastman: Yes it does

src/app/views/query-response/QueryResponse.tsx
```diff
@@ -64,6 +67,22 @@ class QueryResponse extends Component<IQueryResponseProps, {}> {
     ];
 
     if (mode === Mode.Complete) {
+      pivotItems.push(
+        <PivotItem
+          key = 'adaptive-cards'
```
jobala: Do we only want to show adaptive cards for the complete graph explorer experience?

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    if (pending) {
+      return (<Monaco
+        body={'Fetching Adaptive Card ...'}
+      />);
+    }
+
+    if (data) {
+      try {
+        this.adaptiveCard.parse(data);
```
jobala: This parsing will be done everytime the component re-renders which might result in a perfomance hit. Perhaps, move this to `componentDidMount`?

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+      this.props.actions!.getAdaptiveCard(body, sampleQuery);
+    }
+  }
+
+  public componentWillUnmount() {
+    // Remove all references
+    delete this.adaptiveCard;
+  }
+
+  public shouldComponentUpdate(nextProps: IAdaptiveCardProps) {
```
jobala: What do you think of the approach below?

```js
const sameBody = ...
const sameCardData = ...

return !sameBody || !sameCardData
```

src/app/services/actions/adaptive-cards-action-creator.ts
andrueastman: +1

src/app/services/actions/adaptive-cards-action-creator.ts
andrueastman: +1 on specifying the return type. 

Returning first is rendered moot with `return templateTable[query]` raised in later comment

src/app/services/reducers/adaptive-cards-reducer.ts
andrueastman: Well spotted sir!

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
andrueastman: This is not really an issue as the `componentShouldUpdate` prevents unnecessary renders as discussed.

src/app/views/query-response/QueryResponse.tsx
andrueastman: For now I think we can treat this just as the snippets tab and make any changes based on any new feedback.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
andrueastman: From my point of view, the implemented approach is a bit easier to read and follow with the code.

PR #184, opened 2019-09-14
Fix/snippets fetching on keypress

jobala:
## Overview

Requests for code snippets when the url textinput loses focus.

### Demo

N/A

### Notes

Initially, code snippets were fetched every time the `onChange` event was fired. This would have resulted in the api being hit too many times.

## Testing Instructions

* Launch Graph Explorer
* Click on the snippets tab
* Modify the sample url
* Observe that the snippets are fetched only when the sample url textinput control loses focus

### DIFF
--- a/src/app/views/query-runner/QueryRunner.tsx
+++ b/src/app/views/query-runner/QueryRunner.tsx
@@ -81,11 +81,11 @@ export class QueryRunner extends Component<
 
     if (actions) {
       actions.runQuery(sampleQuery);
-    }
+    } 
   };
 
   public render() {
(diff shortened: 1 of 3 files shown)


src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -78,7 +85,7 @@ export class QueryRunner extends Component<
   };
 
   public render() {
-    const { httpMethods } = this.state;
+    const { httpMethods, url } = this.state;
```
Kachulio1: `url` is declared but its value is never read.

PR #202, opened 2019-09-19
removes the loading spinner from the portal

### LINKED ISSUE #178: Bug: Fix the never ending loading progress bar in V2 -tst environment
**Describe the bug**
There's a never ending loading progress bar on the V2 -tst environment' s page. It appears at the bottom of the page, in a distorted way,

**To Reproduce**
1. Go to https://developer.microsoft-tst.com/en-us/graph/graph-explorer-v2/preview?1&setvar=routegraph:https://nghia4:20444
2. Let the page load
3. Scroll down the page
4. You'll see the loading progress bar

**Expected behavior**
The loading progress bar should disappear as soon as the contents load.

**Screenshots**
image

 - OS: [Windows]
 - Browser [Chrome and Edge]
 - Version []

**Additional context**

cwahome:
## Overview
The loading spinner on the portal does not stop spinning even after the app loads.

Now we check the existence of the html tag `<api-explorer>` that houses the loading spinner on the portal website and remove it from the DOM once Graph explorer is loaded

Fixes #178 

## Testing Instructions

* Log on to the tst environment
* Observe a "starting microsoft graph explorer" spinner
* Notice that it is removed from view after loading is complete

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -26,10 +26,16 @@ import { loadGETheme } from './themes';
 import { Mode } from './types/action';
 import { IHistoryItem } from './types/history';
 
+// removes the loading spinner from GE html after the app is loaded
 const spinner = document.getElementById('spinner');
-
 if (spinner !== null) {
   (spinner as any).parentElement.removeChild(spinner);
+}
+
+// removes the loading spinner from the portal team html after GE loads
+const apiExplorer = document.getElementsByTagName('api-explorer')[0];
+if (apiExplorer && apiExplorer !== null) {
+  (apiExplorer as any).parentElement.removeChild(apiExplorer);
 }
 
 initializeIcons();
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -26,12 +26,18 @@ import { loadGETheme } from './themes';
...
+// removes the loading spinner from GE html after the app is loaded
 const spinner = document.getElementById('spinner');
-
 if (spinner !== null) {
   (spinner as any).parentElement.removeChild(spinner);
 }
 
+// removes the loading spinner from the portal team html after GE loads
+const apiExplorer = document.getElementsByTagName('api-explorer')[0];
+if (apiExplorer && apiExplorer !== null) {
```
jobala: Why do you have to check explicitly that apiExplorer is not equal to null?

cwahome: I don't.. Was being super safe

jobala: It's unnecessary, `if (apiExplorer) {..}` should be enough.

PR #206, opened 2019-09-30
Feature/msal migration

cwahome:
## Overview

Changes the authentication implementation from hellojs to msal
implements Msal through a provider to be used in the graph client


src/app/services/graph-client/MsalAuthProvider.ts
```diff
@@ -0,0 +1,116 @@
...
+  private defaultUserScopes: string[];
+
+  constructor() {
+    const clientId = process.env.REACT_APP_CLIENT_ID || '';
+    const options = {
+      redirectUri: window.location,
+      storeAuthStateInCookie: true,
+    };
+    this.defaultUserScopes = DEFAULT_USER_SCOPES.split(' ');
+    this.authProvider = new MSALAuthenticationProvider(clientId, this.defaultUserScopes, options);
```
darrelmiller: This must be using an old version of the MSALAuthenticationProvider, this is not how it works in v2 of msgraph Javascript SDK.

cwahome: Yes. Coz we are still using the sdk version at 1.5.0

darrelmiller: I would like this to be updated to the 2.0 SDK.  Also, we fixed the AuthProvider so that you create the UserAgent externally and then pass it into the AuthProvider, so there should be no need for an AuthProvider to wrap the AuthProvider, you should be able to just use the ImplicitMSALAuthProvider directly without having to wrap it and re-implement the AuthenticationProvider interface.

src/app/views/authentication/Authentication.tsx
```diff
@@ -46,7 +49,7 @@ export class Authentication extends Component<IAuthenticationProps> {
 
 function mapStateToProps(state: any) {
   return {
-    tokenPresent: !!state.authToken
+    tokenPresent: !!state.authToken,
```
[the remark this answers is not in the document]
cwahome: autoadded by the linter

src/app/services/graph-client/MsalAuthProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  public async signIn() {
+    const loginRequest = this.defaultUserScopes;
+
+    if (this.loginType === 'POPUP') {
+      try {
+        const response = await this.userAgentApp.loginPopup(loginRequest);
+        return response;
+      } catch (error) {
+        return Promise.resolve(null);
```
jobala: Why are you resolving the promise with null on errors?

src/app/services/graph-client/MsalAuthProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+   * Checks to see if the token exists in the cache
+   * @param {string[]} scopes optional scopes passed to get the token silently
+   */
+  public getTokenSilent(scopes?: string[]): any {
+    const userScopes = scopes || this.defaultUserScopes;
+    try {
+      const silentToken = this.userAgentApp.acquireTokenSilent(userScopes);
+      return silentToken;
+    } catch (error) {
+      return null;
```
jobala: Why are you returning null?

darrelmiller:
@jobala We do not want to duplicate the GE V1 MSAL code.  We want to use the Graph JavaScript SDK 2.0 with the ImplicitMSALAuthProvider because that code has been written to handle incremental consent, Multi-Factor Authentication, Conditional Access and other Identity Platform related exception cases.  We do not want to have to re-write that code again in Graph Explorer.  Additionally we want to help to validate the Microsoft Graph Javascript SDK by being a consumer of that library.  /cc @MIchaelMainer  @bettirosengugi  @cwahome @jthake-msft

jobala:
@darrelmiller I don't understand the requirements for this too well and I am not familiar with SDK 2.0 so I am going to trust your opinion.

cwahome:
> Please implement MSAL the same way it was implemented in GE classic.

Classic did not use the graph sdk so there was no use for an auth provider.

src/app/services/graph-client/MsalService.ts
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  try {
+    const response = await msalApplication.acquireTokenSilent(loginRequest);
+    return response;
+  } catch (error) {
+    if (requiresInteraction(error)) {
+      if (loginType === 'POPUP') {
+        try {
+          return msalApplication.acquireTokenPopup(loginRequest);
```
andrueastman: It looks like the try catch here may be moot as the promise will not be awaited therefore be returned before an error can be thrown.
Maybe you want to do this?  `return await msalApplication.acquireTokenPopup(loginRequest);`
Checkout the SO question [here](https://stackoverflow.com/questions/38708550/difference-between-return-await-promise-and-return-promise).

src/app/services/graph-client/MsalService.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (loginType === 'POPUP') {
+    try {
+      await msalApplication.loginPopup(loginRequest);
+      const response = await msalApplication.acquireTokenPopup(loginRequest);
+      return response.accessToken;
+    } catch (error) {
+      return null;
+    }
+  } else if (loginType === 'REDIRECT') {
+    await msalApplication.acquireTokenRedirect(loginRequest);
```
andrueastman: Looks like this could throw as well. 
Also, since we are not really using redirects, does it sound fair to remove similar sections from the code?

andrueastman:
👍 This works well for me on edge and chrome.

src/app/services/graph-client/MsalService.ts
cwahome: It does make sense

src/app/views/authentication/profile/Profile.tsx
```diff
@@ -97,7 +99,7 @@ export class Profile extends Component<IProfileProps, IProfileState> {
 
     const persona: IPersonaSharedProps = {
       imageUrl: user.profileImageUrl,
-      imageInitials: this.getInitials(user.displayName),
+      imageInitials: (user.displayName) ? this.getInitials(user.displayName) : '',
```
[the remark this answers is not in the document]
andrueastman: 👍

src/app/services/graph-client/MsalService.ts
andrueastman: 👍

PR #205, opened 2019-09-27
Add version select

Kachulio1:
## Overview
 enable  URL version select
FIxes #200 

### Demo
Screen Shot 2019-09-27 at 20 41 03

PR #215, opened 2019-10-14
Fix: add version select

### LINKED ISSUE #200: Add the version selection bar
V1 has the section where you can select the version number, V2 doesn't have this feature, making the users to run v2 queries by default unless they manually edit the url. 

Introduce the select bar like the one in the screenshot below to enable selection of the version

image

**Additional context**

cwahome:
## Overview

Adds a version selector option
The version selection changes automatically when you select queries from the sidebar.

Improves #205 
Fixes #200 

### Demo

image

## Testing Instructions

* Change the version from v1.0 to beta and notice how the query url changes
* Click on a url in the queries and notice the change in the version selector


jobala:
@bettirosengugi a while back there was the issue reported by @jthake-msft where for certain screen sizes the http methods were shown to have an ellipsis ie `D...` for `DELETE`. This implementation will re-introduce that behaviour. Please advise.

bettirosengugi:
> @bettirosengugi a while back there was the issue reported by @jthake-msft where for certain screen sizes the http methods were shown to have an ellipsis ie D... for DELETE. This implementation will re-introduce that behaviour. Please advise.

@jobala why don't we have the same issue in classic yet it has the three tabs (Request Options, Version and query tabs) ? May be investigate?

bettirosengugi:
> In the current version of Graph Explorer changing the version in the URL changes the selected version in the version dropdown. I am assuming this is how this feature should behave as well. @bettirosengugi  is this correct?

This is correct to give an example if the version selected is V1, the url for /me should be https://graph.microsoft.com/v1.0/me  and when you change the version to beta the url should change to https://graph.microsoft.com/beta/me 

@cwahome  @jobala

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -85,10 +80,23 @@ export class QueryRunner extends Component<
     }
   };
 
-  public render() {
-    const { httpMethods } = this.state;
-    const { graphExplorerMode, isLoadingData } = this.props;
+  private handleOnVersionChange = (option?: IDropdownOption) => {
+    const { sampleQuery } = this.props;
+    if (option) {
```
jobala: Rename this to a more descriptive name.

Variables should describe the data they hold.

for example, urlVersion

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -85,10 +80,23 @@ export class QueryRunner extends Component<
...
+      const urlObject: URL = new URL(sampleQuery.sampleUrl);
+      const requestUrl = urlObject.pathname.substr(5).replace(/\/$/, '');
+      const sampleUrl = `${GRAPH_URL}/${option.text + requestUrl + decodeURI(urlObject.search)}`;
+
+      this.props.actions!.selectQueryVersion(option.text);
```
jobala: This action creator sets the query version. A more appropriate name will be setQueryVersion.

src/app/views/sidebar/history/History.tsx
```diff
@@ -233,8 +233,10 @@ export class History extends Component<IHistoryProps, any> {
   private onRunQuery = (query: IHistoryItem) => {
     const { actions } = this.props;
```
jobala: Unnecessary newline.

jobala:
> This is correct to give an example if the version selected is V1, the url for /me should be https://graph.microsoft.com/v1.0/me and when you change the version to beta the url should change to https://graph.microsoft.com/beta/me

There's also a third type of url version `other`. This is used when a user specifies a version that is neither `v1.0` or `beta`

jobala:
>  why don't we have the same issue in classic yet it has the three tabs (Request Options, Version and query tabs) ? May be investigate?

That's a good question. AFAIK classic uses MWF and `v2` uses Fabric, these are different component providers. It's worth investigating.

cwahome:
> In the current version of Graph Explorer changing the version in the URL changes the selected version in the version dropdown...

Now changes when you change the url in the editor. Good catch @jobala 

> @bettirosengugi a while back there was the issue reported by @jthake-msft where for certain screen sizes the http methods were shown to have an ellipsis ie D... for DELETE. This implementation will re-introduce that behaviour. Please advise.

I have improved the column definitions to reduce the need for the ellipsis in smaller screen sizes

> The above code snippet has been duplicated in various places. It will be wise to abstract it into a function.

I have added a utility function for reuse for this.

> Why do we need a reducer and an action-creator for the version selector?

The version is used and changed at different points of the application.

PR #223, opened 2019-10-16
adds SdkVersion as default header

### LINKED ISSUE #222: Add a header to differentiate Classic and V2 apps
We need a way to differentiate between the two versions of graph explorer we are working on.

add a header to every request to know if it was called from which version of GE. 

Name of header: GE Version
Header description for Classic: GraphExplorer/v03
Header description for new version: GraphExplorer/v04

cwahome:
## Overview

Adds a default header SdkVersion to all requests.

Fixes #222 

## Testing Instructions

* Hit run query
* Check the network tab, and look at the request headers
* Note the addition of the GE-Version header


src/app/services/actions/query-action-creator-util.ts
```diff
@@ -34,6 +34,7 @@ export async function anonymousRequest(dispatch: Function, query: IQuery) {
   const headers = {
     'Authorization': `Bearer ${authToken}`,
     'Content-Type': 'application/json',
+    'GE-Version': 'GraphExplorer/v04',
```
jobala: Why are we using v04 and not v02?

cwahome: There have been other Graph explorers before ...
This is how this one is registered as

jobala: Registered where by who? This version has v2 in it's name

bettirosengugi: > Registered where by who? This version has v2 in it's name O_o

@jobala Before we joined Microsoft, there were 2 versions already developed. Classic is the 3rd version of GE and what we actually call V2 is actually supposed to be V4 hence the naming.

jobala: How will naming this v2 affect everything else?

bettirosengugi: > How will naming this v2 affect everything else? @bettirosengugi do you know if those versions are used? I am aiming for consistency here, it will be confusing to have a version labeled as v2 sending headers identifying it as v4.

It won't affect anything. The versions were retired just like we shall retire Classic once done with it, for continuity purposes, it's important to acknowledge that there were two versions before this and that's why I suggest we use 4.x.x as the new versioning number.

jobala: If it won't affect anything, I don't think it's worth trading it off for consistency. This would have been an easier decision to make early on in the project but now we have v2 in various places like the Azure AppService, GitHub repo, etc renaming it to v4 will mean renaming it every where else.

MIchaelMainer: Making the change to v4 doesn't make sense to me. Customers make no distinction about v2 being the 4th generation. We weren't making this distinction until now and it cost us nothing. Who are we doing this for? What benefits will this provide them?

darrelmiller: @MIchaelMainer @jobala It is true that customers do not see this version and therefore the value we pick makes no difference to them.  However, Yina, who has the longest history with this project, says it should be V4.  I made the mistake of calling it V2, so we are going to fix it before we release.

bettirosengugi: @MIchaelMainer We are adding this header here since we are using the same app id for the 2 versions and this will help us make the distinction on Kusto.

MIchaelMainer: @darrelmiller It still makes little sense to make the change, but since it costs little to make the change and since it is a requirement from the Mother of Graph, it so shall be.

So, we just need to remember that 2 = 4 from now on ;).

MIchaelMainer: @bettirosengugi Does this mean that the header from GE classic is going to be `GraphExplorer/v03`?

darrelmiller: I don't know where the "v04" syntax came from.  Product Versions are usually of the format `n.n` as shown here https://tools.ietf.org/html/rfc7231#section-5.5.3 for User-Agent.  The only reason we are not using User-Agent is because Chrome can't write to the User-agent header.
Therefore we should use `GraphExplorer/3.0` and `GraphExplorer/4.0 `

PR #233, opened 2019-10-17
Fix/double scroll tst

### LINKED ISSUE #216: UI Issues in -tst env for V2
**Describe the bug**
1. When loading the page https://developer.microsoft-tst.com/en-us/graph/graph-explorer/preview the progress button and text is not centered. See Screenshot below>

image

2. Try Clicking on any link on this page: developer.microsoft-tst.com/en-us/graph/graph-explorer/preview it doesn't redirect but instead outlines the Samples tab. Even the NPS prompt doesn't pop up the feedback form

3. Also, the Response Header section at the bottom overflows to the footer of the parent page. See screenshot below: 

image

One more screenshot.

image

 - OS: [Windows]
 - Browser [Edge]

**Additional context**

jobala:
## Overview

Reduces the height of the request editor so that Graph Explorer can fit into the available 'viewport' when hosted within the portal.

### Demo

N/A

### Notes

In [this PR](https://github.com/microsoftgraph/microsoft-graph-explorer-v2/pull/232) I tried setting the height to `1024px` and then adding an overflow property. This worked as far as preventing Graph Explorer from overflowing. The caveat was that, the user would have to scroll twice. So I decided against that implementation.

This implementation reduces the size of the request editor so that Graph Explorer fits in the portal without overflowing.

## Testing Instructions

N/A

Fixes #216

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6693,11 +6693,13 @@
         },
         "balanced-match": {
           "version": "1.0.0",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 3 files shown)


jobala:
After Ignite.

PR #241, opened 2019-10-22
Feature: background colour on request option

### LINKED ISSUE #228: Add a tile for the request options with  the right background color
 Add a tile for the request options with  the right background color according to REST API color code, as the screenshot below: 

image

image

Related to #226

cwahome:
## Overview

Adds a tile for the request options with the right background colour

Fixes #228 

### Demo

Light

image

Dark

image

## Testing Instructions

* Scroll through the sample queries
* Notiice the different colours of the sample methods


jobala:
Which issue does this fix? for purposes of posterity.

src/app/utils/badge-color.ts
```diff
@@ -0,0 +1,26 @@
+import { getTheme } from 'office-ui-fabric-react';
+
+export function getMethodStyle (method: string) {
+  const currentTheme = getTheme();
+
+  method = method.toString().toUpperCase();
```
jobala: Type of method is String. Why do we need to convert it to a string before uppercasing it?

src/app/utils/badge-color.ts
```diff
@@ -0,0 +1,26 @@
...
+    case 'PUT':
+      return currentTheme.palette.magentaDark;
+
+    case 'PATCH':
+      return currentTheme.palette.yellowDark;
+
+    case 'DELETE':
+      return currentTheme.palette.redDark;
+
+    default:
```
jobala: POST already uses the color green, why are we not using a different color for verbs not catered for by the switch statement?

src/app/utils/badge-color.ts
```diff
@@ -0,0 +1,26 @@
+import { getTheme } from 'office-ui-fabric-react';
+
+export function getMethodStyle (method: string) {
```
jobala: This function will read better if renamed to `getStyleFor(method: string) {...}`

src/app/views/sidebar/Sidebar.styles.ts
```diff
@@ -91,6 +91,17 @@ export const sidebarStyles = (theme: ITheme) => {
...
+      display: 'inline-block',
+      textAlign: 'left',
+      marginRight: '15px',
+      lineHeight: 'normal',
+      whiteSpace: 'nowrap',
+      paddingLeft: 10,
+      paddingRight: 5,
+      paddingTop: 3,
+      paddingBottom: 3,
+      color: '#fff',
```
jobala: we can use the value from the theme palette.

cwahome: When we do, and change themes, the white turns to black... which is not as accessible.

src/app/utils/badge-color.ts
cwahome: because we do not expect to get others not catered for. However, I can pick a cool colour to display.

src/app/utils/badge-color.ts
cwahome: Good catch

src/app/utils/badge-color.ts
jobala: It will be cool to pick a cool color

bettirosengugi: Please use the guide below for color codes for REST API verbs. 
image
@jobala @cwahome

andrueastman: And what would be the color for PATCH? It is missing from the list.

src/app/views/sidebar/Sidebar.styles.ts
```diff
@@ -91,6 +91,17 @@ export const sidebarStyles = (theme: ITheme) => {
     badge: {
       fontWeight: FontWeights.bold,
       fontSize: FontSizes.small,
+      display: 'inline-block',
+      textAlign: 'left',
```
andrueastman:
```
textAlign: 'center',
```
Maybe?

src/app/views/sidebar/Sidebar.styles.ts
```diff
@@ -91,6 +91,17 @@ export const sidebarStyles = (theme: ITheme) => {
...
+      marginRight: '15px',
+      lineHeight: 'normal',
+      whiteSpace: 'nowrap',
+      paddingLeft: 10,
```
andrueastman: Maybe make the padding on both right and left. It looks better IMO.

PR #253, opened 2019-10-25
Reduces the list of default user scopes

jobala:
## Overview

Reduces the list of default user scopes

### Demo

N/A

### Notes

N/A

## Testing Instructions

N/A


jobala:
Merging this because I need to C if it works in -tst

PR #257, opened 2019-10-28
Feat/hide login edge

jobala:
## Overview

Hides authentication button in Edge/IE

### Demo

image

### Notes

Authentication is disabled because msal breaks when `acquireTokenPopUp` is called in Edge. We also. can't switch to using redirect because it will break the incremental consent feature.

## Testing Instructions

* Launch GE in Edge
* Observe that you can't log in

### DIFF
--- a/src/app/views/App.tsx
+++ b/src/app/views/App.tsx
@@ -208,6 +208,7 @@ class App extends Component<IAppProps, IAppState> {
     const isEdge = msedge > 0;
     return isEdge;
   }
+
   public optOut = () => {
     const path = location.href;
     const urlObject: URL = new URL(path);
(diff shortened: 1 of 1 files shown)


src/app/views/App.tsx
```diff
@@ -199,6 +199,16 @@ class App extends Component<IAppProps, IAppState> {
     this.props.actions!.toggleSidebar(properties);
   }
 
+  public userIsOnEdge() {
+    const ua = window.navigator.userAgent;
+    const msie = ua.indexOf('MSIE ');
+    const msie11 = ua.indexOf('Trident/');
+    const msedge = ua.indexOf('Edge/');
+    const isIE = msie > 0 || msie11 > 0;
```
andrueastman: Seems that this line is of no use to us. Can we have it removed?
```const isIE = msie > 0 || msie11 > 0;```

PR #293, opened 2019-11-15
Feat/backward compatible share queries

### LINKED ISSUE #143: add a Share Link button on Graph Explorer V2
Add this button on the screenshot below
image

To allow users share the query string on their session as it is by copying it. 

image

Additional requirements.
1. **Enable backward compatibility to enable share links between Classic and V2**
2. **Enable sharing the request body as well is the user has input a key and a value as part of the request body**

jobala:
## Overview

* Allows users to share queries between Graph Explorer(s)
* Adds support for sharing request bodies(v4 only)
* Adds support for sharing queries between v4 & v3
### Demo

image

#### Backwards Compatibility
This Query sharing feature is backward compatible with GE v3. Sharing the request body is only available for v4. Adding support for request body in v3 will involve feature work and as it is now v3 is on a feature freeze.

#### Encoding Request Body
The `request body` shared between instances of GE is `Base64` encoded. In principle the `Base64` string should be smaller than the actual contents of the encoded string helping us avoid url character limit.

## Testing Instructions

* Launch Graph Explorer
* Click Share button
* Copy the url in the dialog
* Open the url on web client running v3 or v4
* Observe that the QueryInput is updated

Close #143

### DIFF
--- a/src/app/views/query-response/QueryResponse.tsx
+++ b/src/app/views/query-response/QueryResponse.tsx
@@ -46,8 +46,7 @@ class QueryResponse extends Component<IQueryResponseProps, { showShareQueryDialo
 
   private generateShareQueryParams = (): string => {
     const { sampleQuery: { sampleBody, sampleUrl, selectedVerb, selectedVersion } } = this.props;
-    const origin = window.location.origin;
-    const pathname = window.location.pathname;
+    const { origin, pathname } = window.location;
     const url = new URL(sampleUrl);
     const graphUrl = url.origin;
     /**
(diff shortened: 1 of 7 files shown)


src/app/views/query-response/QueryResponse.tsx
```diff
@@ -7,14 +8,64 @@ import { ThemeContext } from '../../../themes/theme-context';
...
 import { IQueryResponseProps } from '../../../types/query-response';
 import { Image, Monaco } from '../common';
-import AdaptiveCard  from './adaptive-cards/AdaptiveCard';
+import AdaptiveCard from './adaptive-cards/AdaptiveCard';
 import { darkThemeHostConfig, lightThemeHostConfig } from './adaptive-cards/AdaptiveHostConfig';
 import './query-response.scss';
 import { Snippets } from './snippets';
 
-class QueryResponse extends Component<IQueryResponseProps, {}> {
+class QueryResponse extends Component<IQueryResponseProps, { showShareQueryDialog: boolean, query: string }> {
```
cwahome: move this to an interface as it looks cleaner

Issue #270, opened 2019-10-30
Templated variables are not updated in the request payload


darrelmiller:
**Describe the bug**
Many samples include special templated variables.  e.g {domain} in the sample request payload.  Before sending these requests they should be updated with valid values.

**To Reproduce**
1. Go Create User sample
2. Click on Run Query
3. See 400 bad request

**Expected behavior**
A 200 OK.

PR #298, opened 2019-11-25
Feature: show query tip message

### LINKED ISSUE #297: Display a tip message when a user is logged in and the query has a tooltip
Users need to know the parameters that need changing before sending a request. The tips are already present but not visible

once a user is logged in and they click a query that has a tip, they should see the tip and the query should not automatically run.

cwahome:
## Overview

Displays a tip message for queries with parameters that need changing before running the query.
Related to #270 
FIxes #297

### Demo

image

### Notes

## Testing Instructions

* Log in to Graph Explorer
* On the sidebar in the Groups section, select 'group members'
* Notice the query tip shows up with a tip on how to best run the query


src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -300,6 +308,16 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
   }
 
 }
+function displayTipMessage(actions: any, selectedQuery: any) {
```
jobala: I think we have a Query type that you can use here instead of setting it to any.

src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -300,6 +308,16 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
...
+  if (selectedQuery.tip) {
```
jobala: The functions calling displayTipMessage already check if the tip exists, is there a need to double check?

PR #300, opened 2019-11-26
Task/refactor copy item

jobala:
## Overview

1. Reuses the copy functionality instead of duplicating it.
2. Fixes a small bug with the share query functionality where it would print undefined in the RequestEditor if no request body was shared.

### Demo

Keeping the code dry will help with future refactoring.

### Notes

N/A

## Testing Instructions

N/A


src/app/views/App.tsx
```diff
@@ -127,7 +127,13 @@ class App extends Component<IAppProps, IAppState> {
   }
 
   private hashDecode(requestBody: string): string {
-    return requestBody ? atob(requestBody) : '';
+    const decodedBody = atob(requestBody);
+
+    if (decodedBody === 'undefined') {
```
andrueastman: Out of curiosity, is there a specific case where this would occur?

From [this](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/atob), I can only tell that `atob` will throw if the conversion fails.

jobala: The throw exception occurs for non valid base64. This handles the case where a falsey value is  base64 encoded.

andrueastman: Thanks so much for explaining that. 😄

PR #315, opened 2019-12-09
HB of localized jsons

OfficeGlobal:
HB of localized json files


jobala:
@OfficeGlobal please resolve the conflicts so that we can merge this.

PR #316, opened 2019-12-11
calls acquiretoken popup when silent fails

cwahome:
## Overview

calls acquiretoken popup when silent fails

### DIFF
--- a/src/app/services/graph-client/msal-service.ts
+++ b/src/app/services/graph-client/msal-service.ts
@@ -21,7 +21,7 @@ export async function logIn(): Promise<any> {
       if (requiresInteraction(error)) {
         return acquireTokenWIthInteraction(loginRequest);
        } else {
-         return null;
+        throw error;
        }
     }
   } else if (loginType === LoginType.Redirect) {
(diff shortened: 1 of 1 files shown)


src/app/services/graph-client/msal-service.ts
```diff
@@ -18,13 +18,26 @@ export async function logIn(): Promise<any> {
       const authResponse = await msalApplication.acquireTokenSilent(loginRequest);
       return authResponse;
     } catch (error) {
-      return null;
+      if (requiresInteraction(error)) {
+        return acquireTokenWIthInteraction(loginRequest);
+       } else {
+         return null;
```
jobala: If an error happens we won't be able to track it down because we are returning null, instead, throw the error for now. We'll need to think about handling errors gracefully.

src/app/services/graph-client/msal-service.ts
```diff
@@ -18,13 +18,26 @@ export async function logIn(): Promise<any> {
...
     await msalApplication.loginRedirect(loginRequest);
   }
 }
 
+async function acquireTokenWIthInteraction(loginRequest: AuthenticationParameters) {
+  try {
+    const authResponse = await msalApplication.acquireTokenPopup(loginRequest);
+    return authResponse;
+  } catch (error) {
+    return null;
```
jobala: Same for here.

src/app/services/graph-client/msal-service.ts
```diff
@@ -18,13 +18,26 @@ export async function logIn(): Promise<any> {
...
+  }
+}
+
 function authCallback(error: any, response: any) {
```
jobala: What does this function do?

PR #320, opened 2019-12-19
Feature: change theme

### LINKED ISSUE #219: Add  a change theme button under the drop down bar on profile
Some users would benefit from being able to change theme's in Graph Explorer. 

add a button to change theme with the options : Dark, Light or High Contrast.

thewahome:
## Overview

Enables changing of the current theme using the settings button.

Closes #219 

### Demo

theme-change

### Notes

Reloads the page on theme selection so that the whole app can re-render.

## Testing Instructions

* Load GE and click the gear icon next to the authentication section
* From the drop-down, select change theme
* Choose among Light / Dark / High Contrast
* Wait for the page to reload


jobala:
The page should not reload on theme selection.

src/store/theme-cache.ts
```diff
@@ -0,0 +1,10 @@
+const key = 'CURRENT_THEME';
```
jobala: Does it make sense to rename this to
```js
const selectedTheme = 'SELECTED_THEME';
```

src/store/theme-cache.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export function saveTheme(theme: string) {
+  localStorage.setItem(key, theme);
+}
+
+export function readTheme() {
+  const theme = localStorage.getItem(key) || 'light';
```
jobala: What do you think of creating a default theme variable
```js
const defaultTheme = 'light'

const theme = localStorage.getItem(selectedTheme) || defaultTheme
```

src/store/theme-cache.ts
```diff
@@ -0,0 +1,10 @@
+const key = 'CURRENT_THEME';
+
+export function saveTheme(theme: string) {
```
jobala: There's an enum type for theme, perhaps use that here instead of the string type.

src/messages/en-US.json
```diff
@@ -262,5 +262,6 @@
   "Sign In to try this sample": "Sign In to try this sample",
   "Signing you in...": "Signing you in...",
   "Share Query Message": "Share this link to let people try your current query in the Graph Explorer",
-  "Copy": "Copy"
-}
+  "Copy": "Copy",
+  "Change theme": "Change theme"
+}
```
jobala: Do we need to add light, dark & high contrast here?

src/app/views/settings/Settings.tsx
```diff
@@ -0,0 +1,164 @@
+import {
+  ChoiceGroup, Dialog, DialogType, IconButton
```
jobala: We can put each of the imports on a new line to ease readability.

src/app/views/settings/Settings.tsx
```diff
@@ -0,0 +1,164 @@
...
+} from 'office-ui-fabric-react';
+import React, { Component } from 'react';
+import { injectIntl } from 'react-intl';
+import { connect } from 'react-redux';
+import { bindActionCreators, Dispatch } from 'redux';
+import * as authActionCreators from '../../services/actions/auth-action-creators';
+import * as themeAtionCreators from '../../services/actions/theme-action-creator';
+
+export interface ISettingsProps {
```
jobala: Please place these in a `types` file.

src/app/views/settings/Settings.tsx
```diff
@@ -0,0 +1,164 @@
...
+  actions?: {
+    signOut: Function;
+    changeTheme: Function;
+  };
+  intl?: {
+    message: object;
+  };
+  authenticated: boolean;
+  appTheme: string;
```
jobala: Use the theme enum type instead of string.

src/app/views/settings/Settings.tsx
```diff
@@ -0,0 +1,164 @@
...
+  public toggleThemeChooserDialogState = () => {
+    this.setState({ hideThemeChooserDialog: !this.state.hideThemeChooserDialog });
+  }
+
+  public handleSignOut = () => {
+    this.props.actions!.signOut();
+  }
+
+  public handleChangeTheme = (event: any, option: any) => {
+    const newTheme = option.key;
```
jobala: I don't think it's a good idea to reload the application when the theme changes. We don't have to reload TryIt for it to pick a new theme. You might want to call the `loadGETheme` function here and fix the components that are not picking up the theme.

src/app/views/settings/Settings.tsx
thewahome: The linter we have always used defaults to this.

src/app/views/settings/Settings.tsx
thewahome: On this and this:

> Let's not reload the page on theme change. We might find the loagGETheme function useful.

We actually do need to reload. The reason it works for try it is because once the theme changes on the dev portal it loads a new version of GE with the theme parameters.

loadGETheme does not re-render the screen

thewahome:
> 2. I am not sure how I feel about using a dialog for changing the theme, personally I would have preferred a dropdown like in the image below.
>    image

According to how we have designed the screen, our option at this time would be to have it in a dialog in a 'more actions' button. The design used by the docs team utilises a top nav-bar. Which we do not have. We can use other apps as inspirations but can not copy as is because they will not fit into our context

src/store/theme-cache.ts
thewahome: I don't see how

src/app/views/settings/Settings.tsx
jobala: Nope, it doesn't reload GE, refer to `App.tsx`.

src/app/views/settings/Settings.tsx
jobala: Does it show a lint error when you do this? It doesn't for me.

src/store/theme-cache.ts
jobala: The variable key is not descriptive.

src/types/settings.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface ISettingsProps {
+  actions?: {
+    signOut: Function;
+    changeTheme: Function;
+  };
+  intl?: {
+    message: object;
+  };
+  authenticated: boolean;
+  appTheme: IThemeChangedMessage['theme'];
```
jobala: This works but it creates a dependency with IThemeChangedMessage which can be a source of bugs because the relationship between IThemeChangedMessage and ISettingProps is not obvious.

Can we create a theme enum? And use that instead?

src/app/views/settings/Settings.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { injectIntl } from 'react-intl';
+import { connect } from 'react-redux';
+import { bindActionCreators, Dispatch } from 'redux';
+import { loadGETheme } from '../../../themes';
+import { IThemeChangedMessage } from '../../../types/query-runner';
+import { ISettingsProps } from '../../../types/settings';
+import * as authActionCreators from '../../services/actions/auth-action-creators';
+import * as themeAtionCreators from '../../services/actions/theme-action-creator';
+
+class Settings extends Component<ISettingsProps, any> {
```
jobala: Please define the type for this component's state.

src/app/views/authentication/Authentication.tsx
```diff
@@ -55,10 +56,19 @@ export class Authentication extends Component<IAuthenticationProps, { loginInPro
             <Stack>
               {!tokenPresent &&
                 <>
-                  <Label className={classes.authenticationLabel}>
-                    <Icon iconName='Permissions' className={classes.keyIcon} />
-                    <FormattedMessage id='Authentication' />
-                  </Label>
+                  <div style={{
```
jobala: Place inline styles  in relevant `.styles.ts` file.

src/app/views/settings/Settings.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    const {
+      intl: { messages },
+      authenticated,
+      appTheme
+    }: any = this.props;
+
+    const { hideThemeChooserDialog } = this.state;
+
+    const items: any = [
```
jobala: Items can be defined as a class property and then updated  in `componentDidMount`. Doing computations in a render function is discouraged. The render function is invoked many times.

```js
private items = [];

componentDidMount() {
  if (authenticated) {...}
}

File: src/app/views/settings/Settings.tsx [hunk 13]
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    const menuProperties = {
+      shouldFocusOnMount: true,
+      alignTargetEdge: true,
+      items
+    };
+
+    return (
+      <>
```
jobala: What do the empty tags do?

src/app/views/settings/Settings.tsx
```diff
@@ -0,0 +1,155 @@
...
+                iconProps: { iconName: 'CircleFill' },
+                text: messages.Dark
+              },
+              {
+                key: 'high-contrast',
+                iconProps: { iconName: 'Contrast' },
+                text: messages['High Contrast'],
+              }
+            ]}
+            onChange={(event, option) => this.handleChangeTheme(event, option)}
```
jobala: `handleChangeTheme` does not use the **event** argument.

jobala:
`theme-cache.ts` is in a directory named  **store**  and this being a `redux` application store means something else also the name `theme-cache` implies that it is **acting** as a **cache** which is not the case. 

I recommend renaming the file to util.ts and placing it in **src/themes/**. Let us do away with the store directory and place the history saving/retrieving functionality in the history-util file in the **src/views/history/**.

src/app/views/settings/Settings.tsx
thewahome: They create fragments so when we do not need to nest elements in a div element

src/app/views/settings/Settings.tsx
thewahome: Moving this to be part of the state instead

PR #328, opened 2020-01-09
Core Telemetry Class

jobala:
## Overview

This PR adds a Telemetry class that is an **abstraction** over **ApplicationInsights**. The class provides three public methods.

#### initialize()
Initializes ApplicationInsights.

#### collect(eventName: string, payload: any)
Collects custom events, the payload is an object. I have enforced a standard for event names to end with the word EVENT.

#### trackComponent(ComponentToTrack: ComponentType): ComponentType
Tracks engagement time of a component. The engagement time is the time a user interacts with a component minus idle time. This will be useful when measuring usefulness of the Auth, Permissions, etc tabs.


MIchaelMainer:
Timu zingine hufanya nini [for AppInsights abstractions? I highly suggest that you take a look around for how other teams do this as they probably have considered things we will consider, and then other considerations we won't consider.

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ComponentType } from 'react';
+import ITelemetry from './ITelemetry';
+
+class Telemetry implements ITelemetry {
+  private appInsights: ApplicationInsights;
+  private config: any;
+
+  constructor() {
+    this.config = {
+      instrumentationKey: process.env.REACT_APP_INSTRUMENTATION_KEY,
```
thewahome: What would happen if the instrumentation key is not provided?

jobala: It won't send events to the AzureMonitor instance

src/telemetry/ITelemetry.ts
```diff
@@ -0,0 +1,7 @@
+import { ComponentType } from 'react';
+
+export default interface ITelemetry {
+  initialize(): void;
+  collect(eventName: string, payload: any): void;
```
MIchaelMainer: You have trackComponent and trackException. I propose trackEvent instead of *collect()*. What other types of data might we capture? trackMetrics?

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -3,6 +3,7 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators, Dispatch } from 'redux';
 
+import { RUN_QUERY_EVENT } from '../../../telemetry/event-types';
```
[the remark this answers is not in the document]
jobala: I will remove it.

src/telemetry/ITelemetry.ts
```diff
@@ -0,0 +1,7 @@
...
+  trackComponent(Component: ComponentType): ComponentType;
```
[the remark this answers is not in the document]
jobala: > Is every bit of telemetry we will collect associated with a React component?

Nope, some telemetry will be from user behaviour

src/index.tsx
```diff
@@ -24,6 +24,8 @@ import { readHistoryData } from './app/views/sidebar/history/history-utils';
 import messages from './messages';
 import { store } from './store';
 import './styles/index.scss';
+import { telemetry } from './telemetry';
```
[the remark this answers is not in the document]
jobala: And also, this telemetry class is meant to be used in GE v3 as well.

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,61 @@
+import { ReactPlugin, withAITracking } from '@microsoft/applicationinsights-react-js';
+import { ApplicationAnalytics, ApplicationInsights } from '@microsoft/applicationinsights-web';
+import { ComponentType } from 'react';
+import ITelemetry from './ITelemetry';
+
+class Telemetry implements ITelemetry {
```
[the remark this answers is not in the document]
jobala: Yes, I don't expect us to have other telemetry classes

MIchaelMainer:
Related to https://github.com/microsoftgraph/microsoft-graph-explorer-v2/issues/150

MIchaelMainer:
@bettirosengugi Which milestone should this be in?

src/index.tsx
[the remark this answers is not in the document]
MIchaelMainer: For Angular in GEv3, the logging pattern is to use a service. We should check with observability about the use of https://www.npmjs.com/package/@dagonmetric/ng-log-applicationinsights.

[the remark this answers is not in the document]
MIchaelMainer: Nevermind, there is this: https://github.com/microsoft/ApplicationInsights-JS/tree/master/extensions/applicationinsights-angular-js. I don't think we use much routing in GEv3 so it may make more sense to just use the base AppInsights library.

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,61 @@
...
+  private appInsights: ApplicationInsights;
+  private config: any;
+
+  constructor() {
+    const areWeInDev = process.env.NODE_ENV === 'development';
+
+    this.config = {
+      instrumentationKey: process.env.REACT_APP_INSTRUMENTATION_KEY,
+      disableExceptionTracking: true,
```
xiao-lix: Do you mind explain why is `disableExceptionTracking` set to true here? Are you disabling exception auto collection?

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!this.valid(eventName)) {
+      throw new Error('Invalid telemetry event name');
+    }
+
+    this.appInsights.trackEvent({ name: eventName }, payload);
+  }
+
+  public trackException(error: Error) {
+    const exception: any = { error: error.stack };
+    this.appInsights.trackException(exception);
```
xiao-lix: `trackException` takes `Error` as a param: https://github.com/microsoft/ApplicationInsights-JS#sending-telemetry-to-the-azure-portal and parse it into [ExceptionDetails ](https://github.com/microsoft/ApplicationInsights-JS/blob/18162ec13b07a4839b96e4642ad04b8f9d745ec4/shared/AppInsightsCommon/src/Telemetry/Exception.ts#L120) and put into `data`->`baseData`->`exceptions` bag, so no need to take stack out.

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  public trackReactComponent(ComponentToTrack: ComponentType): ComponentType {
+    const reactPlugin = new ReactPlugin();
```
xiao-lix: I think this function will create multiple reactPlugin and appInsightsAnalytics plugins? AppInsights doesn't support adding a new plugin after initialized. If it is known that React component exist at the time AppInsights is initialized, it is recommend to add react plugin in `config` -> `extensions` on AI initialization, instruction adding react plugin: https://github.com/microsoft/ApplicationInsights-JS/tree/master/extensions/applicationinsights-react-js#basic-usage

Here's a react demo using AI react plugin, hope it helps: https://github.com/Azure-Samples/application-insights-react-demo/blob/000009395cccdf60b0c867fc0a26d995be3ab4a5/src/TelemetryService.js#L29

src/telemetry/telemetry.ts
jobala: Yes, I am disabling exception auto collection.

PR #388, opened 2020-02-25
Task/enable telemetry in tst

jobala:
## Overview

1. Enable telemetry in -tst
2. Specify the source of a telemetry event ie GE v3 or GE v4


src/telemetry/telemetry.ts
```diff
@@ -9,11 +9,13 @@ class Telemetry implements ITelemetry {
   private reactPlugin: any;
 
   constructor() {
+    const { mscc } = (window as any);
```
thewahome: this value is never used. Do we need to declare it here?

src/telemetry/telemetry.ts
```diff
@@ -9,11 +9,13 @@ class Telemetry implements ITelemetry {
...
+
     this.reactPlugin  = new ReactPlugin();
     this.config = {
-      instrumentationKey: process.env.REACT_APP_INSTRUMENTATION_KEY,
+      instrumentationKey: this.getInstrumentationKey(),
       disableExceptionTracking: true,
-      disableTelemetry: this.areWeInDev() ? false : true,
+      disableTelemetry: this.getInstrumentationKey() ? false : true,
```
thewahome: From what I understand, we disable Telemetry when we do not find an instrumentation key. Right?
However it seems that even without a key in the env file, it does not disable telemetry.

I also noticed that if the key is empty, it still does not disable

jobala: That is interesting, following up on your DM to learn more about why telemetry is not disabled.

jobala: Resolving this.

We have to stop and start the server for it to pick up changes from the .env file

PR #397, opened 2020-03-02
Fix: Stop clearing response when editing url

### LINKED ISSUE #390: Bug: When editing a url that you had a response on already, GE deletes the response preview area
**Describe the bug**
When you want to edit a url that you had already queried and gotten a response, GE deletes the response preview area.

**To Reproduce**
1. Go to GE Preview
2. Type any query and run it
3. try editing the query
4. The response preview area is cleared

**Expected behavior**
The response area should stay populated with the previous result till you click the run button or press Enter key on your keyboard.

**Screenshots**
image

**Additional context**

thewahome:
## Overview

Solves the reason why editing the sample url clears the query response.
Fixes #390 

### Notes

N/A

## Testing Instructions

* Run a sample
* Start changing the sample url
* Notice that the response does not get deleted


src/app/services/reducers/query-runner-reducers.ts
```diff
@@ -1,17 +1,19 @@
...
+} from '../redux-constants';
 
 export function graphResponse(state = {}, action: IAction): any {
   switch (action.type) {
     case QUERY_GRAPH_SUCCESS:
       return action.response;
     case VIEW_HISTORY_ITEM_SUCCESS:
       return action.response;
-    case SET_SAMPLE_QUERY_SUCCESS:
+    case QUERY_GRAPH_RUNNING:
```
jobala: Please elaborate on why we made this change. I have not been able to figure out from going through the rest of the code.

thewahome: Once we are running a new query it removes the previous results.

I removed the SET_SAMPLE_QUERY_SUCCESS listener because it was the main cause of the bug

jobala: Why did we replace SET_SAMPLE_QUERY_SUCCESS with QUERY_GRAPH_RUNNING?

jobala:
Response body flickers when user edits url. Is this something you are okay with @bettirosengugi ?

src/app/services/reducers/query-runner-reducers.ts
thewahome: The replacement is because as a query is running, we do not need to see the response of the previously ran request.

thewahome: Notice that it is also at this action `QUERY_GRAPH_RUNNING` that we clear the response status

bettirosengugi:
> Response body flickers when user edits url. Is this something you are okay with @bettirosengugi ?

@jobala @thewahome Flickers? Why does it flicker?

The experience should be consistent to what we have in Classic, the response body stays as is when editing the url, no flickering, no loading, just same data/response body till you hit "Run Query"

src/app/services/reducers/query-runner-reducers.ts
jobala: For this use case. we'll be fine having neither `SET_SAMPLE_QUERY_SUCCESS` nor `QUERY_GRAPH_RUNNING` cases. So we can delete the `QUERY_GRAPH_RUNNING` case here.

thewahome:
> Response body flickers/flashes when a user edits the url

This only happens if you edit the version segment of the url. On other usual use cases of the input, the flickering was stopped. It was actually stopped by changing the action triggered: `SET_SAMPLE_QUERY_SUCCESS` to `QUERY_GRAPH_RUNNING`

Is this flicker a blocking effect @jobala / @bettirosengugi? As it is an intentional flicker now that we have the url as part of the object we set in the sampleQuery state

jobala:
@thewahome I consider the flicker a bug.

PR #401, opened 2020-03-03
Feature: clear history bulk

thewahome:
## Overview

Improves performance by changing how multiple history items are deleted


src/app/services/redux-constants.ts
```diff
@@ -31,3 +31,4 @@ export const FETCH_SCOPES_ERROR = 'SCOPES_FETCH_ERROR';
 export const GET_CONSENT_ERROR = 'GET_CONSENT_ERROR';
 export const GET_CONSENTED_SCOPES_SUCCESS = 'GET_CONSENTED_SCOPES_SUCCESS';
 export const SET_DEVX_API_URL_SUCCESS = 'SET_DEVX_API_URL_SUCCESS';
+export const REMOVE_HISTORY_ITEMS_BULK_SUCCESS = 'REMOVE_HISTORY_ITEMS_BULK_SUCCESS';
```
jobala: Let's rename this to `REMOVE_ALL_HISTORY_ITEMS_SUCCESS`

src/app/services/reducers/request-history-reducers.ts
```diff
@@ -17,6 +21,18 @@ export function history(state: any[] = [], action: IAction): any {
     case REMOVE_HISTORY_ITEM_SUCCESS:
       return state.filter(historyItem => historyItem !== action.response);
 
+    case REMOVE_HISTORY_ITEMS_BULK_SUCCESS:
+      const historyItemsToDelete = Object.values(action.response);
+      const newState: any = [];
+
+      state.forEach((historyItem: IHistoryItem) => {
```
jobala: This is a side effect. Reducers should be pure functions. Having impure reducers affects React's rendering and can cause hard to debug errors.

This also makes deletion of history items have a time complexity of `N squared`. N, when we generate the list of keys and N in `state.forEach`.

src/app/views/sidebar/history/history-utils.ts
```diff
@@ -24,6 +24,16 @@ export const removeHistoryData = async (historyItem: IHistoryItem) => {
...
+export const bulkRemoveHistoryData = async (listOfKeys: string[]) => {
+  historyStorage.iterate((value, key, iterationNumber) => {
+    if (listOfKeys.includes(key)) {
+      historyStorage.removeItem(key);
+    }
+  }).then(() => {
+    return true;
+  });
+};
+
```
jobala: This makes deletion of history a `N cubed` operation.

jobala:
@thewahome @bettirosengugi 

Should we have a dialog asking users if they really want to delete all the history data?

src/app/services/reducers/request-history-reducers.ts
```diff
@@ -17,6 +21,18 @@ export function history(state: any[] = [], action: IAction): any {
...
+        if (!historyItemsToDelete.includes(historyItem.createdAt)) {
+          newState.push(historyItem);
+        }
```
jobala: We can use a  one liner here
```
      return state.filter((historyItem: IHistoryItem) => !historyItemsToDelete.includes(historyItem.createdAt));
```

PR #461, opened 2020-03-26
Feature: Add icon on auth-required queries

### LINKED ISSUE #406: Add a Key icon next to Sample queries that require login and update the tooltips.
We need to display multiple tooltips to show on the sample queries,  to enable this, we need to do the following: 

1. Add the key icon next to all samples queries that require login before a user can run them
2. When you hover over the key icon, the tool tip should be "Login to run this request"
3. When you hover over the sample query, the name of the sample query should be shown in full in the tool tip and the request option e.g. "GET My Profile"
4. When you hover over the Documentation link the tooltip should be the documentation link.
image

thewahome:
## Overview
Fixes #406 

- [x] Add the key icon next to all samples queries that require login before a user can run them

- [x] When you hover over the key icon, the tooltip should be "Login to run this request"

- [x] When you hover over the sample query, the name of the sample query should be shown in full in the tooltip and the request option e.g. "GET My Profile"

- [x] When you hover over the Documentation link the tooltip should be the documentation link.

### Demo

image

### Notes

"Login to run this request" changes to the already localized "Sign In to try this sample"
A key icon is not available in [the list of fabric icons](https://developer.microsoft.com/en-us/fabric#/styles/web/icons) however a Lock icon makes a lot of sense in its place

## Testing Instructions

* Hover over previously mentioned elements to see the tooltips

### DIFF
--- a/src/app/views/sidebar/sample-queries/SampleQueries.tsx
+++ b/src/app/views/sidebar/sample-queries/SampleQueries.tsx
@@ -102,50 +102,55 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
 
   public renderItemColumn = (item: any, index: number | undefined, column: IColumn | undefined) => {
     const classes = classNames(this.props);
+    const {
+      tokenPresent,
+      intl: { messages },
+    }: any = this.props;
 
     if (column) {
       const queryContent = item[column.fieldName as keyof any] as string;
 
       switch (column.key) {
         case 'authRequiredIcon':
-          if (item.method !== 'GET') {
+          if (item.method !== 'GET' && !tokenPresent) {
+            const signInText = messages['Sign In to try this sample'];
             return <TooltipHost
+              tooltipProps={{
+                onRenderContent: () => <div style={{ paddingBottom: 3 }}>
+                  <FormattedMessage id={'Sign In to try this sample'} /></div>
+              }}
+              id={getId()}
+              calloutProps={{ gapSpace: 0 }}
+              styles={{ root: { display: 'inline-block' } }}
+            >
+              <IconButton
+                className={classes.docLink}
+                iconProps={{ iconName: 'Lock' }}
+                title={signInText}
+                ariaLabel={signInText}
+              />
+            </TooltipHost>;
+          }
+          return null;
+
+        case 'button':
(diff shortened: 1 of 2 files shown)


src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -100,41 +100,60 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
...
 
-      const toolTipContent = this.getToolTipContent(item, queryContent);
-
       switch (column.key) {
+        case 'authRequiredIcon':
+          if (item.method !== 'GET') {
+            return <TooltipHost
+            tooltipProps={{
+              onRenderContent: () => <div style={{ paddingBottom: 3 }}>
+                <FormattedMessage id={'Sign In to try this sample'} /></div>
```
ddyett: >sample [](start = 59, length = 6)

this is before your change but my only problem with this word is that if you do sign in and run this command, it's not really a sample.  You are modifying your actual data.

src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -100,41 +100,60 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
...
+            }}
+            id={getId()}
+            calloutProps={{ gapSpace: 0 }}
+            styles={{ root: { display: 'inline-block' } }}
+          >
+            <IconButton
+              className={classes.docLink}
+              iconProps={{ iconName: 'Lock' }}
+              title='Sign In to try this sample'
+              ariaLabel='Sign In to try this sample'
```
ddyett: >Sign In to try this sample [](start = 25, length = 26)

wouldn't this also need to be localized for an accessible reader?

src/app/views/sidebar/sample-queries/SampleQueries.tsx
thewahome: Should we change this to request instead?

ddyett: maybe but i'll leave to PM to decide if it should be changed at all.  It's pre-existing so I'm fine as is in this PR.

PR #480, opened 2020-04-20
Feature: autocomplete

### LINKED ISSUE #343: Integrate Autocomplete endpoint of GE API
It would be good to suggest all possible request urls when a user is typing the url. More to that, if a user doesn't know the url, having a request builder will help them create a url from clicking on the UI

Integration to the Autocomplete endpoint with autocomplete feature when you start typing a url on the query tab. 

Also, a UI request builder, that allows you select a resource and in it displays all the request urls available on that resource , and when you select any of the request urls, it's pasted on the url section and run. 

**Additional context**

[AB#4591](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4591)

thewahome:
## Overview
Integrates with the open api endpoint of the devx api
Fixes #343 [AB#4591](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4591)

### Tasks

- [x] Create redux service

- [x] Connect to available query parameters 

- [x] Create autocomplete component

- [x] Give autocomplete component all the functionalities of the url text field

- [x] Populate autocomplete component with available query parameter results

- [x] Enable drilling into the properties of the listed query parameters

- [x] Support requests other than GET requests

- [x] Support requests with IDs 

- [x] Connect to links

- [x] Enable keyboard navigation when selecting options provided

- [x] Show root entities when no query is passed

### Commands

- `esc` - dismisses suggestions
- `/` or `?` or `CTRL+.` or `CTRL+space` -  initiates autocomplete
- `tab` or `enter` - selects autocomplete option
- `up` or `down` - navigates through autocomplete options available


src/app/services/actions/autocomplete-action-creators.ts
```diff
@@ -0,0 +1,91 @@
...
+    try {
+      const response = await fetch(permissionsUrl, options);
+      if (response.ok) {
+        const autoCompleteOptions = await response.json();
+        const params = {
+          options: autoCompleteOptions,
+          url,
+          verb: sampleQuery.selectedVerb.toLowerCase()
+        };
+        const reduced = getReducedVersion(params);
```
jobala, line 49: The variable name `reduced` and the function name `getReducedVersion` do not communicate their intent. Let's use more descriptive names.

src/app/services/actions/autocomplete-action-creators.ts
```diff
@@ -0,0 +1,91 @@
...
+      throw (response);
+
+    } catch (error) {
+      return dispatch(fetchAutocompleteError(error));
+    }
+  };
+}
+function getReducedVersion(params: any) {
+  const { options, url, verb } = params;
+  const { paths } = options;
```
jobala, line 63: You can do this in a one liner

```const { options: { paths }, url, verb } = params

File: src/app/services/actions/autocomplete-action-creators.ts [hunk 3]
```diff
@@ -0,0 +1,91 @@
...
+  try {
+    const parameters: any[] = [];
+    let rootPath = url;
+    if (url.includes('me/drive')) {
+      rootPath = url.substring(3);
```
jobala, line 68: Why are we doings this only for the `me/drive` path?

thewahome: the me/drive path is the only one that doesn't follow the pattern in the url. It is a bug on the open api library that parses the data

src/app/services/actions/autocomplete-action-creators.ts
thewahome: good catch

src/app/services/actions/autocomplete-action-creators.ts
thewahome: nice

src/app/services/actions/autocomplete-action-creators.ts
thewahome: moving this to it's own file and renaming it parseOpenApiResponse

src/app/services/actions/autocomplete-action-creators.ts
```diff
@@ -0,0 +1,91 @@
...
+export function fetchAutoCompleteOptions(url: string): Function {
+  return async (dispatch: Function, getState: Function) => {
+    const devxApi = getState().devxApi;
+    const sampleQuery = getState().sampleQuery;
+    const headers = {
+      'Content-Type': 'application/json',
+    };
+
+    let permissionsUrl = `${devxApi}/openapi`;
+    permissionsUrl = `${permissionsUrl}?url=/${url}&style=geautocomplete&format=json`;
```
jobala, line 34: It is misleading to refer to this as the `permissionUrl` especially because we have a permissions tab.

src/app/services/actions/autocomplete-action-creators.ts
```diff
@@ -0,0 +1,91 @@
...
+        };
+        const reduced = getReducedVersion(params);
+
+        return dispatch(fetchAutocompleteSuccess(reduced));
+      }
+
+      throw (response);
+
+    } catch (error) {
+      return dispatch(fetchAutocompleteError(error));
```
jobala, line 57: Instead of throwing at line 53 then catching a few lines later, we can just dispatch if `!response.ok`

```js
if (!response.ok) {
  return dispatch(...)
}

...
return dispatch(fetchAutocompleteSuccess)
```

src/app/services/actions/autocomplete-action-creators.ts
```diff
@@ -0,0 +1,91 @@
...
+function getReducedVersion(params: any) {
+  const { options, url, verb } = params;
+  const { paths } = options;
+  try {
+    const parameters: any[] = [];
+    let rootPath = url;
+    if (url.includes('me/drive')) {
+      rootPath = url.substring(3);
+    }
+    const root = paths[`/${rootPath}`];
```
jobala, line 70: `root` might not be the right variable name here. It doesn't not communicate its intent. The data it holds appears to be the odata stuff for this particular url.

src/app/services/actions/autocomplete-action-creators.ts
```diff
@@ -0,0 +1,91 @@
...
+          const newLocal = {
+            name: param.name,
+            items: param.items.enum || null
+          };
+          parameters.push(newLocal);
+        }
+      });
+    }
+    const resp = { url, parameters, verb };
+    return resp;
```
jobala, line 85: You can do `return { url, parameters, verb }`

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -43,8 +44,8 @@ export class QueryRunner extends Component<
 
   private handleOnUrlChange = (newUrl = '') => {
     this.setState({ url: newUrl });
-
     this.changeUrlVersion(newUrl);
+    this.initialiseAutoComplete(newUrl);
```
jobala, line 48: This is misleading, it implies that we are initialising auto complete whenever the url changes. I suggest that we move the check we do to `handleOnUrlChange`.

```js
if (newUrlEndsWith?) {
  this.initialiseAutoComplete()
}
```

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -89,6 +90,16 @@ export class QueryRunner extends Component<
     }
   };
 
+  private initialiseAutoComplete(url: string) {
+    const lastCharacter = url.substring(url.length - 1);
+    if (lastCharacter === '?') {
```
jobala, line 95: Let us move this check to `handleOnUrlChange`

src/app/services/actions/autocomplete-action-creators.ts
thewahome: Covered in this [commit](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/pull/489/commits/48ee2872a62de115af7907d7d7c9253c360b4651)

src/app/services/actions/autocomplete-action-creators.ts
thewahome: Done in [this commit](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/pull/489/commits/dd23c2abcde083427c7343d84be6a46918e75b7e)

src/app/services/actions/autocomplete-action-creators.ts
thewahome: Done [here](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/pull/489/commits/dd23c2abcde083427c7343d84be6a46918e75b7e)

src/app/services/actions/autocomplete-action-creators.ts
thewahome: makes sense. Fixed this [here](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/pull/489/commits/8b38678b5e830b3bbf96df0d4191f6ab7f84e104)

src/app/services/actions/autocomplete-action-creators.ts
```diff
@@ -0,0 +1,58 @@
...
+export function fetchAutoCompleteOptions(url: string): Function {
+  return async (dispatch: Function, getState: Function) => {
+    const devxApi = getState().devxApi;
+    const sampleQuery = getState().sampleQuery;
+    const headers = {
+      'Content-Type': 'application/json',
+    };
+
+    let openApiUrl = `${devxApi}/openapi`;
+    openApiUrl = `${openApiUrl}?url=/${url}&style=geautocomplete&format=json`;
```
zengin, line 35: nit:
```suggestion
    let openApiUrl = `${devxApi}/openapi?url=/${url}&style=geautocomplete&format=json`;
```

src/app/services/reducers/autocomplete-reducer.ts
```diff
@@ -0,0 +1,33 @@
+import { IAction, IApiResponse } from '../../../types/action';
+import { AUTOCOMPLETE_FETCH_ERROR, AUTOCOMPLETE_FETCH_PENDING, AUTOCOMPLETE_FETCH_SUCCESS } from '../redux-constants';
+
+const initialState: IApiResponse = {
+  pending: false,
+  data: [],
+  error: null
+};
+
+export function autoComplete(state = initialState, action: IAction): any {
```
zengin, line 10: any chance return type `any` can be a well defined type?

src/app/utils/open-api-parser.ts
```diff
@@ -0,0 +1,50 @@
+export function parseOpenApiResponse(params: any) {
+  const { options: { paths }, url } = params;
```
zengin, line 2: Looks like this function expects params to be a well-defined type. I think, instead of using `any`, we should define the type and optionally destructure in the parameter list. Example: https://gist.github.com/rileyhilliard/da10fe11b68d1b7feb31ad119c59d516#file-correct-destructure-ts

The advantage of this will be that if someone is to call this function, they will know in which format their data should be. In the current state, they can pass in an object of any form, even though the first line in the function is expecting a well formed input.

src/app/utils/open-api-parser.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  try {
+    const parameters: any[] = [];
+    const requestUrl = Object.keys(paths)[0];
+    const verbs = Object.keys(paths[`${requestUrl}`]);
+    const pathValues: any = Object.values(paths)[0];
```
zengin, line 8: Typing suggestion from above continued: Without types, the caller is at least expected to read the destructuring here and it is hard to tell what qualifies as valid input at a glance.

We can use types in function signature to help callers in understanding the purpose of the function. Comments are useful if there are other expectations of input that can't be defined as part of the formal signature (i.e. types).

src/app/utils/open-api-parser.ts
```diff
@@ -0,0 +1,50 @@
...
+      });
+    });
+
+    return { url, parameters };
+  } catch (error) {
+    return { error };
+  }
+}
+
+function getVerbParameterValues(values: any) {
```
zengin, line 24: same here. The function assumes that `values` has a well defined type and we should reflect that in the input type instead of `any`.

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,344 @@
...
+    };
+  }
+
+  public setFocus() {
+    if (this.autoCompleteRef && this.autoCompleteRef.current) {
+      this.autoCompleteRef.current.focus();
+    }
+  }
+
+  public onChange = (e: any) => {
```
zengin, line 36: please use type `React.ChangeEvent` for `e` if possible.

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,344 @@
...
+    this.initialiseAutoComplete(userInput);
+  };
+
+  public onClick = (e: any) => {
+    const selected = e.currentTarget.innerText;
+    this.appendSuggestionToUrl(selected);
+  };
+
+  private initialiseAutoComplete = (url: string) => {
+    const lastCharacter = url.substring(url.length - 1);
```
zengin, line 58: nit:
```suggestion
    const lastCharacter = url.slice(-1);
```

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,344 @@
...
+
+      case KeyCodes.up:
+        if (activeSuggestion === 0) {
+          return;
+        }
+        this.setState({ activeSuggestion: activeSuggestion - 1 });
+        break;
+
+      case KeyCodes.down:
+        if (activeSuggestion - 1 === filteredSuggestions.length) {
```
zengin, line 110: I think this should be the other way around in order not to go out of bounds:
```suggestion
        if (activeSuggestion === filteredSuggestions.length - 1) {
```

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,344 @@
...
+
+          this.setState({
+            filteredSuggestions: suggestions,
+            suggestions,
+            showSuggestions: true,
+            compare: ''
+          });
+        }
+      }
+    }
```
zengin, line 142: please use "early return" for validation so that the code doesn't form arrowhead anti-pattern, e.g.:

```suggestion
    if (!autoCompleteOptions) {
      return;
    }
    const suggestions: string[] = [];
    const parameters = autoCompleteOptions.parameters;
    if (!parameters) {
      return;
    }
    const parametersWithVerb = parameters.find(parameter => parameter.verb === selectedVerb.toLowerCase());
    if (!parametersWithVerb) {
      return;
    }
    parametersWithVerb.links.forEach((value: string) => {
      suggestions.push(value);
    });
    this.setState({
      filteredSuggestions: suggestions,
      suggestions,
      showSuggestions: true,
      compare: ''
    });
  }
}
```

Related discussion: https://stackoverflow.com/questions/17804005/how-to-prevent-the-arrowhead-anti-pattern

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  public getQueryParameters = () => {
```
zengin, line 145: Function body is exactly the same as displayLinkOptions except for the part code pushes into suggestions. I think duplicated logic in these two functions can be abstracted out as functions:

```
getParametersWithVerb()
{
// code that returns parametersWithVerb
}

setSuggestions(suggestions: string[])
{
  this.setState({
    filteredSuggestions: suggestions,
    suggestions,
    showSuggestions: true,
    compare: ''
  });
}

public getQueryParameters = () => {
  const parametersWithVerb = getParametersWithVerb();
  if (!parametersWithVerb)
  {
    return;
  }
  setSuggestions(parametersWithVerb.values.map(value => value.name));
}

public displayLinkOptions = () => {
  const parametersWithVerb = getParametersWithVerb();
  if (!parametersWithVerb)
  {
    return;
  }
  setSuggestions(parametersWithVerb.links);
}
```

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,344 @@
...
+
+  public componentDidUpdate = (prevProps: IAutoCompleteProps) => {
+    if (prevProps.autoCompleteOptions !== this.props.autoCompleteOptions) {
+      if (this.props.autoCompleteOptions) {
+        this.performLocalSearch(this.props.sampleQuery.sampleUrl);
+      }
+    }
+  }
+
+  private getParameterEnums = (url: string) => {
```
zengin, line 176: I think the helper functions I suggested earlier would be useful to share code and shorten this function as well.

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,344 @@
...
+    const { autoCompleteOptions, sampleQuery: { selectedVerb } } = this.props;
+    if (!autoCompleteOptions) {
+      return;
+    }
+    const parameters = autoCompleteOptions.parameters;
+    const param = url.split('$').pop()!.split('=')[0];
```
zengin, line 182: This line should go into the `if (parameters)` block below where it is used, so that we avoid running the relatively expensive string operations if `parameters` is not valid.

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,344 @@
...
+            showSuggestions: true,
+            compare: ''
+          });
+        }
+      }
+    }
+  }
+
+  private filterSuggestions(userInput: string, previousUserInput: string, compare: string, suggestions: string[]) {
+    let compareString = userInput.replace(previousUserInput, '');
```
zengin, line 205: If `previousUserInput` is `a` and `userInput` is `aa`, `compareString` will be empty string as `replace` function replaces all instances of the first parameter. Is this an edge case that needs to be covered?

ddyett:
Bugs filed under area: autocomplete tag.

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,344 @@
...
+    } = this.state;
+
+    const { fetchingSuggestions, sampleQuery } = this.props;
+
+    const currentTheme = getTheme();
+    const suggestionClass: any = queryInputStyles(currentTheme).autoComplete.suggestions;
+    const suggestionOption: any = queryInputStyles(currentTheme).autoComplete.suggestionOption;
+    const activeSuggestionClass: any = queryInputStyles(currentTheme).autoComplete.suggestionActive;
+    const autoInput: any = queryInputStyles(currentTheme).autoComplete.input;
+    const suggestionTitle: any = queryInputStyles(currentTheme).autoComplete.suggestionTitle;
```
zengin, line 276: To avoid calling queryInputStyles 5 times in a row:
```suggestion
  const { suggestions : suggestionClass,
        suggestionOption,
        suggestionActive : activeSuggestionClass,
        input : autoInput,
        suggestionTitle } = queryInputStyles(currentTheme).autoComplete;
```

src/app/views/query-runner/query-input/QueryInput.styles.ts
```diff
@@ -0,0 +1,95 @@
...
+          cursor: 'pointer',
+          boxShadow: 'none',
+          marginTop: '0px',
+          marginRight: '0px',
+          marginBottom: '0px',
+          marginLeft: '0px',
+          paddingTop: '0px',
+          paddingRight: '32px',
+          paddingBottom: '0px',
+          paddingLeft: '12px',
```
zengin, line 68: General question on GE: Is there a reason why we don't use relative units (i.e. `rem`s) instead of pixels?

src/app/views/query-runner/query-input/QueryInput.styles.ts
```diff
@@ -0,0 +1,95 @@
...
+          lineHeight: '30px',
+          position: 'relative',
+          whiteSpace: 'nowrap',
+          textOverflow: 'ellipsis',
+          borderWidth: '1px',
+          borderStyle: 'solid',
+          overflow: 'hidden',
+          backgroundColor: theme.palette.neutralLight,
+          borderColor: theme.palette.neutralLight,
+          // borderBottom:
```
zengin, line 80: please remove dead code.

src/app/views/query-runner/query-input/QueryInput.styles.ts
thewahome: What is the benefit of relative units over pixels?

src/app/views/query-runner/query-input/util.ts
```diff
@@ -0,0 +1,25 @@
+export function cleanUpSelectedSuggestion(compare: string, userInput: string, selected: string) {
```
zengin, line 1: please add unit tests covering the inputs/outputs of this function.

src/app/views/query-runner/query-input/QueryInput.styles.ts
zengin: The main concern is accessibility when the user changes their default font size in their browser. But I checked after this comment, it looks like GE doesn't use relative units in other places as well, so addressing this shouldn't be in the scope of this PR.

More reading on this:
https://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css

A quick video demonstrating the accessibility benefits: https://www.youtube.com/watch?v=grMe8QREvA8

src/app/services/reducers/autocomplete-reducer.ts
```diff
@@ -0,0 +1,33 @@
...
+      return {
+        ...state,
+        pending: false,
+        data: null,
+        error: action.response
+      };
+    default:
+      return state;
+  }
+}
```
jobala, line 33: Add unit tests for this reducer

src/app/views/query-runner/query-input/AutoComplete.tsx
thewahome: Yes. Interesting that it actually comes up when you have a query with multiple query parameters with similar values. e.g
`https://graph.microsoft.com/v1.0/me/messages?$select=id,subject&orderby=su` fixed it with a slice function instead of the loop I was running.

bettirosengugi:
@thewahome adding new requests from our discussion today:
1. Sort the list of suggestions alphabetically.
2. When a user types something, the suggestions appear, but when you backspace, the suggestions disappear. Task here is to allow suggestions to appear at any point during the typing.
3. Fix this error that made GE freeze, it has occurred 3 times for me, I'm not sure the combinations to repro it but I captured this screenshot the last time it occurred. 
image

ddyett:
@bettirosengugi can we file separate issues for each helps keep track of issues renaming.  if you can also add the autocomplete label onto it.

zengin (approved):
Typescript changes look good to me. I will defer React parts to the React experts.

ElinorW (approved):

darrelmiller:
@thewahome @irvinesunday  I'm assuming at somepoint there was a decision made to drop the host from the URL.  The problem seems to be that in dropping the host we also dropped the version.

```
https://graphexplorerapi.azurewebsites.net/openapi?url=/&style=geautocomplete
```

There does not seem to be any way to get different autocomplete for V1 vs beta.  Can we put the full URL back?

irvinesunday:
> @thewahome @irvinesunday I'm assuming at somepoint there was a decision made to drop the host from the URL. The problem seems to be that in dropping the host we also dropped the version.
> 
> ```
> https://graphexplorerapi.azurewebsites.net/openapi?url=/&style=geautocomplete
> ```
> 
> There does not seem to be any way to get different autocomplete for V1 vs beta. Can we put the full URL back?

@darrelmiller GE can send DevX API the request: _/geautocomplete?url=/me&graphVersion=v1.0_ and DevX API should be able to handle fetching OpenAPI descriptions from either v1.0 or beta endpoints. There is already support for this in DevX API. We need to keep the url relative since we are now doing strict regex matching of the entire url.

PR #481, opened 2020-04-21
Task: Create autocomplete component

thewahome:
## Overview

Creates an autocomplete component that hosts the suggestions

## Demo

- Shows the query parameters for the query
image

- Shows the options within the selected parameter
image

## Notes
- Options are chosen using the tab key or clicking the choice
- Currently, this code fetches the query parameters when the '? - question mark' is hit. Then provides the properties of the query parameters when the '= - equal sign' is hit. 
- Having selected the property, hitting the ', - comma' shows the properties again
- Choosing the '& - ampersand' shows the query parameters again

## Cool stuff
- Search as you type and tab similar to a code editor's autocomplete
- ...

### DIFF
--- a/src/app/views/query-runner/query-input/AutoComplete.tsx
+++ b/src/app/views/query-runner/query-input/AutoComplete.tsx
@@ -18,7 +18,7 @@ class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
       filteredSuggestions: [],
       suggestions: [],
       showSuggestions: false,
-      userInput: this.props.sampleUrl,
+      userInput: this.props.sampleQuery.sampleUrl,
       compare: ''
     };
   }
@@ -27,31 +27,15 @@ class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
     const { suggestions, showSuggestions, userInput: previousUserInput, compare } = this.state;
     const userInput = e.target.value;
 
-    if (showSuggestions && suggestions.length) {
-      let compareString = userInput.replace(previousUserInput, '');
-      if (compare) {
-        compareString = compare + compareString;
-      }
-      // Filter our suggestions that don't contain the user's input
-      const filteredSuggestions = suggestions.filter(
-        (suggestion: any) => {
-          return suggestion.toLowerCase().indexOf(compareString.toLowerCase()) > -1;
-        }
-      );
-
-      this.setState({
-        filteredSuggestions,
-        compare: compareString
-      });
-    }
+    this.props.contentChanged(userInput);
 
     this.setState({
-      activeSuggestion: 0,
-      showSuggestions: true,
       userInput
     });
 
-    this.props.contentChanged(userInput);
+    if (showSuggestions && suggestions.length) {
+      this.filterSuggestions(userInput, previousUserInput, compare, suggestions);
+    }
     this.initialiseAutoComplete(userInput);
   };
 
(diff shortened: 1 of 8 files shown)


src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,264 @@
+import { getTheme, KeyCodes, Label, TextField } from 'office-ui-fabric-react';
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators, Dispatch } from 'redux';
+import { IAutoCompleteProps, IAutoCompleteState } from '../../../../types/auto-complete';
```
jobala, line 5: Separate local imports and  package imports with a new line.

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,264 @@
...
+import * as autoCompleteActionCreators from '../../../services/actions/autocomplete-action-creators';
+import { parseSampleUrl } from '../../../utils/sample-url-generation';
+import { queryInputStyles } from './QueryInput.styles';
+import { cleanUpSelectedSuggestion } from './util';
+
+class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
+
+  constructor(props: IAutoCompleteProps) {
+    super(props);
```
jobala, line 14: You don't need to call `super(props)`.

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,264 @@
...
+      activeSuggestion: 0,
+      filteredSuggestions: [],
+      showSuggestions: false,
+      userInput: selectedSuggestion,
+      compare: ''
+    });
+    this.props.contentChanged(selectedSuggestion);
+  }
+
+  public render() {
```
jobala, line 183: This function can be broken down to smaller function to help with readability.

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,264 @@
...
+      default:
+        break;
+    }
+  }
+
+  public onKeyDown = (e: any) => {
+    const { activeSuggestion, filteredSuggestions, showSuggestions } = this.state;
+
+    switch (e.keyCode) {
+      case KeyCodes.tab:
```
jobala, line 75: Pressing tab appends the suggestion to the url then transfers focus to the `Run Query` button. Can we maintain the focus in QueryInput?

src/app/views/query-runner/query-input/AutoComplete.tsx
thewahome: good catch

src/app/views/query-runner/query-input/AutoComplete.tsx
thewahome: interesting... why is that?

src/app/views/query-runner/query-input/AutoComplete.tsx
thewahome: Yes. I plan on implementing it.

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -0,0 +1,264 @@
...
+    }
+  }
+
+  private getParameterEnums = (url: string) => {
+    const param = url.split('$').pop()!.split('=')[0];
+    const { parameters } = this.props.autoCompleteOptions;
+    const section = parameters.find(k => k.name === `$${param}`);
+    const list: string[] = [];
+    if (section.items.length > 0) {
+      section.items.forEach((element: string) => {
```
Ogutu-Brian, line 131: I don't think this loop is necessary. What if you just had:
```
let list: string[] = section.items;
```
I think it would boost its runtime, what do you think?

PR #489, opened 2020-04-27
Task: support links

thewahome:
## Overview

Creates an open api parser for the results recevied from the devX api.
The parser enables the following tasks:

- [x] Support requests other than GET requests

- [x] Support requests with IDs without breaking

- [ ] Connect to links

Also enhances comments left in #480

### DIFF
--- a/src/app/services/actions/autocomplete-action-creators.ts
+++ b/src/app/services/actions/autocomplete-action-creators.ts
@@ -1,5 +1,6 @@
 import { IAction } from '../../../types/action';
 import { IRequestOptions } from '../../../types/request';
+import { parseOpenApiResponse } from '../../utils/open-api-parser';
 import { AUTOCOMPLETE_FETCH_ERROR, AUTOCOMPLETE_FETCH_PENDING, AUTOCOMPLETE_FETCH_SUCCESS } from '../redux-constants';
 
 export function fetchAutocompleteSuccess(response: object): IAction {
@@ -41,14 +42,13 @@ export function fetchAutoCompleteOptions(url: string): Function {
       const response = await fetch(permissionsUrl, options);
       if (response.ok) {
         const autoCompleteOptions = await response.json();
-        const params = {
+        const parameters = {
           options: autoCompleteOptions,
           url,
           verb: sampleQuery.selectedVerb.toLowerCase()
         };
-        const reduced = getReducedVersion(params);
-
-        return dispatch(fetchAutocompleteSuccess(reduced));
+        const parsedResponse = parseOpenApiResponse(parameters);
+        return dispatch(fetchAutocompleteSuccess(parsedResponse));
       }
 
       throw (response);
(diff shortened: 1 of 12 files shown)


src/app/services/actions/autocomplete-action-creators.ts
```diff
@@ -30,62 +31,28 @@ export function fetchAutoCompleteOptions(url: string): Function {
...
-        const params = {
+        const parameters = {
           options: autoCompleteOptions,
           url,
           verb: sampleQuery.selectedVerb.toLowerCase()
         };
-        const reduced = getReducedVersion(params);
-
-        return dispatch(fetchAutocompleteSuccess(reduced));
+        const parsedResponse = parseOpenApiResponse(parameters);
```
jobala, line 50: This can be an error object instead of `parsedResponse`. `parseOpenApiResponse` returns an error when it occurs. Did you mean to throw the error in `parseOpenApiResponse`?

thewahome: I should have actually... nice catch

src/app/services/actions/autocomplete-action-creators.ts
```diff
@@ -30,62 +31,28 @@ export function fetchAutoCompleteOptions(url: string): Function {
...
     try {
-      const response = await fetch(permissionsUrl, options);
+      const response = await fetch(openApiUrl, options);
       if (response.ok) {
         const autoCompleteOptions = await response.json();
-        const params = {
+        const parameters = {
           options: autoCompleteOptions,
           url,
           verb: sampleQuery.selectedVerb.toLowerCase()
```
ddyett, line 48: >verb: sampleQuery.selectedVerb.toLowerCase() [](start = 10, length = 44)

is this still used?  it looks like you don't pull it out in the parseOpenApiResponse anymore.

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -123,20 +132,32 @@ class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
...
-      });
+    if (parameters) {
+      const parametersWithVerb = parameters.find(parameter => parameter.verb === selectedVerb.toLowerCase());
+      if (parametersWithVerb) {
+        const section = parametersWithVerb.values.find((k: { name: string; }) => {
+          return k.name === `$${param}`;
+        });
+
+        const list: string[] = [];
+        if (section && section.items && section.items.length > 0) {
```
Ogutu-Brian, line 149: Length of `0` is already falsy so I don't think it's necessary to check if `section.items.length > 0`

I think you could just have:
```
if (section && section.items && section.items.length)
```

What do you think?

src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -123,20 +132,32 @@ class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
...
+          section.items.forEach((element: string) => {
```
Ogutu-Brian, line 150: From what I see here. I think this `forEach` loop is not necessary.
Since you are just copying items from `section.items` array, I think you could just change the `const` list variable to a `let` and then assign it to `section.items`  i.e
```
let list: string[] = [];

// the other code

list = section.items;
```
I think this will boost the time complexity here,
what do you think?

PR #497, opened 2020-04-30
Fix: provide account choice when logging in

### LINKED ISSUE #488: A user is not given the option to choose the account they want to sign in with
**Describe the bug**
The log in popup doesn't give you the option to choose the account you want to sign in with. It logs you in immediately with the account you had signed in with previously.

**Expected behavior**
A user should always be provided with the options of the accounts to sign in with once they click sign in.

Related [AB#4588](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4588)

thewahome:
## Overview

Provides account options when signing in

Fixes #488 

### Demo

N/A

### Notes

N/A

## Testing Instructions

1. Sign in and take note of the account you use
1. Sign out of that account
1. Attempt to sign in again, 
1. Notice that you will be provided with new accounts each time you try to repeat steps 1-3


src/app/services/graph-client/msal-service.ts
```diff
@@ -18,6 +18,7 @@ export function getSessionId() {
 export async function logIn(sessionId = ''): Promise<any> {
   const loginRequest: AuthenticationParameters = {
     scopes: defaultUserScopes,
+    prompt: 'select_account'
```
ddyett, line 21: if you only have one account does this cause you to have to select it each time?

thewahome: It is a safe bet... There could be a way to check whether there are other accounts stored in the cache... But I did not investigate this

PR #502, opened 2020-05-05
Fix: try it logout

### LINKED ISSUE #465: Logout on Try it not working
**Describe the bug**
Logout on Try it is not working. 

**To Reproduce**
1. Go to https://review.docs.microsoft.com/en-us/graph/api/user-list?view=graph-rest-1.0&tabs=http&branch=master
2. Click on Try it
3. Login, then logout
4. Reload the page after logout
5. Surprise! You are logged in still.

**Expected behavior**
If you logout and reload the page, you should be logged out. 

**Screenshots**

**Additional context**

thewahome:
## Overview 
Uses a workaround described in issue 1386  of [the msaljs repository](https://github.com/AzureAD/microsoft-authentication-library-for-js/).

Fixes #465 

### Demo

N/A 

### Notes

## Testing Instructions

* Change the graphExplorerMode to TryIt
* Sign in using your preferred account
* Sign out and notice a pop up and choose the account you want to sign out from
* Reload the page
* You would need to sign out again :-)

### DIFF
--- a/src/app/services/graph-client/msal-service.ts
+++ b/src/app/services/graph-client/msal-service.ts
@@ -72,6 +72,20 @@ export function logOut() {
   msalApplication.logout();
 }
 
+export function logOutPopUp() {
+  // @ts-ignore
+  msalApplication.clearCache();
+  // @ts-ignore
+  msalApplication.account = null;
+  // @ts-ignore
+  msalApplication.authorityInstance.resolveEndpointsAsync().then(authority => {
+    const urlNavigate = authority.EndSessionEndpoint
+        ? authority.EndSessionEndpoint
+        : `${msalApplication.authority}oauth2/v2.0/logout`;
+    (msalApplication as any).openPopup(urlNavigate, 'msal', 400, 600);
+  });
+}
+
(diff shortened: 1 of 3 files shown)


src/app/services/graph-client/msal-service.ts
```diff
@@ -72,6 +72,20 @@ export function logOut() {
   msalApplication.logout();
 }
 
+export function logOutPopUp() {
+  // @ts-ignore
```
jobala, line 76: Why are we ignoring Typescript here

thewahome: The methods I am accessing are private and surprisingly accessible as long as you ignore typescript.

It's an allowed workaround until the team releases a logoutPopUp implementation

thewahome:
Good looking out. I have fixed it

src/app/services/graph-client/msal-service.ts
jobala: You can add this as a comment, so that we know to stop ignoring Typescript in the future.

PR #528, opened 2020-05-07
Task: Refine permission column sizes

thewahome:
## Overview

Some columns were not visible

### Demo

image

image


thewahome:
> Let us move the status column a little further to the right.

This is going to be tricky because the screen sizes vary. Keeping it at a safe distance so that it is not hidden on different screens

Issue #533, opened 2020-05-07
Duplicate content type added for sendMail POST request -tst


MIchaelMainer:
**Describe the bug**
Graph Explorer -tst is adding duplicate content-type value to the **Content-Type** header. The sample has a default value content-type header value, and then it appears that Graph Explorer is adding a second header value.

**To Reproduce**
1. Choose the **send an email ** sample.
2. Click on **Run query**

**Expected behavior**
The request should be successful. The response should not show:

```json
{
    "error": {
        "code": "RequestBodyRead",
        "message": "The specified content type 'application/json, application/json' contains either no media type or more than one media type, which is not allowed. You must specify exactly one media type as the content type.",
        "innerError": {
            "request-id": "e42b6e3d-aec0-4e66-8146-9237bfd497d3",
            "date": "2020-05-07T23:22:42"
        }
    }
}
```

**Additional context**
If you remove the content-type header from the request headers, GE adds a content-type header. This appears to be a problem for all POST requests. It may be an issue for other HTTP methods.

PR #537, opened 2020-05-11
Fix: remove duplicate content type header

thewahome:
## Overview

Removes duplicate content-type headers passed in from the sample queries

### Demo

image

## Testing Instructions

* Select send an email from the sidebar
* Click run query
* Message will be sent

### DIFF
--- a/src/app/services/actions/query-action-creator-util.ts
+++ b/src/app/services/actions/query-action-creator-util.ts
@@ -24,12 +24,9 @@ export async function anonymousRequest(dispatch: Function, query: IQuery) {
   const escapedUrl = encodeURIComponent(query.sampleUrl);
   const graphUrl = `https://proxy.apisandbox.msdn.microsoft.com/svc?url=${escapedUrl}`;
   const sampleHeaders: any = {};
-
-  if (query.sampleHeaders) {
+  if (query.sampleHeaders && query.sampleHeaders.length > 0) {
     query.sampleHeaders.forEach(header => {
-      if (header.name !== '' && header.value !== '') {
-        sampleHeaders[header.name] = header.value;
-      }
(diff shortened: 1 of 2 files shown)


src/app/services/actions/query-action-creator-util.ts
```diff
@@ -103,9 +103,13 @@ const makeRequest = (httpVerb: string, scopes: string[]): Function => {
 
     if (query.sampleHeaders) {
       query.sampleHeaders.forEach(header => {
-        if (header.name !== '' && header.value !== '') {
-          sampleHeaders[header.name] = header.value;
+        if (header.name.toLowerCase() === 'content-type') {
```
MIchaelMainer, line 106: We should check that content-type value is set.

MIchaelMainer:
Fixes #533

src/app/services/actions/query-action-creator-util.ts
```diff
@@ -103,9 +103,13 @@ const makeRequest = (httpVerb: string, scopes: string[]): Function => {
...
+          return;
         }
+        if (header.name === '' && header.value === '') {
```
MIchaelMainer, line 109: No use in setting a header if either value is empty string. I think this should be an **or** condition.

src/app/services/actions/query-action-creator-util.ts
MIchaelMainer: Relying on the graph client to set the content type. Please add comment that content-type header is being set by Graph JavaScript client.

src/app/services/actions/query-action-creator-util.ts
thewahome: I actually think this check is redundant now

Issue #540, opened 2020-05-13
Error: Share link, navigate to page, then click sign in


jthake-msft:
**Describe the bug**

**To Reproduce**
1. Go to https://developer.microsoft.com/en-US/graph/graph-explorer/preview
2. Click on My Photo
3. CLick on Share
4. Copy URL
`https://developer.microsoft.com/en-US/graph/graph-explorer/preview?request=me&method=GET&version=v1.0&GraphUrl=https://graph.microsoft.com&requestBody=dW5kZWZpbmVk`
5. navigate to URL in tab
6. Click Sign in to Graph explore
7. Pick an acount
8. ERROR

`Sorry, but we’re having trouble signing you in.

AADSTS50011: The reply URL specified in the request does not match the reply URLs configured for the application: 'de8bc8b5-d9f9-48b1-a8ad-b748da725064'. More details: Reply address did not match because of case sensitivity.`

**Expected behavior**
Signs me in

** suggested fix **
Locale needs to be all lower case.

`https://developer.microsoft.com/en-US/graph/graph-explorer/preview?request=me&method=GET&version=v1.0&GraphUrl=https://graph.microsoft.com&requestBody=dW5kZWZpbmVk`
to
`https://developer.microsoft.com/en-us/graph/graph-explorer/preview?request=me&method=GET&version=v1.0&GraphUrl=https://graph.microsoft.com&requestBody=dW5kZWZpbmVk`

ddyett:
We should fix this.  It does happen with classic as well but it's a simple fix.

PR #551, opened 2020-05-18
Fix: lowercase locales

thewahome:
## Overview

Adds a lower-cased redirect url to the authentication request
Removes preview from the share query link and localizes it

Partially fixes missing url in the client id error in when there is a capitalised locale as mentioned in #540 

### Demo

```
https://developer.microsoft.com/es-es/graph/graph-explorer?request=me&method=GET&version=v1.0&GraphUrl=https://graph.microsoft.com&requestBody=dW5kZWZpbmVk
```

### DIFF
--- a/src/app/services/graph-client/msal-service.ts
+++ b/src/app/services/graph-client/msal-service.ts
@@ -20,6 +20,7 @@ export async function logIn(sessionId = ''): Promise<any> {
   const loginRequest: AuthenticationParameters = {
     scopes: defaultUserScopes,
     prompt: 'select_account',
+    redirectUri: window.location.href.toLowerCase(),
     extraQueryParameters: { mkt: geLocale }
   };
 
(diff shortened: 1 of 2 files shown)


src/app/services/graph-client/msal-service.ts
```diff
@@ -82,8 +83,8 @@ export function logOutPopUp() {
   // @ts-ignore
   msalApplication.authorityInstance.resolveEndpointsAsync().then(authority => {
     const urlNavigate = authority.EndSessionEndpoint
-        ? authority.EndSessionEndpoint
-        : `${msalApplication.authority}oauth2/v2.0/logout`;
+      ? authority.EndSessionEndpoint
+      : `${msalApplication.authority}oauth2/v2.0/logout`;
```
Ogutu-Brian, line 87: Does it mean that every time the version changes, this part of the code will be changed?
I think you could have the version (URL with version) defined somewhere as a constant such that if the version is updated later, a search need not be done to make updates but just a change in some one line. What do you think?

Issue #494, opened 2020-04-29
add tenant switch support in the sign-in area


baywet:
**Is your feature request related to a problem? Please describe.**
Today if I use the graph explorer to `GET /beta/servicePrincipals`, it will return only the service principals on the main tenant I'm part of. But I might have created other service principals on other tenants using the Azure Portal for example.
Whenever I'm writing an application, I can switch tenant by requesting different tokens to MIP. When using Azure I can switch tenant from the portal's UI. It'd be interesting to allow users to switch between the tenants they are part of so they can access different datasets.

**Describe the solution you'd like**
Next to the profile card, in the little gear, a "switch tenant" button that lists out the tenant you are part of. Click on any tenant and you are redirected to complete auth on that specific tenant instead.
![image](https://user-images.githubusercontent.com/7905502/80598927-1676ca80-89f8-11ea-8022-1a5fdbe09bea.png)

**Describe alternatives you've considered**
No alternatives possible using the graph explorer, using Postman collections or hand-written code is the only option at this point.

thewahome:
@baywet might #568 have solved this?

baywet:
@thewahome I believe it partially does, all what's missing now is to query the same API the azure portal uses to get the list of tenants I'm part of and add some kind of UI switch. Or at least make it discoverable so people know it exists.

ddyett:
@RabebOthmani do you think there's more that should be done on this issue or should it be closed?

lehrt:
Hi all, this issue is almost 4 years stale. Graph Explorer is no longer in active development and has moved to maintenance mode. No more major features like this are funded. If you are an Microsoft employee, you can reach out to internal Graph Explorer team about new tasks/feature requests. Closing this issue.

PR #568, opened 2020-05-30
Support for tenanted endpoint in Graph explorer.

agrabhi:
## Overview

Currently to go a tenant in Graph Explorer, we need to create a local account in tenant. After this change a `tenant` queryString parameter can be passed and then all idtoken and access token requests will go to that tenant.

Tenant parameter can be guid or the domain name for the tenant.

**sample request** 
https://developer.microsoft.com/en-us/graph/graph-explorer?tenant=abc.onmicrosoft.com
if no tenant is specified, the request goes to common endpoint. 

## RedirectUri notes
MSAL.JS documentation says that default redirect uri is window.location.href but taking a look at 
[this](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/d259ebab2ebc441a0eecce88381d932664c97fa1/lib/msal-browser/src/config/Configuration.ts#L65) and [this](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9274fac6d100a6300eb2faa4c94aa2431b1ca4b0/lib/msal-browser/src/utils/BrowserUtils.ts#L49) shows that its not complete location but only host part of it. 

So I had to modify the change which specifies the redirect url explicitly since the auth requests fail because of that. the redirect uri will come from MSAL like function now. This change was made by @thewahome in https://github.com/microsoftgraph/microsoft-graph-explorer-v4/pull/551 

## Test cases
1. SIgn in with and without tenant query string.
2. Consent to permissions works with and without tenant query string.

## Testing Instructions
If tenant is specified, user can login to their desired tenant
https://developer.microsoft.com/en-us/graph/graph-explorer?tenant=abc.onmicrosoft.com

### DIFF
--- a/src/app/services/graph-client/msal-service.ts
+++ b/src/app/services/graph-client/msal-service.ts
@@ -17,23 +17,23 @@ export function getSessionId() {
 }
 
 export async function logIn(sessionId = ''): Promise<any> {
-
+ 
   // support for tenanted endpoint
-  let params = new URLSearchParams(location.search);
-  let tenantId = params.get('tenantId');
+  const urlParams = new URLSearchParams(location.search);
+  let tenant = urlParams.get('tenant');
 
-  if (tenantId == null) {
-    tenantId = 'common';
+  if (tenant === null) {
+    tenant = 'common';
   }
 
   const loginRequest: AuthenticationParameters = {
     scopes: defaultUserScopes,
-    authority: `https://login.microsoftonline.com/${tenantId}/`,
+    authority: `https://login.microsoftonline.com/${tenant}/`,
     prompt: 'select_account',
     redirectUri: window.location.href.toLowerCase(),
     extraQueryParameters: { mkt: geLocale }
   };
-
+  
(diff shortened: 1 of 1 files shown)


darrelmiller:
@agrabhi Usually the process is to open an issue to identify the problem and discuss with the team how that feature might be implemented.  Using query parameters as a way to expose public features is  probably not the ideal solution.  We have also been discussing how to address the needs of being able to sign into the B2B tenants and provide access to sovereign clouds.  It would be nice to have a single feature that addresses all of these needs.

/cc @ddyett @bettirosengugi

jiasli:
> Using query parameters as a way to expose public features is probably not the ideal solution

I agree with @darrelmiller. Using `tenant` query parameter does not address feature requests such as 

- https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/2626
- https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/1982
- https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/494

PR #577, opened 2020-06-08
Fix: validate json

### LINKED ISSUE #557: Handle non wellformed JSON for request bodies
**Describe the bug**
Request bodies with trailing JSON commas cause Graph Explorer to perform in an unexpected way. JSON spec does not allow trailing commas.

**To Reproduce**
1. Go to Graph Explorer
2. Sign in.
3. Create a POST or PATCH request with a request body that contains a trailing comma.
4. See error in console.

**Expected behavior**
We show an error message stating that the supplied JSON is invalid. 

**Screenshots**
image

**Desktop **
 - Browser chrome, edge

**Additional context**
Incident 187938447

thewahome:
## Overview

Checks if the sample body is valid before running it. Incase it is not, it provides an error "Malformed JSON body" so that someone can see that there is a syntax error that needs to be fixed

Fixes #557 

### Demo

image

### Notes

The second part in the message that contains the actual syntax error is dynamically generated and thus it's localisation is not possible.

## Testing Instructions

* Log in so that you are able to send a post request
* Add an extra comma anywhere in the sample body and take note of it
* Hit enter / click run query
* Notice that an error appears saying that you have a syntax error and shows you where to find it

### DIFF
--- a/src/app/views/query-runner/QueryRunner.tsx
+++ b/src/app/views/query-runner/QueryRunner.tsx
@@ -68,7 +68,7 @@ export class QueryRunner extends Component<
 
   private handleOnRunQuery = () => {
     const { sampleBody } = this.state;
-    const { actions, sampleQuery,  } = this.props;
+    const { actions, sampleQuery, } = this.props;
     const { intl: { messages } }: any = this.props;
 
     if (sampleBody) {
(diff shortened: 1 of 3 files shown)


MIchaelMainer:
By the way, did you see this: https://github.com/Microsoft/monaco-json? It may provide additional value for this scenario.

PR #591, opened 2020-06-24
Task: reactivate app insights

thewahome:
## Overview

Reactivates app insights
Tracks when a query is ran

### Demo

N/A

### Notes

N/A

## Testing Instructions

* Open devtools
* Hit run query
* Notice a track network call

### DIFF
--- a/src/telemetry/telemetry.ts
+++ b/src/telemetry/telemetry.ts
@@ -57,7 +57,7 @@ class Telemetry implements ITelemetry {
   }
 
   private getInstrumentationKey() {
-    return process.env.REACT_APP_INSTRUMENTATION_KEY || '';
+    return (window as any).InstrumentationKey || process.env.REACT_APP_INSTRUMENTATION_KEY || '';
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/telemetry/telemetry.ts
```diff
@@ -57,7 +57,7 @@ class Telemetry implements ITelemetry {
   }
 
   private getInstrumentationKey() {
-    return (window as any).InstrumentationKey || '';
+    return process.env.REACT_APP_INSTRUMENTATION_KEY || '';
```
jobala, line 60: Update this to `(window as any).InstrumentationKey || process.env.REACT_APP_INSTRUMENTATION_KEY`

We always want to use the InstrumentationKey from the window object and only fallback to the key from env if the window object has no instrumentation key.

thewahome: We intend to use the key in the env since the key in the window will make us write to their instance instead of ours

jobala: The key is ours. We gave it to the portal team so that we can access it through the window object.

PR #619, opened 2020-06-26
Fix: Add tooltips and correct color contrast 

ElinorW:
## Overview

- Added tooltips to the 'minimize sidebar' and 'more actions' icons displayed to enhance keyboard navigation.

- Changed the color of the "patch" drop-down control to fit the color contrast ratio. 

### Demo

image

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/src/app/views/app-sections/AppTitle.tsx
+++ b/src/app/views/app-sections/AppTitle.tsx
@@ -10,13 +10,13 @@ export function appTitleDisplayOnFullScreen(
 
     return <div style={{ display: 'flex', width: '100%' }}>
       <TooltipHost
-        content='Minimise sidebar'
+        content='Minimize sidebar'
         id={getId()}
         calloutProps={{ gapSpace: 0 }}>
           <IconButton
             iconProps={{ iconName: 'GlobalNavButton' }}
             className={classes.sidebarToggle}
-            ariaLabel='Minimise sidebar'
+            ariaLabel='Minimize sidebar'
             onClick={() => toggleSidebar()} />
       </TooltipHost>
       <div className={classes.graphExplorerLabelContainer}>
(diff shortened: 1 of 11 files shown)


jobala:
I see two "more actions" tooltips when hovering over the gear icon.

image

src/app/views/app-sections/AppTitle.tsx
```diff
@@ -9,12 +9,16 @@ export function appTitleDisplayOnFullScreen(
...
 
     return <div style={{ display: 'flex', width: '100%' }}>
-      <IconButton
-        iconProps={{ iconName: 'GlobalNavButton' }}
-        className={classes.sidebarToggle}
-        title='Minimise sidebar'
-        ariaLabel='Minimise sidebar'
-        onClick={() => toggleSidebar()} />
+      <TooltipHost
+        content='Minimise sidebar'
```
ddyett, line 13: misspelt.  should be z

ddyett: also what about localization here?

jobala:
@ddyett @bettirosengugi @ElinorW @thewahome 

I am interested in knowing what you guys think about replacing the **settings** icon with a **context menu** icon. The settings icon is used for updating  user settings or profile in apps. The **context menu** icon is used for showing more actions. 

**Sample Context Menu Icon**
image

src/app/views/app-sections/AppTitle.tsx
jobala: > misspelt. should be z

@ddyett depends on which english pronunciation you're using :alien:

ddyett: all the more reason for localization.  since the most popular page is en-US probably should be z.

ddyett:
not sure how i feel about the icon.  there are some settings like actions and some more actions like actions. i'd have to think about it.

jobala:
@ddyett these are a closer representation of how the context-menu icons look like in uifabric-react
image

bettirosengugi:
> @ddyett these are a closer representation of how the context-menu icons look like in uifabric-react
> image

This is a good call out @jobala We got this feedback from Frederick as well. @ddyett may be we can have this as our first case for A/B testing?

PR #631, opened 2020-07-07
Task: support links

thewahome:
## Overview

Adds support for the links provided by the open api description

### Demo

image

### Notes

Keyboard navigation is flimsy at the moment. Use your mouse to select the options that you want

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/src/app/utils/open-api-parser.ts
+++ b/src/app/utils/open-api-parser.ts
@@ -44,9 +44,7 @@ function getLinkValues(values: any) {
     if (responsesAtIndex200 && responsesAtIndex200.links) {
       const links = Object.keys(responsesAtIndex200.links);
       return links;
-    } else {
-      return null;
     }
   }
-  return null;
+  return [];
 }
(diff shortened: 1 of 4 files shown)


jobala:
@thewahome, we should track flimsy keyboard navigation so that it is not lost.

src/app/utils/open-api-parser.ts
```diff
@@ -34,4 +35,16 @@ function getVerbParameterValues(values: any) {
     });
   }
   return parameterValues;
+}
+
+function getLinkValues(values: any) {
```
jobala, line 40: Do we know the type for values?

thewahome:
> @thewahome, we should track flimsy keyboard navigation so that it is not lost.

It's hand;ed in [this PR](github.com/microsoftgraph/microsoft-graph-explorer-v4/pull/632)

src/app/utils/open-api-parser.ts
thewahome: No

PR #632, opened 2020-07-08
Task: improve autocomplete usability

thewahome:
## Overview

Changes autocomplete suggestion selection to the 'enter' key
Runs query on enter key press when no suggestion is showing
Fixes usability issues when selecting suggestions after filtering as you type
Increases the size of the autocomplete text field to cater for long texts

### Demo

image

### Notes

N/A

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/src/app/views/query-runner/query-input/AutoComplete.tsx
+++ b/src/app/views/query-runner/query-input/AutoComplete.tsx
@@ -317,6 +317,7 @@ class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
 function mapStateToProps(state: any) {
   return {
     sampleQuery: state.sampleQuery,
+    appTheme: state.theme,
     autoCompleteOptions: state.autoComplete.data,
     fetchingSuggestions: state.autoComplete.pending
   };
--- a/src/app/views/query-runner/query-input/AutoComplete.tsx
+++ b/src/app/views/query-runner/query-input/AutoComplete.tsx
@@ -1,4 +1,4 @@
-import { getTheme, KeyCodes, Label, TextField } from 'office-ui-fabric-react';
+import { getTheme, ITextField, KeyCodes, Label, TextField } from 'office-ui-fabric-react';
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators, Dispatch } from 'redux';
@@ -10,7 +10,7 @@ import { queryInputStyles } from './QueryInput.styles';
 import { cleanUpSelectedSuggestion } from './util';
 
 class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
-  private autoCompleteRef: any;
+  private autoCompleteRef: React.RefObject<ITextField>;
 
   constructor(props: IAutoCompleteProps) {
     super(props);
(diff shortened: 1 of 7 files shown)


src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -10,10 +10,13 @@ import { queryInputStyles } from './QueryInput.styles';
 import { cleanUpSelectedSuggestion } from './util';
 
 class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
+  private autoCompleteRef: any;
```
jobala, line 13: I think it's possible to use a predefined type from React.

thewahome:
@jobala I changed the atocomplete suggestion selection to the 'enter' key as mentioned in the comments


The spaces are reduced to make sure the input can take longer urls

jobala:
@thewahome I am not sure what he comment  `I changed the atocomplete suggestion selection to the 'enter' key as mentioned in the comments` is a response to.

> The spaces are reduced to make sure the input can take longer urls

I get this but I the spaces are not reduced. There are no spaces between the controls.

darrelmiller:
I noticed a couple of issues:
- When showing the list of query parameters, the list includes path parameters.  Only parameters where the "in" property is query should be included in the dropdown
- When entering slash after the version, I am not seeing a drop-down for root entities
- It would be really nice if when pressing the slash prior to a path parameter that we could insert a placeholder for the path parameter with it highlighted so if someone continues typing it just overwrites.  This is not a requirement, just a "would be cool if".

ddyett:
structure here is a little nested as this branch goes into another dev branch before the feature branch before the dev branch.  This makes it a little hard to test changes.  Any reason you don't just use one branch under the features branch then do code reviews for changes into that branch.

i'm noticing on this branch keyboard issues where it doesn't scroll properly so focus gets lost. There are also issues where the focus in the uri box goes to the start at point and to the end at others so if i continue typing it sometimes types at the beginning.

Do you want these filed as issues against the feature branch or should these be addressed in this PR.

thewahome:
> structure here is a little nested as this branch goes into another dev branch before the feature branch before the dev branch. This makes it a little hard to test changes. Any reason you don't just use one branch under the features branch then do code reviews for changes into that branch.

@ddyett having smaller PRs makes it easier for people to review within context and for me to fix related things together. This particular branch is related to keyboard / navigation / selection issues that affect usability of the component

src/app/views/query-runner/query-input/util.ts
```diff
@@ -1,24 +1,24 @@
...
-  }
\ No newline at end of file
+    const availableSignPrefixes = ['?', ',', '=', '&', '/'];
+    let entryToRemove = '';
+    let activeSign = '/';
+    availableSignPrefixes.forEach(sign => {
+      const compareStringWithSign = `${sign + compare}`;
+      if (userInput.includes(compareStringWithSign)) {
+        entryToRemove = compareStringWithSign;
+        activeSign = sign;
```
ddyett, line 17: if there's a match here why do you need to continue looping through the other combinations?

thewahome: I see. It makes sense to exit the loop once a match is found

PR #635, opened 2020-07-09
Task: change header text

ElinorW:
## Overview

- Changed the header text on the request and response area to make it more responsive and enable easy navigation for low vision users when the page is resized to 200%.

### Demo
request area with page resized to 200%
image

response area, with page resized to 200%
image

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description


ddyett:
what happens when above 200%?

PR #636, opened 2020-07-13
Fix: failing create teams request

thewahome:
## Overview

Running the ```create a team``` query breaks the app like so
image

The characters previously unsupported when hashEncoding are now supported.

### Notes

## Testing Instructions

* Log in
* Set to POST and run ``` https://graph.microsoft.com/beta/teams ``` adding the request body described in the [docs here](https://docs.microsoft.com/en-us/graph/api/team-post?view=graph-rest-beta&tabs=http#example-6-create-a-team-with-a-non-standard-base-template-type)
* Notice the browser does not break 💯

### DIFF
--- a/src/app/views/common/share.ts
+++ b/src/app/views/common/share.ts
@@ -27,23 +27,23 @@ export const createShareLink = (sampleQuery: IQuery, authenticated?: boolean): s
 
   if (sampleBody && Object.keys(sampleBody).length > 0) {
     const requestBody = hashEncode(JSON.stringify(sampleBody));
-    shareLink = `&requestBody=${requestBody}`;
+    shareLink += `&requestBody=${requestBody}`;
   }
 
   if (sampleHeaders && sampleHeaders.length > 0) {
     const headers = hashEncode(JSON.stringify(sampleHeaders));
-    shareLink = `${shareLink}&headers=${headers}`;
+    shareLink += `${shareLink}&headers=${headers}`;
   }
 
   if (authenticated) {
     const sessionId = getSessionId();
     if (sessionId) {
-      shareLink = `${shareLink}&sid=${sessionId}`;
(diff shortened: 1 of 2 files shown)


src/app/views/common/share.ts
```diff
@@ -27,23 +27,23 @@ export const createShareLink = (sampleQuery: IQuery, authenticated?: boolean): s
...
   if (sampleBody && Object.keys(sampleBody).length > 0) {
     const requestBody = hashEncode(JSON.stringify(sampleBody));
-    shareLink = `&requestBody=${requestBody}`;
+    shareLink += `&requestBody=${requestBody}`;
   }
 
   if (sampleHeaders && sampleHeaders.length > 0) {
     const headers = hashEncode(JSON.stringify(sampleHeaders));
-    shareLink = `${shareLink}&headers=${headers}`;
+    shareLink += `${shareLink}&headers=${headers}`;
```
ddyett, line 35: shouldn't this be removed or stay as before?

src/app/views/query-runner/QueryInput.tsx
```diff
@@ -66,7 +66,7 @@ export class QueryInput extends Component<IQueryInputProps, any> {
       background: getStyleFor(selectedVerb),
     };
 
-    const httpMethodsToDisplay = (mode === Mode.TryIt && !authenticated ) ? [httpMethods[0]] : httpMethods;
+    const httpMethodsToDisplay = (mode === Mode.TryIt || !authenticated) ? [httpMethods[0]] : httpMethods;
```
ddyett, line 69: probably correct but we should consider if users may be confused minus text to say they need to sign in.

src/app/views/common/share.ts
thewahome: It should actually stay as before

PR #637, opened 2020-07-15
Fix: image content types

### LINKED ISSUE #634: Graph Explorer does not render image on beta/me/photo/$value for the fallback case
**Describe the bug**
In /beta, there are cases where Microsoft Graph will not find the user's photo (in Exchange) and will fallback to AAD.  This is documented here: https://docs.microsoft.com/en-us/graph/api/profilephoto-get?view=graph-rest-beta
In the case where Graph does fetch the photo from AAD, the photo is fetched properly - however it's not rendered in Graph Explorer.  The same is not true in POSTMan, where the photo IS rendered correctly

**To Reproduce**
1. Create a test user without an Exchange license/mailbox.  Upload a photo using https://docs.microsoft.com/en-us/graph/api/profilephoto-update?view=graph-rest-beta&tabs=http.  This will put a photo into AAD. 
2. Using POSTMan (and User.Read permission), sign in as that user, and call `GET https://graph.microsoft.com/beta/me/photo/$value`.  The response contains the image that was uploaded
3. Try the same in Graph Explorer

**Expected behavior**
Image/photo is rendered

**Actual behavior**
200 OK but with an empty {} JSON response.

**Additional context**
The only difference I see between photos returned from Exchange (which renders) and AAD (which doesn't) is the Content-Type value.  From exchange the value is "image/jpeg" and from AAD it's "image/**J**peg" - capital J.  Guessing that GE might be case sensitive here and not recognizing this content type, and it falls back to a default of "application/json"

thewahome:
## Overview

Fixes #634 

### Demo

image

### Notes

Creates an image url from all responses with a content type that includes images

## Testing Instructions

* Log in to GE
* Run `https://graph.microsoft.com/beta/me/photo/$value`
* Notice your image is presented

### DIFF
--- a/src/app/services/actions/query-action-creator-util.ts
+++ b/src/app/services/actions/query-action-creator-util.ts
@@ -53,7 +53,7 @@ export function isImageResponse(contentType: string) {
   if (!contentType) { return false; }
   return (
     contentType === 'application/octet-stream' ||
-    contentType.substr(0, 6) === 'image/'
+    contentType.includes('image/')
   );
 }
 
--- a/src/app/services/actions/query-action-creators.ts
+++ b/src/app/services/actions/query-action-creators.ts
@@ -1,9 +1,13 @@
 import { MessageBarType } from 'office-ui-fabric-react';
+
 import { ContentType } from '../../../types/enums';
 import { IHistoryItem } from '../../../types/history';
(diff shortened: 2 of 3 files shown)


ddyett:
not sure the validation steps are correct based on the issue but this looks like it should address it.

PR #644, opened 2020-07-20
Fix: screen reader announces search results status message

ElinorW:
## Overview

Added announced component which enables a screen reader to announce the 'Search sample queries' and 'Search permissions' (search result) information.

## Testing Instructions
* Use Microsoft Edge browser
* Ensure you have a screen reader turned on (e.g Windows Narrator)

1. Navigate to the 'Sample Queries' button by tab key and select it.
2. Navigate 'Sample Queries' search edit field and type 'Retrieve'.
3. Verify the screen reader's behavior. The screen reader announces search results after every letter typed.


src/app/views/query-runner/request/permissions/PanelList.tsx
```diff
@@ -37,6 +40,7 @@ const PanelList = ({ messages, permissions,
           searchValueChanged(event, newValue)}
         styles={{ field: { paddingLeft: 10 } }}
       />
+      <Announced message={`${permissions.length} search results available.`}/>
```
ddyett, line 43: string localization

src/app/views/query-runner/request/permissions/PanelList.tsx
```diff
@@ -1,4 +1,7 @@
-import { DetailsList, DetailsListLayoutMode, IColumn, Label, SearchBox, SelectionMode } from 'office-ui-fabric-react';
+import {
+  Announced, DetailsList, DetailsListLayoutMode, IColumn,
```
jobala, line 2: Place each imported class on a newline.

```
import {
Announced,
DetailList,
...
} from '..'
```

src/app/views/query-runner/request/permissions/PanelList.tsx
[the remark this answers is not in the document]
ElinorW: yeah it does

ElinorW: Since it's only a message that is being read, I tested this out with a Spanish narrator, and the narrator translates it itself

PR #645, opened 2020-07-21
Fix: Add aria-labels and correct colour contrast

### LINKED ISSUE #495: Accessibility issue on the tool-tip on the persona card
**Describe the bug**
The persona card's tooltip on the secondary text has a colour accessibility problem

**Screenshots**
image

ElinorW:
## Overview

- Add aria-labels for screen reader support.
- Altering the secondary text color on the persona card's tooltip to fit the acceptable color contrast ratio

Fixes #495

### DIFF
--- a/src/app/views/query-response/QueryResponse.tsx
+++ b/src/app/views/query-response/QueryResponse.tsx
@@ -91,9 +91,12 @@ class QueryResponse extends Component<IQueryResponseProps, IQueryResponseState> 
                 calloutProps={{ gapSpace: 0 }}
                 styles={{ root: { display: 'inline-block' } }}
               >
-                <IconButton onClick={this.handleShareQuery} className='share-query-btn' iconProps={{
-                  iconName: 'Share'
-                }} />
+                <IconButton
+                  onClick={this.handleShareQuery}
+                  className='share-query-btn'
+                  iconProps={{ iconName: 'Share'}}
+                  aria-label={'Share'}
+                />
               </TooltipHost>
             </div>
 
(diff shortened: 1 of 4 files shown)


src/app/views/query-response/QueryResponse.tsx
```diff
@@ -108,9 +111,12 @@ class QueryResponse extends Component<IQueryResponseProps, IQueryResponseState>
...
                 styles={{ root: { display: 'inline-block' } }}
               >
-                <IconButton onClick={this.toggleModal} className='share-query-btn' iconProps={{
-                  iconName: 'MiniExpandMirrored'
-                }} />
+                <IconButton
+                  onClick={this.toggleModal}
+                  className='share-query-btn'
+                  iconProps={{ iconName: 'MiniExpandMirrored'}}
+                  aria-label={'Expand response'}
```
thewahome, line 118: I have noticed that you are using aria-label and ariaLabel interchangeably. Do they both work? If so, we can stick to one (ariaLabel)

ElinorW: Yeah while reading the Fluent UI docs, under some components the say that the 'ariaLabel' is obsolete, so yeah I definitely think we should stick to aria-label

src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -330,6 +330,7 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
           placeholder={messages['Search sample queries']}
           onChange={this.searchValueChanged}
           styles={{ field: { paddingLeft: 10 } }}
+          ariaLabel={'Search'}
```
thewahome, line 333: this changes to aria-label?

ElinorW: Yes

PR #672, opened 2020-08-20
Telemetry for Graph Explorer usage

millicentachieng:
## Overview
Capture Graph Explorer usage data

**List of Events**
-	BUTTON_CLICK_EVENT
-	TAB_CLICK_EVENT
-	LINK_CLICK_EVENT
-	DROPDOWN_CHANGE_EVENT
-	LISTITEM_CLICK_EVENT

**Custom Properties**
Varies with event and component
-  ComponentName e.g. History Tab
- SampleName
- SelectedVersion

etc


src/telemetry/event-types.ts
```diff
@@ -1,3 +1,8 @@
 // Event names MUST end with the word EVENT
 
-export const RUN_QUERY_EVENT = 'RUN_QUERY_EVENT';
\ No newline at end of file
+export const RUN_QUERY_EVENT = 'RUN_QUERY_EVENT';
```
ddyett, line 3: is the plan to get rid of this one in a later PR?

src/app/views/sidebar/history/History.tsx
```diff
@@ -39,7 +41,8 @@ export class History extends Component<IHistoryProps, any> {
 
 
   public componentDidMount = () => {
-    this.generateGroupedList(this.props.history);
+      this.generateGroupedList(this.props.history);
+      telemetry.trackEvent(TAB_CLICK_EVENT, { ComponentName: "History Tab" });
```
ddyett, line 45: technically this is when the component is coming up not when the tab is clicked.  any reason this wasn't added in the onclick event?

src/app/views/sidebar/history/History.tsx
```diff
@@ -364,19 +373,22 @@ export class History extends Component<IHistoryProps, any> {
...
   }
 
   private deleteQuery = async (query: IHistoryItem) => {
-    const { actions } = this.props;
+    const { actions } = this.props;     
     if (actions) {
-      actions.removeHistoryItem(query);
+        actions.removeHistoryItem(query);
     }
+    this.trackHistoryItemEvent(BUTTON_CLICK_EVENT, 'Delete History Item Button', query);
```
ddyett, line 391: are these names easy to query on?

src/app/views/sidebar/history/History.tsx
```diff
@@ -406,6 +418,18 @@ export class History extends Component<IHistoryProps, any> {
...
   }
 
+  private trackHistoryItemEvent = (eventName: string, componentName: string, 
+      query: IHistoryItem) => {
+    telemetry.trackEvent(
+      eventName,
+      {
+        ComponentName: componentName,
+        ItemIndex: query.index,
+        QueryUrl: query.url,
```
ddyett, line 428: i think we should remove the url because it can contain user id information in some urls.  we'd need to strip it out but i don't think we necessarily need it so we can just pull it for now.

src/app/services/actions/query-action-creators.ts
```diff
@@ -96,6 +96,7 @@ async function createHistory(response: Response, respHeaders: any, query: IQuery
 
 
   const historyItem: IHistoryItem = {
+    index: 0,
```
ddyett, line 99: should this be initialized to -1 instead since 0 is a valid index.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
```diff
@@ -21,6 +24,12 @@ class AdaptiveCard extends Component<IAdaptiveCardProps> {
     if (hostConfig) {
       this.adaptiveCard.hostConfig = new AdaptiveCardsAPI.HostConfig(hostConfig);
     }
+    telemetry.trackEvent(TAB_CLICK_EVENT, 
+     {
+       ComponentName: 'Adaptive Cards Tab',
+       QueryUrl: sampleQuery ? sampleQuery.sampleUrl : null,
```
MIchaelMainer, line 30: Does sampleQuery capture what is in the URL text box? If so, we can't capture this as it can contain information that can identify a user.

ddyett, line 30: is there any user entered information such as userid or other ids

MIchaelMainer: The best that we can capture is the entity set / singleton being accessed. If we were to do that, I'd like for us to add code to validate that we only capture the entity set / singleton which is the first segment after version. Anything after this segment can be identifiable information.

src/app/views/query-response/snippets/snippets-helper.tsx
```diff
@@ -7,23 +7,37 @@ import { getSnippet } from '../../../services/actions/snippet-action-creator';
...
       key={language}
       headerText={language}
     >
-      <Snippet language={language} />
+      <Snippet language={language} sampleQuery={sampleQuery} />
     </PivotItem>
   ));
 }
 
+
```
ddyett, line 37: NIT: clean up spaces.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
MIchaelMainer: This comment is applicable to all places where we capture query URL.

src/app/views/query-response/snippets/snippets-helper.tsx
```diff
@@ -7,23 +7,37 @@ import { getSnippet } from '../../../services/actions/snippet-action-creator';
...
   language: string;
+  sampleQuery: IQuery
 }
 
 export function renderSnippets(supportedLanguages: string[]) {
+  const sampleQuery = useSelector((state: any) => state.sampleQuery, shallowEqual);
+  telemetry.trackEvent(TAB_CLICK_EVENT, 
+   {
+     ComponentName: 'Code Snippets Tab',
+     QueryUrl: sampleQuery.sampleUrl,
```
ddyett, line 24: verify there's no user id information.

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -107,6 +107,11 @@ export class QueryRunner extends Component<
         sampleUrl,
         selectedVersion: queryVersion
       });
+      telemetry.trackEvent(DROPDOWN_CHANGE_EVENT, 
+       { 
+         ComponentName: 'Version Change Dropdown', 
+         SelectedVersion: queryVersion
```
ddyett, line 113: there's only two options but but we may want to consider capturing the old value and new value.

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -5,7 +5,7 @@ import { connect } from 'react-redux';
 import { bindActionCreators, Dispatch } from 'redux';
 
 import { telemetry } from '../../../telemetry';
-import { RUN_QUERY_EVENT } from '../../../telemetry/event-types';
+import { RUN_QUERY_EVENT, DROPDOWN_CHANGE_EVENT } from '../../../telemetry/event-types';
```
ddyett, line 8: does not look like this is used in this file.

ddyett:
we should also check the RUN_QUERY_EVENT and make sure it is not transmitting the url with ids.  we need some code to strip them out.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
thewahome: We can get the request url that doesn't have the user information. There is a function that can strip out the rest of the info

src/app/views/query-response/snippets/snippets-helper.tsx
```diff
@@ -32,8 +46,6 @@ function Snippet(props: ISnippetProps) {
    */
   language = language.toLowerCase();
 
-
-  const sampleQuery = useSelector((state: any) => state.sampleQuery, shallowEqual);
```
thewahome, line 36: Why did we remove this and pass it in through the props?

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
thewahome: The function [parseSampleUrl](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/blob/4d9dc4cfbf036f50518210e507fe5762a73450a5/src/app/utils/sample-url-generation.ts#L3) can get the request url and we can leave out the search parameters. However, we can add a new function that can substitute personal information e.g an email, id with a token

src/app/views/query-response/snippets/snippets-helper.tsx
```diff
@@ -7,23 +7,37 @@ import { getSnippet } from '../../../services/actions/snippet-action-creator';
 import { Monaco } from '../../common';
 import { genericCopy } from '../../common/copy';
 
+import { telemetry } from '../../../../telemetry';
+import { BUTTON_CLICK_EVENT, TAB_CLICK_EVENT } from '../../../../telemetry/event-types';
+import { IQuery } from '../../../../types/query-runner';
+
 interface ISnippetProps {
   language: string;
+  sampleQuery: IQuery;
```
thewahome, line 16: The sample query is already present as a redux property in the snippets component. We do not need to pass it in as a property

src/app/views/sidebar/history/History.tsx
millicentachieng: I found out I could use the onLinkClick callback from IPivotProps interface.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
MIchaelMainer: We need to remove more than just query parameters (I assume that is what you mean by search parameters). We need to remove everything after the entityset/singleton.

https://graph.microsoft.com/v1.0/{entityset|singleton}

src/app/views/sidebar/history/History.tsx
millicentachieng: Yes, I'll share the queries on the design doc.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
millicentachieng: Yes, I am working on a function to redact the PII.

src/app/views/query-runner/QueryRunner.tsx
millicentachieng: It's being used to track version change.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
millicentachieng: @MIchaelMainer, I am capturing the query URL to help identify a code snippet or adaptive card, e.g. which are the most popular code snippets.

src/app/services/actions/query-action-creators.ts
```diff
@@ -96,6 +96,7 @@ async function createHistory(response: Response, respHeaders: any, query: IQuery
 
 
   const historyItem: IHistoryItem = {
+    index: -1,
```
thewahome, line 99: What does this property help with?

src/app/views/query-response/pivot-items/pivot-items.tsx
```diff
@@ -84,6 +90,29 @@ export const getPivotItems = (properties: any) => {
...
+      telemetry.trackEvent(TAB_CLICK_EVENT, { ComponentName: 'Adaptive Cards Tab', QueryUrl: queryUrl });
+      break;
+    }
+    case 'code-snippets': {
+      telemetry.trackEvent(TAB_CLICK_EVENT, { ComponentName: 'Code Snippets Tab', QueryUrl: queryUrl});
+      break;
+    }
+    default: {
+      break;
+    }
```
thewahome, line 112: To reduce code repetition of the trackEvent function, the switch can be changed so that it receives the only thing that is changing. in this case it is the component name

```suggestion
  let componentName = '';
  switch (tabKey) {
    case 'adaptive-cards': {
      componentName = 'Adaptive Cards Tab';
      break;
    }
    case 'code-snippets': {
      componentName = 'Code Snippets Tab';
      break;
    }
    default: {
      break;
    }
  }
  telemetry.trackEvent(TAB_CLICK_EVENT, { ComponentName: componentName, QueryUrl: queryUrl });
```

Another option, would be to pass the tabKey directly in the function and not do a switch case.

```suggestion
   telemetry.trackEvent(TAB_CLICK_EVENT, { ComponentName: tabKey, QueryUrl: queryUrl });
```

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
millicentachieng: I've excluded query parameters from the URL. Any other PII that may be in the remaining part is being redacted.

millicentachieng: I've excluded query parameters from the URL. Any other PII that may be in the remaining part is being redacted.

src/app/services/actions/query-action-creators.ts
millicentachieng: We are using it for telemetry to check how far a user goes in the history list.

src/app/views/query-response/pivot-items/pivot-items.tsx
millicentachieng: I'll take the first suggestion, thanks.

src/app/utils/sample-url-sanitizer.ts
```diff
@@ -0,0 +1,37 @@
+function redactPii(str: string)
+{
+  const piiRegex = [
```
ddyett, line 3: @MIchaelMainer  can you give this a look over to make sure there's no other formats that may be needed.

src/app/views/query-runner/request/Request.tsx
```diff
@@ -80,13 +86,35 @@ export class Request extends Component<IRequestComponent, any> {
...
+        break;
+      }
+      case 'auth': {
+        telemetry.trackEvent(TAB_CLICK_EVENT, { ComponentName: 'Auth Tab'});
+        break;
+      }
+      default: {
+        break;
+      }
+    }
```
thewahome, line 102: [nit pick] This can be summarised to a one-liner like
```suggestion
        telemetry.trackEvent(TAB_CLICK_EVENT, { ComponentName: `${tabKey} tab`});
```

src/telemetry/telemetry.ts
```diff
@@ -43,7 +45,17 @@ class Telemetry implements ITelemetry {
 
   private filterFunction(envelope: any) {
     // Identifies the source of telemetry events
-    envelope.baseData.name = 'Graph Explorer v4';
+    envelope.tags['ai.cloud.role'] = 'Graph Explorer v4';
+
+    // Redact PII from query url
+    const properties = envelope.baseData.properties || {};
+    for (const property in properties) {
+      if (property === 'QueryUrl')
```
MIchaelMainer, line 53: We shouldn't be doing this. I thought there was agreement that we would stay away from URLs and req/resp bodies. 

First, it is impossible for us to know all potential PII in URLs exposed by the different service APIs. Even if we nail it today, we don't review all new URLs so it is very possible that new forms of PII come onboard. So, take ID format. That format appears to be a guid. There is no guarantee that all identifiers or GUIDs. For example, OneDrive creates folder identifiers like:
`01DVF26FAOKNM7R54BENH3CL6W4KKY5MCL`. I have no idea how you regex for that.

Or OneDrive can create file names like:
`PUT /me/drive/root:/FolderA/FileB.txt:/content` <-- This case we can regex the content between the `:`, but I could imagine other scenarios where we don't have tokens to use for regex matches.

Second, we will need to support other requirements for the eventuality that we capture PII. There are significant requirements here IIRC for purging this information, protecting it, etc.

I feel strongly that we shouldn't capture URLs and depend on telemetry sanitization. With that, I think it is safe to capture URLs up through the entity set or singleton only if we validate that the user has entered a valid top level entityset or singleton. We'd want to validate as a user could conceivably enter PII by accident as the entity set. So, we could capture `https://graph.microsoft.com/{endpoint}/{entityset}/{MASKED*}` excluding the masked part.

src/app/utils/sample-url-sanitizer.ts
MIchaelMainer: We shouldn't do this. Explanation in my review comment in telemetry.ts.

MIchaelMainer (changes requested):

src/telemetry/telemetry.ts
millicentachieng: I'll remove URL from telemetry data. Let's talk some more on this.

MIchaelMainer (approved):

MIchaelMainer:
@millicentachieng I'd like to review this on staging.

millicentachieng:
In this PR, Query URLs which might contain PII are not being captured. The field QueryUrl has been added to some telemetry data points as a placeholder for the sake of building a proper schema for reporting. 

We will include QueryUrl in a later PR after we are confident of proper query normalization and redaction of PII.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
```diff
@@ -99,4 +100,6 @@ function mapDispatchToProps(dispatch: Dispatch): object {
 
 // @ts-ignore
 const IntlAdaptiveCard = injectIntl(AdaptiveCard);
-export default connect(mapStateToProps, mapDispatchToProps)(IntlAdaptiveCard);
+// @ts-ignore
```
ddyett, line 103: can we comment why we are ignoring something

src/app/views/sidebar/history/History.tsx
```diff
@@ -406,6 +424,17 @@ export class History extends Component<IHistoryProps, any> {
...
   }
 
+  private trackHistoryItemEvent = (eventName: string, componentName: string,
+      query: IHistoryItem) => {
+    telemetry.trackEvent(
+      eventName,
+      {
+        ComponentName: componentName,
+        ItemIndex: query.index,
+        QueryUrl: ''
```
ddyett, line 434: we should think about a different name for this so it's clear we are not capturing the query url.  more like the query signature.

src/app/views/sidebar/history/History.tsx
```diff
@@ -496,9 +525,9 @@ function mapDispatchToProps(dispatch: Dispatch): object {
   };
 }
 
-
-// @ts-ignore
-const styledHistory = styled(History, sidebarStyles);
+const styledHistory = styled(History, sidebarStyles as any);
```
ddyett, line 528: can we avoid the use of any?

src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -79,8 +79,21 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
 
   public onDocumentationLinkClicked = (item: any) => {
     window.open(item.docLink, '_blank');
+    this.trackDocumentLinkClickedEvent(item);
   };
 
+  private trackDocumentLinkClickedEvent(item: any) {
```
ddyett, line 85: use an interface so we keep type checking.

src/telemetry/telemetry.ts
```diff
@@ -30,20 +30,20 @@ class Telemetry implements ITelemetry {
   }
 
   public trackEvent(eventName: string, payload: any) {
-    this.appInsights.trackEvent({ name: eventName }, payload);
+    this.appInsights.trackEvent({ name: eventName, properties: payload });
```
ddyett, line 33: there should be a change here to re-enable telemetry so telemetry must not be flowing.

src/telemetry/telemetry.ts
```diff
@@ -30,20 +30,20 @@ class Telemetry implements ITelemetry {
...
-  public trackReactComponent(ComponentToTrack: ComponentType): ComponentType {
-    return withAITracking(this.reactPlugin, ComponentToTrack);
+  public trackReactComponent(ComponentToTrack: ComponentType, componentName?: string): ComponentType {
+    return withAITracking(this.reactPlugin, ComponentToTrack, componentName);
   }
 
   private filterFunction(envelope: any) {
     // Identifies the source of telemetry events
-    envelope.baseData.name = 'Graph Explorer v4';
+    envelope.tags['ai.cloud.role'] = 'Graph Explorer v4';
```
ddyett, line 46: Why are we using the cloud role for this name?  Normally that means more like it's a front end machine or something.

millicentachieng: I borrowed the idea from the fact that the property could be used in a microservices environment to identify individual services.  

To remove confusion, I've added application name as a custom property to be able to properly identify GE insights.

src/telemetry/telemetry.ts
```diff
@@ -14,7 +14,7 @@ class Telemetry implements ITelemetry {
     this.config = {
       instrumentationKey: this.getInstrumentationKey(),
       disableExceptionTracking: true,
-      disableTelemetry: true,
+      disableTelemetry: false,
```
ddyett, line 17: i think this code used to check for instrumentkey before.  you should just revert back to the version before Michael's change

src/app/views/sidebar/history/History.tsx
```diff
@@ -23,6 +25,7 @@ import { classNames } from '../../classnames';
 import { sidebarStyles } from '../Sidebar.styles';
 import { createHarPayload, exportQuery, generateHar } from './har-utils';
 
+
```
ddyett, line 28: NIT: remove space

src/app/views/sidebar/history/History.tsx
millicentachieng: Addressed in next PR

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
millicentachieng: Addressed in next PR

ddyett (approved):
Please look at squashing since we don't need all 46 of the commits here in the history.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
```diff
@@ -99,4 +100,6 @@ function mapDispatchToProps(dispatch: Dispatch): object {
...
+const trackedComponent = telemetry.trackReactComponent(IntlAdaptiveCard, AdaptiveCard.name);
+export default connect(mapStateToProps, mapDispatchToProps)(trackedComponent);
```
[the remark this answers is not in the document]
millicentachieng: From the data we've collected so far in test, the only property being auto-captured is `Component Name`. The other properties are those I've added in the filter function, `IsAuthenticated` and `ApplicationName`.

src/app/views/query-runner/request/auth/Auth.tsx
```diff
@@ -35,5 +36,5 @@ export function Auth(props: any) {
     }
   </div>);
 }
-
-export default styled(Auth, authStyles as any);
+const trackedComponent = telemetry.trackReactComponent(Auth, Auth.name);
```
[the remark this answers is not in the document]
millicentachieng: One of the requirements was to record how much time a user spends on the Auth tab. The `trackReactComponent` function gives us that metric. 

Are you suggesting that the metric is irrelevant?

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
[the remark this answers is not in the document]
MIchaelMainer: According to the code in withAITracking, it captures all properties. If we ever add another property that is not intended to be captured, it will be captured. Maybe I'm reading this incorrectly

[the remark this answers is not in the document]
millicentachieng: This piece of code from [here](https://github.com/microsoft/ApplicationInsights-JS/blob/5bd00322e30483552d75a09df81094d29237a967/extensions/applicationinsights-react-js/src/withAITracking.tsx#L69),  tells me that `Component Name`  is the only property captured. 
```
const engagementTime = this.getEngagementTimeSeconds();
const metricData: IMetricTelemetry = {
        average: engagementTime,
        name: 'React Component Engaged Time (seconds)',
        sampleCount: 1
      };

const additionalProperties: { [key: string]: any } = { 'Component Name': componentName };
reactPlugin.trackMetric(metricData, additionalProperties);
```
I might be missing what you are seeing

src/app/views/query-runner/request/auth/Auth.tsx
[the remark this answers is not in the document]
millicentachieng: Noted :)

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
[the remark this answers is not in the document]
ddyett: I doubt this would be a problem but just double check things are as expected.  It's a HOC which from the code I don't see a problem with but one never knows.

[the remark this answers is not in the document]
MIchaelMainer: I miss what I was seeing. Now that I look at it, I was mistaken.  It only captures the name like you said. Resolving this. 

But one thing this does bring up is that we are passing an entire component and we may not be aware of what type information AI captures from our component.

MIchaelMainer (approved):

MIchaelMainer (approved):

PR #677, opened 2020-08-27
Feature: History items ttl

### LINKED ISSUE #674: History data in indexDB should have a ttl < 12 months
Is your feature request related to a problem? Please describe.
Add a creationDate field for history items and delete history items when they surpass ttl. I propose a ttl of no more than 30 days. I hypothesize that history is not likely accessed within a day or two of being created.
Describe the solution you'd like
We add a field to history items to represent the date when a history item was created. When history is accessed, we query for and delete history items that are older than the ttl.
Additional context
For EU cookie compliance
[AB#5862](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/5862)

thewahome:
## Overview

Fixes #674 

### Notes

Shows only the history items that fit the threshold, delete the others

### DIFF
--- a/src/app/views/sidebar/history/history-utils.ts
+++ b/src/app/views/sidebar/history/history-utils.ts
@@ -13,27 +13,24 @@ export async function writeHistoryData(historyItem: IHistoryItem) {
 export async function readHistoryData(): Promise<IHistoryItem[]> {
   let historyData: IHistoryItem[] = [];
   const keys = await historyStorage.keys();
-  for (const element of keys) {
-    if (historyItemAgeYoungerThan(30, element)) {
-      const historyItem: IHistoryItem = await historyStorage.getItem(element);
+  for (const creationTime of keys) {
+    if (historyItemHasExpired(creationTime)) {
+      const historyItem: IHistoryItem = await historyStorage.getItem(creationTime);
       historyData = [...historyData, historyItem];
     } else {
-      historyStorage.removeItem(element);
+      historyStorage.removeItem(creationTime);
     }
   }
   return historyData;
 }
 
-function historyItemAgeYoungerThan(ageInDays: number, createdAt: string): boolean {
+function historyItemHasExpired(createdAt: string): boolean {
+  const ageInDays: number = 30;
   const dateToCompare = new Date();
   dateToCompare.setDate(dateToCompare.getDate() - ageInDays);
-  const compareDate = dateToCompare.getTime();
-
-  const createdDate = new Date(createdAt).getTime();
-  if (createdDate < compareDate) {
-    return false;
-  }
-  return true;
+  const expiryDate = dateToCompare.getTime();
+  const createdTime = new Date(createdAt).getTime();
+  return (createdTime >= expiryDate);
 }
 
 export const removeHistoryData = async (historyItem: IHistoryItem) => {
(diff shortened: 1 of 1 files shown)


src/app/views/sidebar/history/history-utils.ts
```diff
@@ -13,12 +14,28 @@ export async function readHistoryData(): Promise<IHistoryItem[]> {
...
+      historyStorage.removeItem(element);
+    }
   }
   return historyData;
 }
 
+function historyItemAgeYoungerThan(ageInDays: number, createdAt: string): boolean {
+  const dateToCompare = new Date();
+  dateToCompare.setDate(dateToCompare.getDate() - ageInDays);
+  const compareDate = dateToCompare.getTime();
```
zengin, line 30: consider renaming `compareDate` to `timeToCompare`

src/app/views/sidebar/history/history-utils.ts
```diff
@@ -13,12 +14,28 @@ export async function readHistoryData(): Promise<IHistoryItem[]> {
...
+
+  const createdDate = new Date(createdAt).getTime();
```
zengin, line 32: consider renaming `createdDate` to `createdTime`

src/app/views/sidebar/history/history-utils.ts
```diff
@@ -13,12 +14,28 @@ export async function readHistoryData(): Promise<IHistoryItem[]> {
...
+  if (createdDate < compareDate) {
+    return false;
+  }
+  return true;
```
zengin, line 36:
```suggestion
  return (createdTime >= timeToCompare);
```

jobala (changes requested):
What do you think of using an 'expire' metaphor here? We have a hard requirement of deleting history items older than 30 days so we can think of these items as having expired. Using this metaphor will have the following effect on the code.

```
const hasExpired = hasHistoryItemExpired(historyItem)

if (hasExpired) {
  // remove expired history item
} else {
  ...
}
```

We can apply the same metaphor to the date comparisons ie

```
if (createdDate < expiryDate) {
  ...
}
```

Curious to know what @zengin thinks of applying the metaphor to variable names instead of using `timeToCompare` and `createdTime`.

We should also add unit tests for this functionality, it should be possible to mock localForage

zengin:
@jobala, +1 on switching the terminology to "expiry". I agree that it will be easier to reason about with that change since the concept is well-understood in the world of cookies etc.

thewahome:
I am sold on switching the terminology to 'expiry'

thewahome:
@jobala I have added [an issue](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/689) to tackle the unit tests in another PR

src/app/views/sidebar/history/history-utils.ts
```diff
@@ -13,27 +13,24 @@ export async function writeHistoryData(historyItem: IHistoryItem) {
...
-  const compareDate = dateToCompare.getTime();
-
-  const createdDate = new Date(createdAt).getTime();
-  if (createdDate < compareDate) {
-    return false;
-  }
-  return true;
+  const expiryDate = dateToCompare.getTime();
+  const createdTime = new Date(createdAt).getTime();
+  return (createdTime >= expiryDate);
```
zengin, line 33: I think this should be the other way around to match the `expiry` concept, i.e. it has expired if created earlier than 30 days ago.

That will require the flipping the if/else block above as well, because we want to remove the item from storage if it has expired.

```suggestion
  return (createdTime < expiryDate);
```

zengin (approved):
🚢

jobala (approved):

PR #681, opened 2020-09-02
Feat/show long queries

### LINKED ISSUE #667: When the request url is longer adding a suggestion does not move to the end
1. Launch Graph Explorer
2. Enter the following 'https://graph.microsoft.com/v1.0/me?$select=deletedDateTime,ageGroup'
3. Type ',' and select something else

Result: The end of the url is not shown so it's difficult to continue typing.

Expectation: we should see the end of the url

jobala:
## Overview
This modifies the query input field such that it adapts to show long urls.

### Demo
N/A

### Notes

N/A

## Testing Instructions

* Enter a long sample url
* Select a suggestion
* Observe that you can see the suggesstion you selected

Fixes #667

### DIFF
--- a/src/app/views/query-runner/query-input/AutoComplete.tsx
+++ b/src/app/views/query-runner/query-input/AutoComplete.tsx
@@ -265,12 +265,20 @@ class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
       <>
         <TextField
           className={autoInput}
+          styles={() => ({
+            fieldGroup: { minHeight: 30 },
+            field: { height: 30 },
+
+          })}
           type='text'
+          autoAdjustHeight={true}
           autoComplete='off'
           onChange={this.onChange}
           onKeyDown={this.onKeyDown}
           value={sampleQuery.sampleUrl}
           componentRef={this.autoCompleteRef}
+          multiline={true}
+          resizable={false}
           onRenderSuffix={(fetchingSuggestions) ? this.renderSuffix : undefined}
         />
         {showSuggestions && userInput && filteredSuggestions.length > 0 &&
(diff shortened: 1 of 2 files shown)


src/app/views/query-runner/query-input/AutoComplete.tsx
```diff
@@ -265,12 +265,20 @@ class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
       <>
         <TextField
           className={autoInput}
+          styles={() => ({
+            fieldGroup: { minHeight: 30 },
+            field: { height: 30 },
+
+          })}
```
thewahome, line 272: `  styles={
            fieldGroup: { minHeight: 30 },
            field: { height: 30 },
          }`

What is the difference between this style definition and the one you added?

jobala: There's no difference, it's possible to use a style object or a style function.

ddyett:
just not sure i like the look of it.  Can you specify a maximum height?

jobala:
> just not sure i like the look of it. Can you specify a maximum height?

Yes it's possible but the user will have to scroll to get to the end of the url

ddyett:
> > just not sure i like the look of it. Can you specify a maximum height?
> 
> Yes it's possible but the user will have to scroll to get to see the cursor

yeah but i can really distory the UI.  Plus the issue isn't that we never scroll to the end, i think it was just after pasting and immediately typing.

jobala:
@ddyett I believe I have this fixed now. After entering a long url and then picking a suggestion, the selected suggestion is visible.

ddyett (approved):

PR #691, opened 2020-09-12
Task: update azure pipeline

thewahome:
## Overview

Separates build and release pipelines
Publishes artifacts


MIchaelMainer (approved):

azure-pipelines.yml
```diff
@@ -1,10 +1,10 @@
 pool:
-  vmImage: 'vs2017-win2016'
+  vmImage: 'ubuntu-latest'
```
MIchaelMainer, line 2: We should have a path spec to ignore changes that don't impact deliverable

MIchaelMainer (changes requested):

azure-pipelines.yml
```diff
@@ -1,89 +1,179 @@
...
+          destination: 'azureBlob'
+          storage: 'graphstagingblobstorage'
+          containerName: 'staging/vendor/bower_components/explorer-v2/build'
+          cleanTargetBeforeCopy: true
+        condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/dev'))
+        displayName: 'Deploy to -tst'
+        enabled: false
+
+      - task: AzureRmWebAppDeployment@3
+        condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
```
MIchaelMainer, line 161: With the successful deployment to tst, you can now remove these disabled tasks. Still keep the notification disabled as we still need to create that connection.

PR #696, opened 2020-09-16
Task: Reduce padding/spacing on request header 

### LINKED ISSUE #427: Reduce the padding/spacing in the Request headers so that more key Values are seen on the small space available
At the moment the request header area looks like this: 
image

It would look way better if we reduced the spacing as indicated by the arrows below:
image

ElinorW:
## Overview
Reduce padding/spacing on request header to enhance key-values visibility
Fixes #427 

#### Before: 
image

#### After: 
image


thewahome (approved):
LGTM

PR #713, opened 2020-09-25
disable methods to notify not logged in users

### LINKED ISSUE #700: Can't select method other than GET
When building the request, it's not possible to select any other method than GET.

**To Reproduce**
1. Go to Graph Explorer
2. Click on GET
3. See error

**Expected behavior**
Be able to select any request method

**Screenshots**
image

 - OS: macOS
 - Browser Chrome
 - Version 85
 
[AB#7487](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7487)

thewahome:
## Overview

Shows subtle notification to make sure that the user can know the difference between being logged in and not being logged in.
Fixes #700 

### Demo

Logged in:
image

Not logged in:
image

### DIFF
--- a/src/app/views/query-runner/QueryInput.tsx
+++ b/src/app/views/query-runner/QueryInput.tsx
@@ -1,9 +1,8 @@
-import { Dropdown, TextField } from 'office-ui-fabric-react';
+import { Dropdown, Icon, IDropdownOption, TextField } from 'office-ui-fabric-react';
 import React, { Component } from 'react';
 import { injectIntl } from 'react-intl';
 import { connect } from 'react-redux';
 
-import { Mode } from '../../../types/enums';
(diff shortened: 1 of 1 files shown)


bettirosengugi:
@thewahome on hover, is there a message letting the user know that they need to log in to access the greyed out request options?

thewahome:
> @thewahome on hover, is there a message letting the user know that they need to log in to access the greyed out request options?

No. Once the element is disabled, the hover elements do not work. It thus made sense to add familiar icons to show that these options are locked similar to those for the sidebar element

ddyett:
not sure how useful this is if the user does not understand what needs to be done to unlock.  It may be better if they are enabled and the act of enabling forces the user to sign in.

PR #734, opened 2020-10-12
Telemetry: Normalize Query URL and Redact PII

millicentachieng:
## Overview

This PR aims to trim out parts of the Query URL and de-identify sections of the URL that could contain sensitive data.

Below is a telemetry data capture code snippet.
```
function trackTabClickEvent(query: IQuery, tabTitle: string) {
  const normalizedUrl = normalizeQueryUrl(query.sampleUrl);
  telemetry.trackEvent(TAB_CLICK_EVENT,
    {
      ComponentName: `${tabTitle} tab`,
      QuerySignature: `${query.selectedVerb} ${normalizedUrl.incomingUrl}`,
      QueryString: normalizedUrl.queryString
    });
}
```

What the `normalizeQueryUrl` function does:
* Replace all values of ODATA query parameters with `<value>` while retaining property names e.g.  
      - `GET https://graph.microsoft.com/v1.0 /users?$search=displayName:<value>`
      - `GET https://graph.microsoft.com/v1.0/me/messages?$filter=from/emailAddress/address eq <value>`
* All other query parameter values will be removed e.g.
      -  `GET https://graph.microsoft.com/v1.0/bookingBusinesses/<id>/calendarView?start=<value>&end=<value>`


src/app/views/query-response/pivot-items/pivot-items.tsx
```diff
@@ -130,14 +131,24 @@ export const getPivotItems = (properties: any) => {
...
   }
 };
 
+function trackTabClickEvent(query: IQuery, tabTitle: string) {
+  const normalizedUrl = normalizeQueryUrl(query.sampleUrl);
+  telemetry.trackEvent(TAB_CLICK_EVENT,
+    {
+      ComponentName: `${tabTitle} tab`,
+      QuerySignature: `${query.selectedVerb} ${normalizedUrl.incomingUrl}`,
+      QueryString: normalizedUrl.queryString
```
MIchaelMainer, line 148: I suggest we rename `QueryString` as `QueryStringSignature` and `normalized*` becomes `sanitized*`. Also specify the return type.

This comment is applicable to all functions that call trackEvent..

MIchaelMainer (changes requested):

MIchaelMainer:
And we need test cases

src/app/utils/query-url-normalization.ts
```diff
@@ -0,0 +1,136 @@
...
+  incomingUrl = decodeURIComponent(incomingUrl);
+
+  // Extract query string
+  let queryString: string = '';
+  const urlParts = incomingUrl.split('?');
+  if (urlParts.length > 1) {
+    const startOfQueryString = incomingUrl.indexOf('?');
+    queryString = incomingUrl.substring(startOfQueryString, incomingUrl.length - startOfQueryString);
+    queryString = processQueryParams(queryString.slice(1));
+  }
```
thewahome, line 28: A reusable function created to get the parts of the url exists as `parseSampleUrl` in the `sample-url=generation.ts` file.

src/app/utils/query-url-normalization.ts
```diff
@@ -0,0 +1,136 @@
...
+
+function processQueryParams(queryString: string) {
+   const params = queryString.split('&');
+   let result: string = '';
+   if (params.length)
+   {
+      params.forEach(param => {
+       result += removeSensitiveQueryParamValues(param) + '&';
+      });
+      result = result.slice(0, -1);
```
andrueastman, line 80: It would be helpful to add a comment to show that we are removing the last ampersand

thewahome (changes requested):

src/app/views/query-response/pivot-items/pivot-items.tsx
millicentachieng: Removed QueryString and concatenated it to the sanitized URL.

src/app/utils/query-url-normalization.ts
```diff
@@ -0,0 +1,136 @@
...
+  // Replace IDs and ID placeholders with generic {id}
+  incomingUrl = incomingUrl.replace(keyIdRegex, '/<id>');
+  incomingUrl = incomingUrl.replace(guidRegex, '<id>');
+  incomingUrl = incomingUrl.replace(numberRegex, '<id>');
+
+  // Redact PII
+  incomingUrl = incomingUrl.replace(emailRegex, '<email>');
+
+  // Drop entity/action namespace
+  incomingUrl = incomingUrl.replace('microsoft.graph.', '');
```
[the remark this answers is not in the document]
millicentachieng: I've removed this part

src/app/utils/query-url-normalization.ts
millicentachieng: I've removed this part. I am using an existing function to do this.

src/app/utils/query-url-normalization.ts
```diff
@@ -0,0 +1,136 @@
...
+    case '$expand':
+    case '$select':
+    case '$format':
+    case '$orderby': {
+        break;
+    }
+    /**
+     * Query URLs will look like the examples below after processing,
+     * GET /me/people/?$search=<value>
+     * GET /users?$search=displayName:<value>
```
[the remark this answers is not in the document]
millicentachieng: The syntax from the [documentation](https://docs.microsoft.com/en-us/graph/query-parameters#search-parameter) is `https://graph.microsoft.com/users?$search="displayName:Michael Mainer"`

If there's an OR or AND, it's handling that as well but only if the search term is between quotes. I'm updating the regex to handle unquoted search text

src/app/utils/query-url-sanitization.ts
```diff
@@ -15,74 +17,44 @@ const guidRegex =
...
-  incomingUrl = urlParts[0];
-
-  // Drop casts that use the /$/ pattern
-  incomingUrl = incomingUrl.split('/$/')[0];
+  let sanitizedUrl = url.split('?')[0];
 
   // Normalize parameters in param=<arbitraryKey> format with param={value} format
-  incomingUrl = incomingUrl.replace(functionParamInitialRegex, '<value>');
-  incomingUrl = incomingUrl.replace(functionParamFinalRegex, '<value>');
+  sanitizedUrl = sanitizedUrl.replace(functionParamInitialRegex, '<value>');
```
thewahome, line 31: Is it possible for us to get mention the resource in the generated arbitrary key?
e.g for `/me/messages/abcd` we can replace it with `/me/messages/{message-id}`

The reason for this is I'd like to reuse this in the autocomplete work

src/tests/utils/sanitize-url.spec.tsx
```diff
@@ -1,32 +1,91 @@
...
+
+  list.forEach(element => {
+    it(`return ${element.isAllAlphabetic} for ${element.key}`, () => {
+      const key = isAllAlpha(element.key);
+      expect(key).toBe(element.isAllAlphabetic);
+    });
+  });
+});
+
+describe('isDepraction should ', () => {
```
MIchaelMainer, line 22: I just realized that I've been saying deprecated wrong all of these years. `depracate` should be `deprecate`

MIchaelMainer (commented):
What more do we have here?

src/tests/utils/sanitize-url.spec.tsx
millicentachieng: I've added more tests

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,49 +1,49 @@
+import { arrayify } from 'tslint/lib/utils';
```
MIchaelMainer, line 1: remove this

src/app/views/query-runner/request/Request.tsx
```diff
@@ -86,18 +88,29 @@ export class Request extends Component<IRequestComponent, any> {
...
     const tabTitle = item.props.title;
     if (tabTitle) {
-      telemetry.trackEvent(TAB_CLICK_EVENT, { ComponentName: `${tabTitle} tab`, QuerySignature: '' });
+      this.trackTabClickEvent(tabTitle);
     }
   }
 
+  private trackTabClickEvent(tabTitle: string) {
+    const { sampleQuery }: any = this.props;
+    const sanitizedUrl = sanitizeQueryUrl(sampleQuery.sampleUrl);
```
thewahome, line 97: The sampleQuery is not part of the properties in the component. This will lead to a crash when a tab is clicked

src/app/utils/query-parameter-sanitization.ts
```diff
@@ -0,0 +1,396 @@
...
+ * - GET /categories?$orderby=products/$count
+ */
+function sanitizeOrderByQueryOptionValue(queryOptionValue: string): string {
+  const sortingExpressions = queryOptionValue.split(',');
+
+  sortingExpressions.forEach((expr, index) => {
+    const expressionParts = expr.split(' ').filter(x => x !== ''); // i.e. property name and sort order
+    let propertyName = expressionParts[0].trim();
+    if (!isPropertyName(propertyName)) {
+      propertyName = '<invalid-value>';
```
zengin, line 193: consider using `<invalid-property>` to distinguish from other `<invalid-value>`s

src/app/utils/query-parameter-sanitization.ts
```diff
@@ -0,0 +1,396 @@
...
+    // Sanitize segment in the form of "pizza" and "body:excitement"
+    if (QUOTED_TEXT_REGEX.test(segment)) {
+      if (!segment.includes(':')) {
+        sanitizedQueryString += ` <value>`;
+      }
+      else {
+        // Extract property name
+        let propertyName = segment.substring(1, segment.indexOf(':')).trim();
+        if (!isAllAlpha(propertyName)) {
+          propertyName = '<property>';
```
zengin, line 248: Q: What is the reason to redact property names as PII? And it seems to be redacting only in the case of invalid property names. I don't fully understand this if block.

src/app/utils/query-url-sanitization.ts
```diff
@@ -0,0 +1,79 @@
+import { GRAPH_URL } from '../services/graph-constants';
+import { isAllAlpha, sanitizeQueryParameter } from './query-parameter-sanitization';
+import { parseSampleUrl } from './sample-url-generation';
+
+// Matches strings with deprecation identifier
+const DEPRECATION_REGEX = /^[A-Za-z]+_v2$/gi;
```
zengin, line 6: `A-Z` is redundant given that `/i` matches case-insensitive.

src/app/utils/query-url-sanitization.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  const params = queryString.split('&');
+  let sanitizedQueryParams: string = '';
+  if (params.length) {
+    params.forEach(param => {
+      sanitizedQueryParams += sanitizeQueryParameter(param) + '&';
+    });
+    sanitizedQueryParams = sanitizedQueryParams.slice(0, -1);
+  }
+  return `?${sanitizedQueryParams}`;
```
zengin, line 78:
```suggestion
  return queryString.split('&').map(sanitizeQueryParameter).join('&');
```

zengin (changes requested):

src/app/utils/query-parameter-sanitization.ts
millicentachieng: It's mainly to handle errors that our users make. I'm making the assumption that if the property name is not  valid, it could contain sensitive data. Just being cautious.

zengin: fair enough! Thanks for the explanation.

src/app/utils/query-parameter-sanitization.ts
```diff
@@ -0,0 +1,396 @@
...
+const ALL_ALPHA_REGEX = /^[a-z]+$/i;
+const POSITIVE_INTEGER_REGEX = /^[1-9]+$/;
+// Matches the format json or application/json
+const MEDIA_TYPE_REGEX = /^[a-z]+\/{0,1}[a-z]+$/i;
+// Matches the format key=value
+const KEY_VALUE_REGEX = /^[a-z]+=[a-z]+$/i;
+// Matches property name patterns e.g. displayName or from/emailAddress/address
+const PROPERTY_NAME_REGEX = /^([a-z]+(\/?\b[a-z]+\b)+)|([a-zA-Z]+)$/i;
+// Matches pattterns within quotes e.g "displayName: Gupta"
+const QUOTED_TEXT_REGEX = /^["']([^"]*)['"]$/;
```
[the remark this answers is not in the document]
millicentachieng: The double quotes are acceptable within the $search query option. Examples are [here](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_SystemQueryOptionsearch). In my sanitization, I am only looking for them within the $search query option. It works as expected with escaped single quotes as well.

The example you've given would be sanitized to http://host/service/People(<people-id>).

src/app/utils/query-parameter-sanitization.ts
```diff
@@ -0,0 +1,396 @@
...
+const COMPARISON_OPERATORS = ['eq', 'ne', 'gt', 'ge', 'lt', 'le'];
+const LOGICAL_OPERATORS = ['and', 'or', 'not'];
+const COLLECTION_OPERATORS = ['any', 'all'];
+const QUERY_FUNCTIONS = ['startswith', 'endswith', 'contains', 'substring', 'indexof', 'concat'];
+
+// REGEXES
+const ALL_ALPHA_REGEX = /^[a-z]+$/i;
+const POSITIVE_INTEGER_REGEX = /^[1-9]+$/;
+// Matches the format json or application/json
+const MEDIA_TYPE_REGEX = /^[a-z]+\/{0,1}[a-z]+$/i;
```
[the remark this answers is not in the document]
millicentachieng: I'm not really validating the types but their formats. They could be anything e.g. `application/json;odata=verbose` or `xml` or an invalid type like `media/type` as long as it's all text separated by `/` or  text that is all letters.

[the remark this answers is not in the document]
zengin: Because it only matches characters on both sides of `/`, sanitization process will mark `application/octet-stream` as `<invalid-media-type>`. Is that desired?

For full list of media types, please refer to http://www.iana.org/assignments/media-types/media-types.xhtml, which contains non-alpha characters as well. I don't know which subset of those are allowed in the formats you mention though.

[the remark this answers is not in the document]
millicentachieng: Thanks for sharing the list. I'll use it to update the regex. Much  appreciated.

src/app/utils/query-parameter-sanitization.ts
```diff
@@ -0,0 +1,419 @@
...
+
+    case '$count': {
+      if (!isBooleanString(value)) {
+        value = '<invalid-value>';
+      }
+      break;
+    }
+
+    case '$select': {
+      value = sanitizeSelectQueryOptionValue(value);
```
thewahome, line 94: I like how the functions are specific 👍🏾

src/app/utils/query-parameter-sanitization.ts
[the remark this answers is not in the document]
millicentachieng: I updated the regex

src/app/utils/query-parameter-sanitization.ts
```diff
@@ -0,0 +1,415 @@
...
+const ARITHMETIC_OPERATORS = ['add', 'sub', 'mul', 'div', 'divby', 'mod'];
+const COMPARISON_OPERATORS = ['eq', 'ne', 'gt', 'ge', 'lt', 'le'];
+const LOGICAL_OPERATORS = ['and', 'or', 'not'];
+
+// REGEXES
+const ALL_ALPHA_REGEX = /^[a-z]+$/i;
+const POSITIVE_INTEGER_REGEX = /^[1-9]\d*$/;
+// Matches media type formats
+// Examples: https://www.iana.org/assignments/media-types/media-types.xhtml
+const MEDIA_TYPE_REGEX = /^(([a-z]+\/)?\w[\w+-.]+)$/i;
```
zengin, line 12: I would like to propose the following instead, because having a `()` matches parts like `application/` as a separate group and we don't need that, So I removed the parentheses.

Also regarding words after the `/`, we can have "0 or more words" in addition to the first word, instead of "1 or more words". So I changed the last `+` to `*`.

Those changes break matching single word, which you previously had, so I added that with `|[a-z]+` to be more explicit about it.

Depending on how you consume this regex, you may need the outermost parentheses. Please make sure to run the unit tests :)

Please let me know if this makes sense.

```suggestion
const MEDIA_TYPE_REGEX = /^[a-z]+\/\w[\w+-.]*|[a-z]+$/i;
```

zengin (changes requested):

src/app/utils/query-parameter-sanitization.ts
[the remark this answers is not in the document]
MIchaelMainer: We also need to support image/jpeg

GET https://graph.microsoft.com/v1.0/me/photo/$value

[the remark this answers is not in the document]
MIchaelMainer: We don't want to miss out on this image: 
image

MIchaelMainer (approved):
Approve pending Mustafa's last comments resolution.

Thank you for putting examples for each URL form that is sanitized. That was useful.

src/app/utils/query-parameter-sanitization.ts
millicentachieng: Here are patterns that the original regex was supposed to match:
- `application/json`
- `application/1d-interleaved-parityfec`
- `application/atom+xml`

or their abbreviations
- `json`
- `1d-interleaved-parityfec`
- `atom+xml` 
(not sure if the last two abbreviations are true, but you get where I am going with this)

The outermost parentheses are required..

I get having the part `\w[\w+-.]+` like this `\w[\w+-.]*`  but not the first part about removing the parentheses. Could you please give more info if you can?

src/app/utils/query-parameter-sanitization.ts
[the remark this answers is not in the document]
millicentachieng: Michael, I realized the comment was misleading. I've updated it.

src/app/utils/query-parameter-sanitization.ts
zengin: ok, this makes sense. I thought you were matching application, not JSON. Thanks for the examples.

zengin (approved):

thewahome (approved):

PR #741, opened 2020-10-15
Task/code formatting

jobala:
## Overview

Add code formatting settings for *vscode*.

### Demo

N/A

### Notes

This is only a short term fix because we are all using VisualStudio Code. It will be better to use [prettier](https://prettier.io/) for formatting because it is editor independent.

## Testing Instructions

* Write something with the wrong indentation
* Save the document
* Note that the code is formatted correctly to fix indentation


jobala:
@thewahome any idea why this is failing? I have not touched any code files.

ElinorW (approved):

thewahome:
> @thewahome any idea why this is failing? I have not touched any code files.

Must be the github actions running... It can be ignored

PR #743, opened 2020-10-19
task:enhance sanitise query function

thewahome:
## Overview

This work enhances the work done by @millicentachieng in #734 to redact PII from the query url.

## Notes
Assumptions made are that URL segments with identifiers are never full alphabetic strings. They will be strings with numbers / special characters. 

## Tasks
- [x] Checks the URL segments for identifier characters 

- [x] Replaces the id with the resource name e.g. `messages-id` / `tasks-id`

- [x] Adds tests to validate scenarios

### DIFF
--- a/src/app/utils/query-url-sanitization.ts
+++ b/src/app/utils/query-url-sanitization.ts
@@ -4,10 +4,6 @@ import { parseSampleUrl } from './sample-url-generation';
 // Matches pattterns within quotes e.g "displayName: Gupta"
 const quotedTextRegex = /"([^"]*)"/g;
 
-// Matches patterns with alphanumeric characters e.g <CONGZUWfGUu4msTgNP66e2UAAySi>
-const alphaNumericRegex = /^[a-z0-9]+$/i;
-
-
 export function sanitizeQueryUrl(url: string): string {
   url = decodeURIComponent(url);
 
@@ -34,48 +30,70 @@ export function sanitizeQueryUrl(url: string): string {
     }
   });
 
-  return `${urlPrefix}${sanitizedUrl}${queryString}` ;
+  return `${urlPrefix}${sanitizedUrl}${queryString}`;
 }
 
 function sanitizeQueryParameters(queryString: string): string {
-   const params = queryString.split('&');
-   let result: string = '';
-   if (params.length)
-   {
-      params.forEach(param => {
-       result += sanitizeQueryParameterValue(param) + '&';
-      });
-      result = result.slice(0, -1);
-   }
-   return result;
+  const params = queryString.split('&');
+  let result: string = '';
+  if (params.length) {
+    params.forEach(param => {
+      result += sanitizeQueryParameterValue(param) + '&';
+    });
+    result = result.slice(0, -1);
+  }
+  return result;
 }
 
 /*
-* A string contains an identifier if:
-*   1. is an alphanumeric string
+* A string contains an identifier if either:
+*   1. it is an alphanumeric string
 *   2. special characters are present
 */
-function containsIdentifier(segment: string): boolean {
+export function containsIdentifier(segment: string): boolean {
   return isAlphaNumericString(segment) || hasSpecialCharacters(segment);
 }
 
-function hasSpecialCharacters(segment: string): boolean {
+export function hasSpecialCharacters(segment: string): boolean {
   const specialCharacters = ['.', '=', '@', '-'];
-  return specialCharacters.some((character) =>  segment.includes(character));
+  return specialCharacters.some((character) => segment.includes(character));
 }
 
-function isAlphaNumericString(segment: string): boolean {
-  return !!segment.match(alphaNumericRegex);
+export function isAlphaNumericString(str: string): boolean {
+  let code, i, len;
+  let isNumeric = false;
+  let isAlpha = false;
+
+  for (i = 0, len = str.length; i < len; i++) {
+    code = str.charCodeAt(i);
+
+    switch (true) {
+      // check if all values are 0-9
+      case code > 47 && code < 58:
+        isNumeric = true;
+        break;
+
+      // check if all values are A-Z or a-z
+      case (code > 64 && code < 91) || (code > 96 && code < 123):
+        isAlpha = true;
+        break;
+
+      // not 0-9, not A-Z or a-z
+      default:
+        return false;
+    }
+  }
+
+  return isNumeric && isAlpha;
 }
 
 function sanitizeQueryParameterValue(param: string) {
-  if (!param.includes('='))
-  {
+  if (!param.includes('=')) {
     return param;
   }
   param = decodeURIComponent(param);
   const key = param.split('=')[0];
-  switch (key){
+  switch (key) {
     // We do not expect sensitive data in these OData query params, nothing needs to be done
     case '$top':
     case '$skip':
@@ -84,7 +102,7 @@ function sanitizeQueryParameterValue(param: string) {
     case '$select':
     case '$format':
     case '$orderby': {
-        break;
+      break;
     }
     /**
      * Query URLs will look like the examples below after processing,
@@ -93,16 +111,16 @@ function sanitizeQueryParameterValue(param: string) {
      * GET /users?$search=displayName:<value> OR mail:<value>
      */
     case '$search': {
-      param = param.replace(quotedTextRegex,  (capture) => {
+      param = param.replace(quotedTextRegex, (capture) => {
         if (!capture.includes(':')) {
-            return '<value>';
+          return '<value>';
         }
         // Drop quotes enclosing property and text to search
         capture = capture.replace('"', '');
         const property = capture.split(':')[0];
         return `${property}:<value>`;
-        });
-        break;
+      });
+      break;
     }
     /**
      * Examples
@@ -114,9 +132,9 @@ function sanitizeQueryParameterValue(param: string) {
       break;
     }
     default: {
-        param = `${key}=<value>`;
-        break;
+      param = `${key}=<value>`;
+      break;
     }
   }
-  return  param;
+  return param;
 }
--- a/src/app/utils/query-url-sanitization.ts
+++ b/src/app/utils/query-url-sanitization.ts
@@ -1,21 +1,10 @@
 import { GRAPH_URL } from '../services/graph-constants';
 import { parseSampleUrl } from './sample-url-generation';
 
-// Matches patterns like "('<1f7ff346-c174-45e5-af38-294e51d9969a>')" or "('<key>')"
-const keyIdRegex = /\(\'\<?\{?[ ?0-9a-zA-Z-]*\}?\>?\'\)/g;
-
-// Matches patterns like "(query=<key>)" or "(itemat=<key>,mode=<mode>)"
-const functionParamInitialRegex = /(?<=\=).*?(?=,)/g;
-const functionParamFinalRegex = /(?<=\=).[^,]*(?=\))/g;
-
 // Matches pattterns within quotes e.g "displayName: Gupta"
 const quotedTextRegex = /"([^"]*)"/g;
 
-// Matches PII patterns
-const numberRegex =  /(?<=\/)\d+\b/g; // number between forward slashes
-const emailRegex = /([a-z0-9_\-.+]+)@\w+(\.\w+)*/gi;
-const guidRegex =
- /(\{){0,1}[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}(\}){0,1}/gi;
+const alphaNumericRegex = /[^a-z0-9]/g;
 
 
 export function sanitizeQueryUrl(url: string): string {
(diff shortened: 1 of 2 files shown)


src/app/utils/query-url-sanitization.ts
```diff
@@ -1,44 +1,40 @@
...
 // Matches pattterns within quotes e.g "displayName: Gupta"
 const quotedTextRegex = /"([^"]*)"/g;
 
-// Matches PII patterns
-const numberRegex =  /(?<=\/)\d+\b/g; // number between forward slashes
-const emailRegex = /([a-z0-9_\-.+]+)@\w+(\.\w+)*/gi;
-const guidRegex =
- /(\{){0,1}[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}(\}){0,1}/gi;
+// Matches patterns with alphanumeric characters e.g <CONGZUWfGUu4msTgNP66e2UAAySi>
+const alphaNumericRegex = /[^a-z0-9]/g;
```
zengin, line 8: This matches a single character that is **not** in a to z and 0 to 9 intervals so it doesn't match alphanumeric characters (hence doesn't match with the comment above).

src/app/utils/query-url-sanitization.ts
```diff
@@ -54,6 +50,15 @@ function sanitizeQueryParameters(queryString: string): string {
    return result;
 }
 
+/*
+* Special characters present in the url show that the
+* segment contains an identifier
+*/
+function hasSpecialCharacters(segment: string): boolean {
+  const specialCharacters = ['.', '=', '@', '-'];
+  return specialCharacters.some((character) =>  segment.includes(character) || !!segment.match(alphaNumericRegex));
```
zengin, line 59: `!!segment.match(alphaNumericRegex)` doesn't depend on `character` so this line can be more efficient if replaced by:
```suggestion
  return !!segment.match(alphaNumericRegex) || specialCharacters.some((character) =>  segment.includes(character));
```

That said, I am confused about the logic due to the definition of alphaNumericRegex. Maybe a few tests targeting this function directly can help document the expected behavior here.

zengin (changes requested):

src/app/utils/query-url-sanitization.ts
thewahome: I ended up changing to a function. I could not get the regex to check whether the string contains both letters and numbers.

src/app/utils/query-url-sanitization.ts
thewahome: Nice catch @zengin . I have added the tests to verify this

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,99 @@
...
+  }
+  return result;
+}
+
+/*
+* A string contains an identifier if either:
+*   1. it is an alphanumeric string
+*   2. special characters are present
+*/
+export function containsIdentifier(segment: string): boolean {
```
MIchaelMainer, line 53: *Comment is for info purposes only, no action needed at this time besides opening a tracking work item to make key sanitization based on OData entity collections*

While I think it is safer to assume (but not totally safe) that a segment that has a special character is an identifier (I don't resource names will have a special character), I don't think it is safe to assume that an identifier has to have both alpha and numeric characters for an identifier. 

[From 4.1 Entity-Ids and Entity References](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#_Toc31358841): The entity-id MUST be an IRI as defined in [[RFC3987]](https://tools.ietf.org/html/rfc3987). It could be just a number, it could be just a letter. There is a Python package for validating this form: https://pypi.org/project/rfc3987/. I didn't find one on npm.

src/tests/utils/sanitize-url.spec.tsx
```diff
@@ -1,32 +1,80 @@
-import { sanitizeQueryUrl } from "../../app/utils/query-url-sanitization";
+import { containsIdentifier, isAlphaNumericString, sanitizeQueryUrl } from '../../app/utils/query-url-sanitization';
+
+describe('containsIdentifier should ', () => {
+  const list = [
+    { key: 'oIx3zN98jEmVOM', contains: true },
+    { key: 'oIx3zN98jEmVOM-4mUJzSGUANeje', contains: true },
+    { key: 'planner', contains: false }
```
MIchaelMainer, line 7: According to odata spec, `planner` could be an identifier, but we aren't in a position to support that.. Can you add more tests for other known identifier types from the actual service? Use ids from the demo tenant. Include DriveItem - files, DriveItem - folders, messageId (same form as the other Outlook based APIs), teams, anda few other workloads.

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,99 @@
...
+  const urlPrefix = `${GRAPH_URL}/${queryVersion}`;
+  sanitizedUrl = sanitizedUrl.replace(urlPrefix, '');
 
-  // Replace IDs and ID placeholders with generic {id}
-  sanitizedUrl = sanitizedUrl.replace(keyIdRegex, '/<id>');
-  sanitizedUrl = sanitizedUrl.replace(guidRegex, '<guid>');
-  sanitizedUrl = sanitizedUrl.replace(numberRegex, '<number>');
+  const sections = sanitizedUrl.split('/');
+  sections.forEach(sec => {
+    if (containsIdentifier(sec)) {
```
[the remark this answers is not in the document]
MIchaelMainer: It won't always be true since we don't follow RFC 3987, which we couldn't since all segments could be an identifier. `containsIdentifier()` works on serviceable identification of identifiers, but should be considered incomplete. We will want a work item to track URL key sanitization based on entity collections. 

This comment is for information only and can be resolved.

MIchaelMainer (changes requested):
I think this approach will work for now for the current entity keys, but we will do need a tracking item to make entity key sanitization based on CSDL definitions for entity collections and not based on regex as the CSDL definitions are the only deterministic way to sanitize URLs of entity keys.

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,99 @@
...
 
-  // Normalize parameters in param=<arbitraryKey> format with param={value} format
-  sanitizedUrl = sanitizedUrl.replace(functionParamInitialRegex, '<value>');
-  sanitizedUrl = sanitizedUrl.replace(functionParamFinalRegex, '<value>');
+  /*
+  * Remove url prefix and query version.
+  * They have special characters used in the check that follows
+  */
+  const urlPrefix = `${GRAPH_URL}/${queryVersion}`;
+  sanitizedUrl = sanitizedUrl.replace(urlPrefix, '');
```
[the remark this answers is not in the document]
thewahome: Makes sense...

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,92 @@
...
   url = decodeURIComponent(url);
 
   // Extract query string
-  const { search } = parseSampleUrl(url);
+  const { search, queryVersion, requestUrl } = parseSampleUrl(url);
   const queryString: string = sanitizeQueryParameters(search);
 
   // Drop query string
-  let sanitizedUrl = url.split('?')[0];
+  let sanitizedUrl = requestUrl;
```
thewahome, line 15:
```suggestion
  let sanitizedUrl = '';
```

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,92 @@
...
 
-  // Normalize parameters in param=<arbitraryKey> format with param={value} format
-  sanitizedUrl = sanitizedUrl.replace(functionParamInitialRegex, '<value>');
-  sanitizedUrl = sanitizedUrl.replace(functionParamFinalRegex, '<value>');
-
-  // Replace IDs and ID placeholders with generic {id}
-  sanitizedUrl = sanitizedUrl.replace(keyIdRegex, '/<id>');
-  sanitizedUrl = sanitizedUrl.replace(guidRegex, '<guid>');
-  sanitizedUrl = sanitizedUrl.replace(numberRegex, '<number>');
+  const sections = sanitizedUrl.split('/');
```
thewahome, line 17:
```suggestion
  const sections = requestUrl.split('/');
```

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,97 @@
...
-  // Normalize parameters in param=<arbitraryKey> format with param={value} format
-  sanitizedUrl = sanitizedUrl.replace(functionParamInitialRegex, '<value>');
-  sanitizedUrl = sanitizedUrl.replace(functionParamFinalRegex, '<value>');
+/*
+* A string contains an identifier if either:
+*   1. it is an alphanumeric string
+*   2. special characters are present
+*/
+export function containsIdentifier(segment: string): boolean {
+  return isAlphaNumericString(segment)
```
zengin, line 45: replace the implementation with `[A-Za-z0-9]+`

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,97 @@
...
+export function hasSpecialCharacters(segment: string): boolean {
+  const specialCharacters = ['.', '=', '@', '-'];
+  return specialCharacters.some((character) => segment.includes(character));
+}
 
-  // Redact PII
-  sanitizedUrl = sanitizedUrl.replace(emailRegex, '<email>');
-  return `${sanitizedUrl}${queryString}` ;
+export function isNumericString(segment: string): boolean {
+  const numberRegex = /^[0-9]*$/;
```
zengin, line 56: `^` is negation for selecting the character set. see regex101.com for testing the regular expressions.

src/app/utils/query-url-sanitization.ts
zengin: hasSpecialCharacters can be included in the full regex like this: `[A-Za-z0-9\.@=-]+`

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,97 @@
...
   // Extract query string
-  const { search } = parseSampleUrl(url);
+  const { search, queryVersion, requestUrl } = parseSampleUrl(url);
   const queryString: string = sanitizeQueryParameters(search);
 
-  // Drop query string
-  let sanitizedUrl = url.split('?')[0];
+  let resourceUrl = requestUrl;
+  const sections = resourceUrl.split('/');
+  sections.forEach(sec => {
```
zengin, line 16: I think the logic will be simpler if we check for non-id's and ignore the sanitization process on them.
Non-IDs as of today are all alpha characters, so `^[A-Za-z]+$` could be used to match a full string with alpha characters.
Another possible non-ID format that Michael raised is that the deprecation logic may have `segment_v2` temporarily and we shouldn't sanitize those parts too.

So something like:
```
if (isAllAlpha(sec) || isDeprecation(sec))
{
  // ignore sanitization
}

**Review by zengin — CHANGES_REQUESTED** [review]

File: src/app/utils/query-url-sanitization.ts (hunk 3 above)
Comment (thewahome) [reply src/app/utils/query-url-sanitization.ts]: Noted. I have recorded this in #745

File: src/tests/utils/sanitize-url.spec.tsx (hunk 4 above)
Comment (thewahome) [reply src/tests/utils/sanitize-url.spec.tsx]: Noted

**Comment by jobala** [comment]
@thewahome please add steps on how to test these changes.

File: src/app/utils/query-url-sanitization.ts [hunk 12]
```diff
@@ -1,67 +1,74 @@
...
 // Matches pattterns within quotes e.g "displayName: Gupta"
 const quotedTextRegex = /"([^"]*)"/g;
 
-// Matches PII patterns
-const numberRegex =  /(?<=\/)\d+\b/g; // number between forward slashes
-const emailRegex = /([a-z0-9_\-.+]+)@\w+(\.\w+)*/gi;
-const guidRegex =
- /(\{){0,1}[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}(\}){0,1}/gi;
+// matches strings that are all alphabets
+const allAlphaRegex = /^[A-Za-z]+$/;
```
jobala, line 8: If you use `alphabets` as the variable name, you won't need the comment above.

jobala (changes requested):
Add tests agains URLs with query strings.

src/app/utils/query-url-sanitization.ts
zengin: I think the comment is a good clarification because there is a detail that this regex matches only a full string as opposed to succeeding on a substring match too.

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,74 @@
...
+  return !!segment.match(allAlphaRegex);
+}
+
+/**
+ * @param segment part of the url string to test
+ * depracated resources may have `_v2` temporarily
+ * @returns boolean
+ */
+export function isDeprecation(segment: string): boolean {
+  return !!segment.match(depracationRegex);
```
zengin, line 62: typo: `deprecationRegex`

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,74 @@
...
 
-  // Drop query string
-  let sanitizedUrl = url.split('?')[0];
-
-  // Normalize parameters in param=<arbitraryKey> format with param={value} format
-  sanitizedUrl = sanitizedUrl.replace(functionParamInitialRegex, '<value>');
-  sanitizedUrl = sanitizedUrl.replace(functionParamFinalRegex, '<value>');
+  let resourceUrl = requestUrl;
+  const sections = requestUrl.split('/');
+  sections.forEach(segment => {
```
zengin, line 22: please add a comment here that mentions
```
what is skipped in the sanitization process:
- NonIDs that are url segments from entities/entity sets/navigations from the metadata. They are expected to contain only alpha characters.
- NonIDs that indicate deprecation. It is of the form <another_non_id>_v2.

Remaining url segments are assumed to be IDs that need to be sanitized.
```
If there is a bug in this it will be easy for the person investigating this to understand the assumptions we are making here.

zengin (approved):

src/app/utils/query-url-sanitization.ts
jobala: @zengin I missed the detail  in the comment and understood what the regex does after trying it out. @thewahome what do you think of adding what the regex matches  as part of the comment?

```
/**
 * will match: abc
 * won't match: ab2c
**/
```

MIchaelMainer (approved):

src/app/utils/query-url-sanitization.ts
```diff
@@ -1,67 +1,74 @@
...
+export function isAllAlpha(segment: string): boolean {
+  return !!segment.match(allAlphaRegex);
+}
+
+/**
+ * @param segment part of the url string to test
+ * depracated resources may have `_v2` temporarily
+ * @returns boolean
+ */
+export function isDeprecation(segment: string): boolean {
```
zengin, line 61: I have already signed off on this, but an edge case woke me up from my sleep last night :D

- An email can contain `_v2` and we shouldn't mark it as Non-ID in that case. We can either add a new condition here that the string doesn't include `@` symbol in addition to containing `_v2`. Or we can improve the regular expression for the deprecation regex to always start with only alpha characters before `_v2`, something like: `^[A-Za-z]+_v2$`

zengin (changes requested):

PR #746, opened 2020-10-22
remove window instrumentation key

thewahome:
## Overview

We have our instrumentation key that is unreachable by the function because the window instrumentation key is never empty.
We need to change it so that the staging environment can point to its own staging key


millicentachieng:
LGTM

millicentachieng (approved):

PR #759, opened 2020-11-10
Task: cache autocomplete results

### LINKED ISSUE #724: Autocomplete should cache URL patterns
**Describe the bug**
Autocomplete doesn't cache the results for a URL pattern. I see it making repeat APIs calls for the same URL pattern. 

**To Reproduce**
1. Open Fiddler, capture
2. Type `https://graph.microsoft.com/v1.0/users/123/`, delete `123/`, the add `123/`.

**Expected behavior**
No new API call is made as we already downloaded the autocomplete suggestions for that URL pattern.
 
[AB#6318](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6318)

thewahome:
## Overview

Uses indexed db to store the results of the autocomplete query so that subsequent calls do not call the API. This effectively speeds up autocomplete triggers based on the same url

Fixes #724

### Demo

image

### Notes

Request URLs with keys are replaced with generic IDs so that they are easily recreated for storage and searching in the cache.

### Tasks

- [x] Create store for autocomplete options
- [x] Read and write options to the store
- [x] Add a time to live for the options in the store

## Testing Instructions

* Run the app. Trigger an autocomplete search e.g `me/messages?`
* Using the developer console, notice that the request URL is placed as a key in the indexedDb storage
* perform another autocomplete search e.g `me/todo?`
* Using the developer console, notice that there was a network request that was made
* Redo `me/messages?` and notice that it is no longer calling the API


src/app/views/query-runner/query-input/auto-complete/auto-complete.cache.ts
```diff
@@ -0,0 +1,38 @@
...
+const autoCompleteStorage = localforage.createInstance({
+  storeName: 'autocomplete',
+  name: 'GE_V4'
+});
+
+export async function storeAutoCompleteContentInCache(content: IParsedOpenApiResponse) {
+  autoCompleteStorage.setItem(content.url, content);
+}
+
+export async function getAutoCompleteContentFromCache(url: string) {
```
jobala, line 13: Specify return type.

src/app/views/query-runner/query-input/auto-complete/auto-complete.cache.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+
+/**
+ * Autocomplete options may be changed by the API.
+ * We do not want to keep the options in the cache for too long
+ * @param createdAt
+ */
+function autoCompleteOptionHasExpired(createdAt: string): boolean {
+  const ageInDays: number = 7;
```
jobala, line 32: I think you can leave out the type here to be consistent or specify the type for all variable assignments.

jobala (changes requested):
It is possible to improve storage performance by storing the hash of a path instead of the path itself because in the worst case scenario the path can be really long.

jobala:
#### Design Proposal

@thewahome what do you think of abstracting the logic for fetching suggestions into a SuggestionProvider module that has the simple interface `getSuggestion(...)` then under the hood it will figure out whether to get the suggestion from cache or network.

Ideally, someone else with a similar use case should be able to use this module without understanding how it works.

Let me know what you think.

thewahome:
> #### Design Proposal
> @thewahome what do you think of abstracting the logic for fetching suggestions into a SuggestionProvider module that has the simple interface `getSuggestion(...)` then under the hood it will figure out whether to get the suggestion from cache or network.
> 
> Ideally, someone else with a similar use case should be able to use this module without understanding how it works.
> 
> Let me know what you think.

@jobala this sounds good and worth looking into... We can track this as technical debt?

jobala:
> @jobala this sounds good and worth looking into... We can track this as technical debt?

@thewahome if it was up to me I would say we do this before merging. I will leave it up to @ddyett and @MIchaelMainer to make the call.

src/modules/suggestions/suggestions.ts
```diff
@@ -0,0 +1,46 @@
...
+    // and returns them
+    const localOptions = await getSuggestionsFromCache(url);
+    if (localOptions) {
+      return getSuggestionsFromCache(url);
+    }
+
+    return this.fetchSuggestionsFromNetwork(url, api);
+  }
+
+  public async fetchSuggestionsFromNetwork(url: string, api: string): Promise<IParsedOpenApiResponse | null> {
```
jobala, line 20: This method should be private.

jobala (approved):

Issue #208, opened 2019-10-04
Collapse / Expand request/response components


ghost:
**Is your feature request related to a problem? Please describe.**
The request headers take up a lot of space especially when someone sends a GET request and has to scroll down to see the response. 
 
**Describe the solution you'd like**
Add a caret button on the left of both components that enables a user to expand the component when they want to see more details or collapse it when it is not mainly in use

**Describe alternatives you've considered**
To display the request and response components vertically as described in #133 

**Additional context**
When both are collapsed they can look like this:
![image](https://user-images.githubusercontent.com/53429535/66206488-f167e680-e6b8-11e9-9676-d04dc33a20cf.png)

When the response tabs is expanded
![image](https://user-images.githubusercontent.com/53429535/66206571-25430c00-e6b9-11e9-904a-d553f6642b8b.png)

When both are expanded
![image](https://user-images.githubusercontent.com/53429535/66206668-6affd480-e6b9-11e9-89d6-80340bab6cbb.png)

thewahome:
Closing this as #822 reduces the views to a collapsed state

PR #766, opened 2020-11-18
Feature: resizable components

### LINKED ISSUE #278: As a user, I want to be able to change the height of the request area and response area ratio
In Graph Explorer v3, I can hover over with my mouse the area between the request area and response area. I can click and then drag to make the height different. I would like to be able to do that in v4. 

The main reason for this is so that I can see all the permissions in the list visually without having to scroll. It's great when you are teaching and presenting on a screen.

Task:
#311

### LINKED ISSUE #630: Response body preview pane extends beyond view port
**Describe the bug**
Large response bodies scroll handle goes beyond view port.

**To Reproduce**
1. POST https://graph.microsoft.com/v1.0/me/calendar/getSchedule?$top=5 with a large enough time window that will have many results. https://docs.microsoft.com/en-us/graph/api/calendar-getschedule?view=graph-rest-1.0&tabs=http
2. Click on the response body and scroll down. The scroll handle will leave the view port as you scroll down for a sufficiently large response.

**Expected behavior**
The scroll handle stays within view port.

 - OS: Windows
 - Browser: Chrome

GE_ResponsePreview
 
[AB#7483](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7483)

### LINKED ISSUE #693: Feedback Bar for samples with clickable link
We need a new type of Message Bar in graph explorer that allows requesting feedback to our customers for specific queries.
Overview: [Onboarding survey to Graph Explorer](https://microsoft-my.sharepoint-df.com/:w:/p/sanooran/ERs8P6XhlzRCoLSQpFfcqYQBxEy4saBV-r3RvjZHjJynxg?e=22B1ZD)

This is the first Pull request attempting at this, using the Tip bar: https://github.com/microsoftgraph/microsoft-graph-devx-content/pull/51

**Requirements for the new Feedback Message Bar**

1. Allow clickable URL formatting (with Markdown, HTML code, or a custom tag you decide). This is needed for opening the survey.
2. All the links will be opened by default in a new tab.
3. Show the feedback bar independently of the user login: it should be shown even with the demo tenant data.
4. The message bar should persist after the user clicks "Run query".
5. The message bar should dismiss only when:
  - The user changes a sample
  - The user clicks X to close the bar.
  - The user clicks the link to open the survey.
6. If the user clicked the link or dismissed the bar with the X, the Message bar *for that sample* should not appear anymore. This can be achieved by storing a LOCAL cookie with the id of the sample when the user clicks the link or dismiss the feedback bar:
  - If the cookie is present next time the user clicks the sample, the feedback bar will not be shown.
  - If the user clears their cache / cookies, then the feedback bar will be shown again, nothing related to the user data should be store server-side.

@thewahome @bettirosengugi
[AB#6062](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6062)

thewahome:
## Overview
Enables the resizing of the response and request components so that a user has the ability to increase the size of the views. 
Increasing the size of the request component automatically increases the size of the response view

Closes #278 
Fixes #630 

### Demo

### Notes

## Testing Instructions

* Hover over the bottom border of the request area
* Notice the presence of the resize handles
* Resize the component up or down.


darrelmiller:
Hey @thewahome it is great to see this work progressing.  I took a quick look at the latest deployment of this feature and I seemed to run into some issues.  Here's the video https://microsoft-my.sharepoint-df.com/:v:/p/darrmi/EVSWOhI-oMtNkg4eOLExBfYB-v624hZOn-6VbPiNP0dFrA?e=NfJVnH

- There is a lower resize point on the response window that doesn't seem to work as intended. I'm not sure if this is needed.
- The upper resize point at the top of the notification message resizes the lower portion of the screen but does not resize the request body text box.
- The sidebar resize seems to get stuck if you resize it too small.

thewahome:
Thanks so much for taking a look at this 🙂... I will check out the video @darrelmiller

millicentachieng:
@thewahome Could you please add a description of what this PR aims to do?

ElinorW (commented):
@thewahome when one clicks on the other headers e.g Request Header, the page breaks

src/app/views/query-runner/request/permissions/Permission.tsx
```diff
@@ -280,7 +284,8 @@ export class Permission extends Component<IPermissionProps, IPermissionState> {
             {!panel && <TabList
               permissions={permissions}
               columns={this.getColumns()}
-              classes={classes}
+              maxHeight={tabHeight}
+              classes={''}
```
ElinorW, line 288: this is empty, is it needed?

src/app/views/query-runner/QueryRunner.tsx
```diff
@@ -125,7 +128,7 @@ export class QueryRunner extends Component<
 
   public render() {
     return (
-      <div>
+      <>
```
ElinorW, line 131: this tag seems to be empty

package.json
```diff
@@ -58,6 +58,7 @@
     "postcss-loader": "3.0.0",
     "postcss-preset-env": "6.5.0",
     "postcss-safe-parser": "4.0.1",
+    "re-resizable": "^6.9.0",
```
ElinorW, line 61: remove caret

jobala (changes requested):
@thewahome 

I found the following issues

**The Request and QueryInput components are not spaced.**

image

**Share and expand are no longer at the top right of the Response component**
image

**Spacing issue in the permission component**

image


**ux**
Most products have a divider which is dragged to resize windows, I think we should do the same for Graph Explorer. Instead of dragging Request component's bottom border to resize, we should have a divider between Request and Response component.

darrelmiller:
Why can I not shrink the request area smaller than this?  
image

src/app/views/query-runner/QueryRunner.tsx
thewahome: Yeah it is called a [fragment](https://reactjs.org/docs/fragments.html). I didn't want to add extra items to the DOM

thewahome:
> Why can I not shrink the request area smaller than this?
> image

src/types/dimensions.ts
```diff
@@ -0,0 +1,11 @@
+export interface IDimensionProperies {
+    width: string;
+    height: string;
+}
+
+export interface IDimensions {
+    sidebar: IDimensionProperies;
```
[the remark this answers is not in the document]
thewahome: Not now. Removing this so that if required, it can be added back later

thewahome:
> Why can I not shrink the request area smaller than this?
> image

The intention was to have some of the items to remain visible even after resize @darrelmiller

ddyett:
if i run a request then try changing this there's flashing in the response section due to reformatting.  run a request->have a response loaded->hold down left mouse button and keep resizing.

ddyett:
it feels pretty limited in terms of how i can get the request section out of my way if I don't want it.  Have we looked at something like an ability to expand/collapse the entire section?

src/app/views/common/dimensions-adjustment.ts
```diff
@@ -0,0 +1,5 @@
+export function convertVhToPx(height: string, adjustment: any) {
```
ddyett, line 1: doesn't this have to be a number?

src/app/views/common/monaco/Monaco.tsx
```diff
@@ -32,19 +33,24 @@ function formatDocument(editor: any) {
...
 
   let { body } = props;
-  const { onChange, language, readOnly } = props;
+  const { onChange, language, readOnly, height } = props;
 
   if (body && typeof body !== 'string') {
     body = JSON.stringify(body);
   }
 
+  let itemHeight = '300px';
```
ddyett, line 42: let itemHeight = height ?? '300px'

src/app/views/query-response/pivot-items/pivot-items.tsx
```diff
@@ -23,7 +24,10 @@ import { Snippets } from '../snippets';
 export const getPivotItems = (properties: any) => {
 
   const { headers, body, mode, sampleQuery } = properties;
-  const resultComponent = displayResultComponent(headers, body);
+  let { height } = properties;
```
ddyett, line 27: this makes height and any type so we lost all typing support. just pass the height into the convert directly.

src/app/views/query-response/pivot-items/pivot-items.tsx
```diff
@@ -59,6 +63,40 @@ export const getPivotItems = (properties: any) => {
     );
   }
 
+  function displayResultComponent() {
```
ddyett, line 66: seems like we're calling this every time we resize so it may be causing the weirdness I see when I have content loaded in the response and try to resize.

src/app/views/query-runner/request/Request.tsx
```diff
@@ -113,19 +118,47 @@ export class Request extends Component<IRequestComponent, any> {
     });
   }
 
+  private setRequestAndResponseHeights = (requestHeight: string) => {
+    const dimen = { ...this.props.dimensions };
+    dimen.request.height = requestHeight;
+    const response = 90 - parseFloat(requestHeight.replace('vh', ''));
```
ddyett, line 124: should probably name this variable so it's clear what the 90 represents.

src/app/views/query-runner/request/Request.tsx
```diff
@@ -113,19 +118,47 @@ export class Request extends Component<IRequestComponent, any> {
...
+        style={{
+          border: 'solid 1px #ddd',
+          marginBottom: 10,
+        }}
+        onResize={(e: any, direction: any, ref: any, d: any) => {
+          if (ref && ref.style && ref.style.height) {
+            this.setRequestAndResponseHeights(ref.style.height);
+          }
+        }}
+        maxHeight={800}
```
ddyett, line 144: probably should be constant values

src/app/views/query-runner/request/permissions/TabList.tsx
```diff
@@ -11,11 +11,12 @@ interface ITabList {
   classes: any;
   renderItemColumn: Function;
   renderDetailsHeader: Function;
+  maxHeight: any;
```
ddyett, line 14: string

src/app/views/sidebar/Sidebar.styles.ts
```diff
@@ -4,7 +4,7 @@ import {
 } from '@uifabric/styling';
 
 export const sidebarStyles = (theme: ITheme) => {
-  const pageHeight = '660px';
+  const pageHeight = '60vh';
```
ddyett, line 7: it's a little difficult to track which variable carry vh and which carry px so it may be useful to indicate it in the name.

thewahome:
> it feels pretty limited in terms of how i can get the request section out of my way if I don't want it. Have we looked at something like an ability to expand/collapse the entire section?

#208 had described this use case. We can re-open this and track it as a new task

ElinorW (approved):

jobala (approved):

PR #775, opened 2020-11-30
Fix: sample query sample body

thewahome:
## Overview

Noticed that the sample body was not displayed as HTML in the editor.
Enhanced code to get and display data in the DetailList to a reusable function

### Notes

Refactoring has been done to generate groups from a list into a reusable function so that the logic is handled in one place. 

## Testing Instructions

* Log in to graph explorer
* search for "create page" in the samples
* Click the sample
* Notice the HTML is displayed


millicentachieng (dismissed):

millicentachieng (commented):
There's an error in the editor: `Expected a JSON object, array or literal`. Is there a way to have this can be handled?

Also, are we expecting any other formats other than JSON and HTML?

millicentachieng:
image

thewahome:
@millicentachieng this is expected . We do not specify the type of content to display in the request body at the moment, so we default to a string. The editor is able to format it properly. Which works.

PR #777, opened 2020-12-04
Track errors 

millicentachieng:
## Overview

This PR aims to add functionality for collecting error data within the app.

Here's an example of how we capture error data for a link that failed to op
```
telemetry.trackException(
  new Error(LINK_ERROR), 
  SeverityLevel.Error, 
  {
    ComponentName: 'Documentation link',
    SampleId: 'ef1b25b4-e6c7-4e91-88a1-6358f5fff714',
    Link: 'https://developer.microsoft.com/en-us/graph/docs/api-reference/v1.0/resources/users'
  });
```

The categories of errors to be captured will be:
-	LINK_ERROR 
Errors in external links that have been malformed or that return status code that do not indicate success
-	COMPONENT_ERROR
Errors when a component fails to render
-	NETWORK_ERROR
Errors from HTTP requests
-	OPERATIONAL_ERROR
Any other errors caught within code that are temporary e.g. those originating from user input


thewahome (approved):

src/app/utils/graph-toolkit-lookup.ts
```diff
@@ -1,7 +1,33 @@
+import { SeverityLevel } from '@microsoft/applicationinsights-web';
 import templates from '../../graph-toolkit-examples';
+import { telemetry } from '../../telemetry';
+import { LINK_ERROR } from '../../telemetry/error-types';
 import { IQuery } from '../../types/query-runner';
+import { sanitizeQueryUrl } from './query-url-sanitization';
 import { parseSampleUrl } from './sample-url-generation';
 
+
+export async function validateToolkitUrl(url: string, sampleQuery: IQuery): Promise<void> {
```
MIchaelMainer, line 10: Is this the only scenario for which we'd want to validate URLs (now and into the future)? I suggest pulling this out of this file and make this generally available for all components to validate URLs.

ddyett:
is there a blocker on getting this in?

src/app/utils/graph-toolkit-lookup.ts
```diff
@@ -1,7 +1,33 @@
...
+          throw Error(response.statusText);
+        }
+      })
+      .catch(error => {
+        const sanitizedUrl = sanitizeQueryUrl(sampleQuery.sampleUrl);
+        telemetry.trackException(
+          new Error(LINK_ERROR),
+          SeverityLevel.Error,
+          {
+            ComponentName: 'Graph toolkit link',
```
jobala, line 23: What do you think of making the component names constants? It will make it easier to update them and will give us the added benefit of knowing what components are instrumented.

src/app/utils/graph-toolkit-lookup.ts
millicentachieng: I've moved the method to a separate file.

src/app/utils/graph-toolkit-lookup.ts
millicentachieng: I've done this, thanks

PR #778, opened 2020-12-04
Fix: stop display of options when fetching suggestions

thewahome:
## Overview

Noticed that as the request for autocomplete options is running, the results of the previous run are displayed.
If the call is running, only the spinner should show until the content is found or not.


ElinorW (approved):
LGTM!

PR #790, opened 2021-01-11
Fix: modify permissions tab UI

### LINKED ISSUE #653: Improve user experience on the "Modify permissions" tab
The Modify permissions tab is congested and sometimes confusing for users to use. The following tasks will help improve the user experience on the tab:

Remove the Display string column since the information on this column is repeated on the description column
Before the customer signs in - replace the following UI text by one that identifies the appropriate next step, which is to sign in with an account that can consent to the selected permission.
Replace this:
“The following permissions are required to run the query. To consent to the permissions, click Consent.”
By:
“One of the following permissions is required to run the query. Sign in with an account that can consent to the permission you will choose.”
After the customer signs in – update the UI text to the following:
“One of the following permissions is required to run the query. Select a permission and click Consent.”
When no permissions are listed on the Modify Permissions tab, add the message to redirect users to "Select permissions" option under settings.
Wrap contents on the "Description" column to take even 3 lines to ensure scrolling to the right is not necessary
[AB#5626](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/5626)

ElinorW:
## Overview

- Remove 'display string' column
- Wrap information contained under 'Description' column
- Change description message dependent on whether a user has signed in or not

Fixes #653

### Demo

image

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/src/app/views/query-runner/request/permissions/TabList.tsx
+++ b/src/app/views/query-runner/request/permissions/TabList.tsx
@@ -10,16 +10,20 @@ interface ITabList {
   classes: any;
   renderItemColumn: Function;
   renderDetailsHeader: Function;
+  permissionToken: boolean;
 }
 
-const TabList = ({ permissions, columns, classes, renderItemColumn, renderDetailsHeader }: ITabList) => {
+
+const TabList = ({ permissions, columns, classes, renderItemColumn, renderDetailsHeader, permissionToken }
+  : ITabList) => {
   return (
     <>
       <Label className={classes.permissionLength}>
         <FormattedMessage id='Permissions' />&nbsp;({permissions.length})
       </Label>
       <Label className={classes.permissionText}>
-        <FormattedMessage id='permissions required to run the query' />
+        {!permissionToken && <FormattedMessage id='sign in to consent to permissions' />}
+        {permissionToken && <FormattedMessage id='permissions required to run the query' />}
       </Label>
       <DetailsList styles={{ root: { minHeight: '300px' } }}
         items={permissions}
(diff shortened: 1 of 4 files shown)


src/app/views/query-runner/request/permissions/TabList.tsx
```diff
@@ -10,16 +10,20 @@ interface ITabList {
   classes: any;
   renderItemColumn: Function;
   renderDetailsHeader: Function;
+  permissionToken: boolean;
 }
 
-const TabList = ({ permissions, columns, classes, renderItemColumn, renderDetailsHeader }: ITabList) => {
+
+const TabList = ({ permissions, columns, classes, renderItemColumn, renderDetailsHeader, permissionToken }
```
thewahome, line 17: Using the [useSelector](https://react-redux.js.org/api/hooks#useselector) React Hook may help to reduce passing too many arguments into the TabList component since it has access to the redux state. This can be used to pick up whether the user is logged in or not.

thewahome (changes requested):

src/messages/GE.json
```diff
@@ -306,7 +306,7 @@
   "Snippet not available": "Snippet not available",
   "Select different permissions": "To try out different Microsoft Graph API endpoints, choose the permissions, and then click Consent.",
   "Search permissions": "Search permissions",
-  "permissions not found": "We did not find permissions",
+  "permissions not found": "We did not find permissions. To view a list of all available permissions, click on the Settings gearbox and then click on Select Permissions.",
```
ddyett, line 309: is gearbox a thing people understand?  just never seen it before

jobala:
Admin consent column is off.


image

src/messages/GE.json
ElinorW: I will rename that to settings icon

src/app/views/query-runner/request/permissions/TabList.tsx
ElinorW: Ah okay perfect! Thanks for this!

thewahome (approved):

PR #813, opened 2021-01-27
Feature/additional telemetry

millicentachieng:
- Added additional telemetry
- Added `itemKey` as property on pivot item to be used to construct component name for telemetry. Initially, I was using the `title` property but this would be translated into multiple languages depending on locale thus making querying collected data difficult.


src/app/views/App.tsx
```diff
@@ -8,6 +8,9 @@ import { connect } from 'react-redux';
 import { bindActionCreators, Dispatch } from 'redux';
 
 import { geLocale } from '../../appLocale';
+import { telemetry } from '../../telemetry';
+import { SIDEBAR_HAMBURGER_BUTTON } from '../../telemetry/component-names';
+import { BUTTON_CLICK_EVENT } from '../../telemetry/event-types';
```
thewahome, line 13: Just a question, No action required. 
Can the constants be exported together so that we only have one import statement related to telemetry?

millicentachieng: It will only mean that we have to introduce the idea of namespaces in our code

thewahome (approved):

PR #814, opened 2021-01-27
Fix: support different content types

### LINKED ISSUE #626: Support all different content types and fallback to showing text
The default behavior should be to render as text.  We should have a case for `application/octet-stream` that doesn't render it as text and just shows a message such as "Binary Response".  Ideally this would allow a download of that binary response.

https://github.com/microsoftgraph/microsoft-graph-explorer-v4/blob/e3ce0206d787fef1928f2b6821e607996e2773d2/src/app/services/actions/query-action-creator-util.ts#L79-L90

This will fix the rendering of report endpoints that return text/csv and will future proof us for other text formats that might be used.   
[AB#6253](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6253)

thewahome:
## Overview

Partially fixes #626 
Now successfully supports HTML responses.

### Demo

Request: https://graph.microsoft.com/v1.0/me/onenote/pages/{pages-id}/content

Before
image

After
image

### Notes

I can not conclusively say that it fixes #626 since I have not been able to test any call that returns an `edm.stream` response. It errors out with a 403, which is weird since I already have the required permission consented to. 

image

## Testing Instructions

* Run this query: https://graph.microsoft.com/v1.0/me/onenote/pages/{pages-id}/content after replacing the `{pages-id}` token with the appropriate page id
* Notice the HTML response is currently visible


jobala:
> I can not conclusively say that it fixes #626 since I have not been able to test any call that returns an edm.stream response. It errors out with a 403, which is weird since I already have the required permission consented to.

@thewahome you can create an issue to track edm.stream responses separately so that we don't stop tracking it once you merge this in.

jobala (dismissed):

jobala:
Not sure how useful to the user showing html is, what do you guys think for previewing the HTML? @thewahome @ddyett ?

thewahome:
> Not sure how useful to the user showing html is, what do you guys think for previewing the HTML? @thewahome @ddyett ?

Could be a good idea... We could go for a split screen for code and preview.

ddyett:
> > Not sure how useful to the user showing html is, what do you guys think for previewing the HTML? @thewahome @ddyett ?
> 
> Could be a good idea... We could go for a split screen for code and preview.

should open an issue to track rendering the html

thewahome:
> > > Not sure how useful to the user showing html is, what do you guys think for previewing the HTML? @thewahome @ddyett ?
> > 
> > 
> > Could be a good idea... We could go for a split screen for code and preview.
> 
> should open an issue to track rendering the html

Done. #826

PR #815, opened 2021-01-27
Fix: permissions tab UI

### LINKED ISSUE #560: If there are no permissions on "Modify Permissions" tab redirect users to Select Permissions option
The Modify permissions tab does not have the full list of all permissions  since the document we are using doesn't have the full list of permissions and their end points linked. 

If no permissions are available show a message directing users to the Select permissions options under settings. 

**Additional context**
 
[AB#7505](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7505)

### LINKED ISSUE #739: Change the wording on Modify permissions tab
There are some queries that do not have permissions when you click on Modify permissions and the messaging we are relaying to our users is that the tab is experimental and may have missing permissions and redirect them to select permissions for that reason there is need to:

1. Change the title of the tab to "Modify permissions (preview)"
2. Add a tooltip with the string "This feature is in preview and might not currently show all permissions for some queries"
3. If no permissions are returned on the tab include the message "Permissions for the query are missing on this tab, use Select permissions option in the settings gear next to your profile to see the full list of Microsoft Graph permissions and select the permission(s) you want and consent to them from there. " 
[AB#7466](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7466)

ElinorW:
## Overview

- [x] Rename the permissions tab.
- [x] Add a tooltip explaining why the feature is on preview mode.
- [x] Redirect users to a full list of permissions if none are displayed on the permissions tab.

Fixes #739 
fixes #560 

### Demo
1. Permissions tab with tooltip
image

2. Permissions tab when there are no permissions (User signed in)
image

3. Permissions tab when there are no permissions (User not signed in)
image


src/messages/GE.json
```diff
@@ -334,7 +334,7 @@
   "Sidebar minimized": "Sidebar minimized",
   "Sidebar maximized": "Sidebar maximized",
   "Response area expanded": "Response area expanded",
-  "Close expanded response area": "Close expanded response area"
-  
-}
-
+  "Close expanded response area": "Close expanded response area",
+  "permissions not found in permissions tab": "Permissions for the query are missing on this tab, use Select permissions option in the settings icon next to your profile to see the full list of Microsoft Graph permissions and select the permission(s) you want and consent to them from there.",
```
thewahome, line 338: This message may not apply to someone who is not logged in since they would not be able to see that button. How about splitting this message further so that it can be displayed on the condition that someone is logged in? Or having the message explain that they'd have to log in?

ElinorW: Good point. I will create a separate message for a user that isn't logged in.

jobala:
@ElinorW it will be a better user experience to be able to open the permissions panel from the the modify permissions tab instead of showing them instructions on how to open the permissions panel.

src/app/views/query-runner/request/permissions/TabList.tsx
```diff
@@ -65,3 +68,17 @@ function displayNoPermissionsFoundMessage() {
     <FormattedMessage id='permissions not found in permissions tab' />
   </Label>);
 }
+
+function displayNotSignedInMessage() {
+  return (<Label style={{
```
thewahome, line 73: Let's move these inline styles to the styles file

src/app/views/query-runner/request/permissions/TabList.tsx
```diff
@@ -55,11 +58,27 @@ export default TabList;
 function displayNoPermissionsFoundMessage() {
   return (<Label style={{
     display: 'flex',
+    paddingLeft: '10px',
```
thewahome, line 61: Let's move the inline styles to a styles file

thewahome (approved):

PR #822, opened 2021-01-29
Fix: shrink request section

### LINKED ISSUE #818: I should be able to shrink the request section in order to show more response area
Today I cannot shrink the section enough.  It's munimum height is a fair amount.  I'd like to be able to almost just have the request url box and the response immediately below it. 
[AB#7820](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7820)

thewahome:
## Overview

Shrinks the request component to just the pivot item labels.

Fixes #818 

### Demo

image

### Notes

N/A

## Testing Instructions

* Resize and notice that the request component is not in the way


src/app/views/query-runner/request/headers/Headers.styles.ts
```diff
@@ -3,11 +3,11 @@ import { FontSizes } from '@uifabric/styling';
 export const headerStyles = () => {
   return {
     container: {
-      width: '100%',
+      // width: '98%',
```
ElinorW, line 6: commented style can be deleted if not in use

ElinorW (approved):

PR #835, opened 2021-02-09
Feature: clickable links on message bar

### LINKED ISSUE #693: Feedback Bar for samples with clickable link

### LINKED ISSUE #705: Graph links in tip messages to populate the url address bar
### Is your feature request related to a problem? Please describe.
A sample query may contain a tip message that describes what to do with the query. A common scenario is a query can have a **token** to display an item in the url that can be replaced eg `{message-id}` or `{task-id}`. A tip message is used here to tell the user what query to run in order to get the data to replace the **token** with.

The user then highlights the url in the tip message, copies it into the url editor and runs the query to get the data they can use to replace the **token**.

### Describe the solution you'd like
The ask here is that once a tip message is displayed, the url described in the message can be made clickable and once clicked, it should replace the url displayed on the url editor

### Describe alternatives you've considered
Leaving it as is and letting the user copy the link and paste it themselves
[AB#6125](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6125)

thewahome:
## Overview
Makes links available in the samples clickable

Fixes #693 
Fixes #705 

### Demo

Has clickable link: 
image

### Notes

Some tip messages will not be shown when the user is not signed in

#### Scenario 1 - External link
* Click the `list Azure AD gallery apps` example from the sidebar
* A tip message with a clickable link should be displayed
* The click should open a new tab to the feedback survey

#### Scenario 2 - Graph API link
* Run the `get ediscovery case` example from the sidebar
* A tip message with a clickable link should be displayed
* The click should populate the query input area with the clicked url


package.json
```diff
@@ -81,7 +81,7 @@
   "scripts": {
     "start": "node scripts/start.js",
     "build": "node scripts/build.js && node versioned-build.js",
-    "test": "node scripts/test.js --no-watch --testPathIgnorePatterns=src/tests/accessibility",
+    "test": "node scripts/test.js --watch --testPathIgnorePatterns=src/tests/accessibility",
```
ElinorW, line 84: why did you change this part?
won't it look for the accessibility test, which is a separate job?

thewahome: the `--watch` flag makes the tests watch for changes as you work. So if you change a file, as long as you had ran `npm run test` you would see the results of the tests immediately without running it again. with the `--no-watch` you need to run the tests command again

The `--testPathIgnorePatterns` flag is a different parameter that helps the runner know which file to ignore

ElinorW (approved):

jobala (changes requested):
Why is the filename `status.util.ts` and not `status.ts`?

src/app/utils/status.util.ts
```diff
@@ -0,0 +1,46 @@
...
+      return;
+    }
+  });
+  return exists;
+}
+
+export function extractUrl(value: string): any {
+  const matches = value.match(/\bhttps?:\/\/\S+/gi);
+  if (matches) {
+    return matches;
```
jobala, line 43: You can achieve the same with

```js
return value.match(...)
```

src/tests/utils/url-to-html.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { linkExists, extractUrl, replaceLinks, convertArrayToObject } from "../../app/utils/status.util";
+
+describe('url should', () => {
```
jobala, line 3: It's a best practise to put the test suit name here, something like `URL TestSuite`

jobala (approved):

PR #855, opened 2021-02-18
fix: autocomplete-onfocusout

### LINKED ISSUE #721: Autocomplete doesn't collapse onFocusOut
**Describe the bug**
Autocomplete dropdown UI doesn't auto close when the address bar loses focus.

**To Reproduce**
1. Add a `/` in the URL bar
2. Click on any place else in the UI
3. See that the autocomplete UI doesn't close

The only way autocomplete closes is if I select an autocomplete option.

**Expected behavior**
When the URL bar loses focus, autocomplete dialog closes. 

**Screenshots**

image

 - OS: Windows
 - Browsers: Edge and Chrome

 
[AB#7463](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7463)

thewahome:
## Overview
Autocomplete suggestions should collapse when the autocomplete component loses focus
Fixes #721 

### Demo

### Notes

OnFocusOut is not supported in React hence the use of onBlur. 

## Testing Instructions

* Trigger an autocomplete call
* Once the options show up click outside the component
* Notice that the suggestions close

### DIFF
--- a/src/app/views/query-runner/query-input/auto-complete/AutoComplete.tsx
+++ b/src/app/views/query-runner/query-input/auto-complete/AutoComplete.tsx
@@ -13,14 +13,12 @@ import { sanitizeQueryUrl } from '../../../../utils/query-url-sanitization';
 import { parseSampleUrl } from '../../../../utils/sample-url-generation';
 import { translateMessage } from '../../../../utils/translate-messages';
 import { queryInputStyles } from '../QueryInput.styles';
-
-
 import {
   cleanUpSelectedSuggestion, getLastCharacterOf,
+  getLastSymbolInUrl,
   getParametersWithVerb
 } from './auto-complete.util';
 import SuggestionsList from './SuggestionsList';
-
 
 class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
   private autoCompleteRef: React.RefObject<ITextField>;
@@ -53,7 +51,7 @@ class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
     }, 10);
   }
 
-  public onBlur = (e: any) => {
+  public updateUrlContent = (e: any) => {
     const userInput = e.target.value;
     this.props.contentChanged(userInput);
   };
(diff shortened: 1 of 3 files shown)


src/app/views/query-runner/query-input/auto-complete/AutoComplete.tsx
```diff
@@ -402,6 +402,14 @@ class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
     return max;
   }
 
+  onFocusOut = (event: any) => {
```
jobala, line 405: What do you think of renaming this function to closeSuggestionDialog?

thewahome: Its purpose isn't to close the dialog. It's to monitor the focusOut event of the div and its children. closing the dialog could be the name of the function that closes the dialog and reused however since it's already a one-liner implementation that is explanatory, I do not exactly know the effect

jobala: From a user's perspective, what happens when the function is called?

thewahome: I see what you mean. What do you think of `handleClickOutside`? It encapsulates that we are handling a user's activity when they click outside. The reader may be misinformed by `closeSuggestionDialog` since they'd expect that this function is also called when someone selects a suggestion

ElinorW:
There's an issue with regards to accessibility here, a user navigating the site using the Tab key won't be able to write a query or even edit one in the query input field.
It's inaccessible using the Tab key.

src/app/views/query-runner/query-input/auto-complete/AutoComplete.tsx
jobala: @thewahome `handleClickOutside` does not communicate the intent because it doesn't clearly answer the question what does this function do? Can we rename the function that handles selecting a suggestion to `selectSuggestion` so as to avoid the ambiguity.

jobala (approved):

ElinorW (approved):

PR #875, opened 2021-03-03
Track time it takes to get a response from DevX API

millicentachieng:
## Overview

Add a custom event which allows us to capture the time it takes for a call to our DevX API to return a response.

Application Insights JS SDK offers two methods to use to track events:

- `appInsights.startTrackEvent(eventName)`, to be called just before the tracked call
- `appInsights.stopTrackEvent(eventName, properties)` to be called just after the tracked call

Application Insights will add a custom property `duration` with the API call duration in milliseconds.

## Testing Instructions

* Open Developer Tools on browser then go Network tab
* Fetching samples, code snippets, permissions or autocomplete options should trigger a tracked API call
* Inspect data sent to Application Insights. Below is an example of how the data should look like:
image


jobala (commented):
## Suggestion

Instead of using `trackApiCallEvent` to make network requests consider overriding `fetch` because developers won't have to remember to use `trackApiCallEvent` instead of `fetch` and this approach also maintains the intuitiveness of using `fetch` for network calls.

You can override fetch with the following snippet.
```javascript
windows.fetch = telemetry.trackedApiCall
```
Remember to do this in App.tsx.

zengin:
@jobala I wouldn't recommend changing a global function at the JavaScript API level. `trackApiCallEvent` is a descriptive name implying what it is doing. This way, we also make sure that telemetry is "opt-in". Otherwise developers may start emitting telemetry unintentionally if `fetch` is redefined. I understand that optional `componentName` can act as the arbiter of this decision, but that is more implicit and easy to cause a mistake.

@millicentachieng network calls to DevX API are failing when I use the static deployment. Constructed URL seems to be wrong. Is that expected?
```
Request URL: https://jolly-sand-0ac78c710-875.centralus.azurestaticapps.net/[object%20Object]/api/graphexplorersnippets?lang=javascript
```

millicentachieng:
Thanks for the feedback @zengin and @jobala.

I'm leaning towards having the tracking being opt-in as well. @ddyett, what's your take?

Regarding the API call for fetching snippets, it's an [issue](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/876) that's been fixed not too long ago. I'll merge the fix into this branch.

src/telemetry/telemetry.ts
```diff
@@ -65,6 +66,27 @@ class Telemetry implements ITelemetry {
     validateExternalLink(url, componentName);
   }
 
+  public async trackApiCallEvent(componentName: string, url: string,
+    options: IRequestOptions): Promise<Response> {
+    const properties: { [key: string]: string } = {
+      ComponentName: componentName,
+      IsRequestSuccessful: 'false',
+      Url: url
```
ddyett, line 74: shouldn't this be sanitized, as it contains query parameters that could have sensitive info.

ddyett:
`I'm leaning towards having the tracking being opt-in as well. @ddyett, what's your take?`

I wouldn't globally override either.  My expectation would have been that trackApiCallEvent accepts the function so it's explicit what's been called. That way you can just expose from telelmetry a trackevent and it becomes non-specific to making a web call or some other action we want to track.

thewahome (commented):
### Suggestion 2: Middleware

We can take advantage of redux middleware which will allow us to keep the code as-is and add the side-effects in a more react way.

Middleware has access to the state and action properties which would be useful to create the custom event. It would also give us the advantage of one source of truth for when we need to change/add/remove a parameter to track.

Here is a[ short introduction to redux middleware.](https://www.metaltoad.com/blog/overview-redux-middleware-react-applications)

src/telemetry/telemetry.ts
millicentachieng: Yes, I think the url will need to be sanitized. I'll check to see if there more that I need to do other than stripping out query parameter values?

jobala:
@millicentachieng 

It makes sense for users to explicitly opt in to track requests and we can make the design better by separating the tracking and network request functionality because it is not obivous from the name that `trackApiCallEvent` also makes a network call.

The **tracking** and **network call** can be separated using the **decorator pattern** and we'll have the added advantage of not modifying `fetch` calls to `trackApiCallEvent` calls. The code is show below

```typescript
@telemetry.trackApiCallEvent
fetch(...)
```

The **decorator implementation** below shows how you can use a decorator to get the elapsed time.

```typescript
const trackApiCallEvent = (
  target: Object,
  propertyKey: string,
  descriptor: PropertyDescriptor
) => {
  const method = descriptor.value;

  descriptor.value = function (...args) {
    const start = ...
    const result = method.apply(this, args);
    const finish = ...
    const elapsedTime = finish - start
    return result;
  };
```

millicentachieng:
@jobala, @thewahome I've reviewed your comments on using a decorator or middleware but I have another suggestion. Application Insights has a setting where you can enable auto-tracking of fetch calls. The result is what we are looking for.  Here's a sample of the data collected:

```
{
    "baseType": "RemoteDependencyData",
    "baseData": {
        "id": "|eca9e4e2d022464eb9291ffe5cb60194.e3e36b21410349b5",
        "ver": 2,
        "name": "GET https://graphexplorerapi.azurewebsites.net/samples",
        "resultCode": "200",
        "duration": "00:00:01.835",
        "success": true,
        "data": "GET https://graphexplorerapi.azurewebsites.net/samples",
        "target": "graphexplorerapi.azurewebsites.net",
        "type": "Fetch",
        "properties": {
            "HttpMethod": "GET",
            "ApplicationName": "Graph Explorer v4",
            "IsAuthenticated": "true"
        },
        "measurements": {}
    }
}
```
The duration and result of the call is captured. 

We can then have a processor that filter's out fetch requests that we do not need.

thewahome:
This looks good!! Would the preprocessor be able to also filter out the url contents for PII?

thewahome:
Do we plan on using it for all our external calls? Including the run query?

millicentachieng:
@thewahome Yes, the processor will sanitize the urls.

millicentachieng:
It also captures results from run query. I'm thinking that can be an important metric too to see which queries have a high failure rate, whether caused by missing permissions or wrongly formatted query.

millicentachieng:
@thewahome, @jobala Please review the new changes

src/telemetry/telemetry.ts
```diff
@@ -112,11 +92,34 @@ class Telemetry implements ITelemetry {
...
+    ];
+    return telemetryTypesToInclude.includes(baseType);
+  }
+
+  private filterRemoteDependencyData(envelope: ITelemetryItem) {
+    const baseData = envelope.baseData || {};
+    if (envelope.baseType === 'RemoteDependencyData') {
+      const targetsToInclude = {
+        GRAPHAPI: 'graph.microsoft.com',
+        DEVXAPI: 'graphexplorerapi.azurewebsites.net'
```
thewahome, line 110: These urls are present in the graph-constants.ts file. They can be called from there directly

thewahome: Sometimes `https://proxy.apisandbox.msdn.microsoft.com` when making a call as an anonymous guest. We should capture that too here

src/telemetry/filters.ts
```diff
@@ -0,0 +1,74 @@
...
+  const baseData = envelope.baseData || {};
+  if (envelope.baseType !== 'RemoteDependencyData') {
+    return true;
+  }
+
+  const targetsToInclude = [GRAPH_URL, DEVX_API_URL, GRAPH_API_SANDBOX_URL];
+  const urlObject = new URL(baseData.target || '');
+  if (!targetsToInclude.includes(urlObject.origin)) {
+    return false;
+  }
```
thewahome, line 34: It could be more likely for the `urlObject.origin` to contain the base URLs than the other way round.

millicentachieng: The aim is to compare base URLs on both sides

src/telemetry/filters.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  // Checks if user is authenticated
+  const accessTokenKey = 'msal.idtoken';
+  const accessToken = localStorage.getItem(accessTokenKey);
+  telemetryItem.properties.IsAuthenticated = accessToken ? true : false;
+
+  return true;
+}
+
+export function modifyTelemetryItemProperties(envelope: ITelemetryItem) {
```
MIchaelMainer, line 64: I strongly suggest that we rename this `sanitizeUrlFragment`.

src/telemetry/filters.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+  return true;
+}
+
+export function addCommonTelemetryItemProperties(envelope: ITelemetryItem) {
+  const telemetryItem = envelope.baseData || {};
+  telemetryItem.properties = telemetryItem.properties || {};
+
+  // Identifies the source of telemetry events
+  telemetryItem.properties.ApplicationName = 'Graph Explorer v4';
```
MIchaelMainer, line 54: We should capture the version of Graph Explorer that is emitting these events. After every deployment of Graph Explorer, there is a period where there are older versions of Graph Explorer loaded into a user's browsers and CDN caches haven't been updated with the latest bundle. We could have an old and new version emitting events at the same time. Capturing information about version and other variation identifiers could be useful in the future in case we want to do AB testing.

MIchaelMainer (commented):
Have we performed a review of the Application Insights data captured with these changes to verify that we aren't capturing customer data? What were the user actions in the staging site and did they exercise all of the changes in the this PR?

src/telemetry/filters.ts
```diff
@@ -0,0 +1,74 @@
+import { ITelemetryItem } from '@microsoft/applicationinsights-web';
+import {
+  DEVX_API_URL,
+  GRAPH_API_SANDBOX_URL,
+  GRAPH_URL,
```
MIchaelMainer, line 5: We also make calls to mgt.dev to the toolkit component. Should we add that here? Do we need to consider sanitization of those requests and responses since they appear to contain user information?

src/telemetry/telemetry.ts
```diff
@@ -19,40 +31,52 @@ class Telemetry implements ITelemetry {
       instrumentationKey: this.getInstrumentationKey(),
       disableExceptionTracking: true,
       disableAjaxTracking: true,
+      disableFetchTracking: false,
```
MIchaelMainer, line 34: Does this mean we track Fetch information for any call or just calls to Microsoft Graph and the DevX API? If the outer application makes a Fetch call, are we capturing that? What about Fetch to the identity endpoint?

src/telemetry/filters.ts
millicentachieng: We are currently only interested in DevX and Graph APIs. Sanitization is being done for requests to Graph API. There's nothing to sanitize for DevX API calls.

src/telemetry/telemetry.ts
millicentachieng: Enabling fetch tracking logs all fetch calls. We then drop all but calls to Graph and DevX APIs.

src/app/utils/query-url-sanitization.ts
```diff
@@ -33,25 +36,48 @@ export function isFunctionCall(segment: string): boolean {
 }
 
 /**
- * @param url - query url to be sanitized e.g. https://graph.microsoft.com/v1.0/users/{user-id}
+ * Sanitize Graph API Sandbox URL used when a user is not signed in
+ * @param url - URL to be sanitized
+ */
+export function sanitizeGraphAPISandboxUrl(url: string): string {
```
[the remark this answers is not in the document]
millicentachieng: We are sanitizing data for calls to the Sandbox to be safe.

src/telemetry/filters.ts
millicentachieng: I've created an [issue](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/905) for capturing version information

src/telemetry/telemetry.ts
MIchaelMainer: It would be helpful to have this commented in the code so that future devs understand that a change to the filtering could result in inadvertent capture of PII.

src/telemetry/filters.ts
MIchaelMainer: Got it, I now see that DevX API calls for snippets and autocomplete send info in the body and that the permissions API calls have a sanitized URL. 

`https://graphexplorerapi.azurewebsites.net/permissions?requesturl=/users/{users-id}/messages&method=GET`

src/telemetry/filters.ts
millicentachieng: I'm thinking renaming it to `sanitizeTelemetryItemUriProperty` might better capture what we are doing in the method. It's supposed to only work for telemetry items with the `uri` property.

jobala (approved):

thewahome (approved):

thewahome:
@millicentachieng If you'd like to test the telemetry on the staging site, you can:
-  go to the [workflow file
](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/blob/697c2a41673430c91a19d32cdc3bef69c33f9825/.github/workflows/azure-static-web-apps-jolly-sand-0ac78c710.yml#L21) in the Build And Deploy stage, add.
```
- name: Build And Deploy
        env:
          REACT_APP_INSTRUMENTATION_KEY: ${{ secrets.REACT_APP_INSTRUMENTATION_KEY }}
```

- in the [secrets page](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/settings/secrets/actions) you can add a new secret: **REACT_APP_INSTRUMENTATION_KEY**

Once the staging environment is created, you will be able to test your changes

PR #883, opened 2021-03-05
Feature: msal upgrade

### LINKED ISSUE #688: Investigate and implement use of MSAL.js 2.0 which is now available
The authentication library we use on Graph Explorer just released version 2.0 that has improvements. For this reason there is need to investigate how we can implement the new version and replace the old one and take up the process of implementing this. The materials are available here: 
1. https://developer.microsoft.com/en-us/microsoft-365/blogs/msal-js-2-0-supports-authorization-code-flow-is-now-generally-available/
2. https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-browser 
[AB#6235](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6235)

thewahome:
## Overview

Upgrade to msal 2.0
Closes #688 

### Notes

LogutPopUp [is still in design](https://github.com/AzureAD/microsoft-authentication-library-for-js/discussions/3123). the default logout implementation opens a redirect flow to log someone out. 

We now use a work-around to initiate a popup flow when logging out of the try-it experience. And clear the cache so that the account is logged out of the browser

#### Local
* You will need a registered client ID. Ensure that you have set the platform in your client ID to SPA and moved your redirect URLs on the azure portal [app registrations]

#### Online
* Go to [https://jolly-sand-0ac78c710-883.centralus.azurestaticapps.net](https://jolly-sand-0ac78c710-883.centralus.azurestaticapps.net)

#### Steps
1. logging in and out of:
-- MSA account
-- MS account
-- Demo tenant account

2. Permissions:
-- Consent to 1 scope related to a specific url (MSA/MS/Demo)
-- Consent to many scopes through the panel (MSA/MS/Demo)

3. Send authenticated requests to the graph


src/app/services/graph-client/auth/authentication-module.ts
```diff
@@ -0,0 +1,102 @@
...
+      const allAccounts = this.msalApplication.getAllAccounts();
+      if (allAccounts && allAccounts.length > 0) {
+        return allAccounts[0];
+      }
+    }
+    return undefined;
+  }
+
+  public async getToken() {
+    const silentRequest = {
```
ddyett, line 31:
```suggestion
    const silentRequest: SilentRequest = {
```

src/app/services/graph-client/auth/authentication-module.ts
```diff
@@ -0,0 +1,102 @@
...
+      scopes: defaultScopes,
+      authority: this.getAuthority(),
+      account: this.getAccount()
+    };
+    const authResponse = await this.msalApplication.acquireTokenSilent(silentRequest);
```
ddyett, line 36:
```suggestion
    return await this.msalApplication.acquireTokenSilent(silentRequest);
```

src/app/services/graph-client/auth/authentication-module.ts
```diff
@@ -0,0 +1,102 @@
...
+  public async getAuthResult(scopes: string[] = [], sessionId?: string): Promise<AuthenticationResult> {
+    const userScopes = (scopes.length > 0) ? scopes : defaultScopes;
+    const silentRequest: SilentRequest = {
+      scopes: userScopes,
+      authority: this.getAuthority(),
+      account: this.getAccount()
+    };
+
+    try {
+      const authResponse: AuthenticationResult = await this.msalApplication.acquireTokenSilent(silentRequest);
```
ddyett, line 49:
```suggestion
      return await this.msalApplication.acquireTokenSilent(silentRequest);
```

src/app/services/graph-client/auth/custom-authentication-provider.ts
```diff
@@ -0,0 +1,26 @@
+import { PublicClientApplication } from '@azure/msal-browser';
+import { AuthenticationProvider } from '@microsoft/microsoft-graph-client';
+
+import { AuthenticationModule } from './authentication-module';
+
+export class CustomAuthenticationProvider implements AuthenticationProvider {
+
+  private msalApplication: PublicClientApplication;
```
ddyett, line 8:
```suggestion
  private readonly msalApplication: PublicClientApplication;
```

src/app/services/graph-client/msal-service.ts
```diff
@@ -1,101 +1,31 @@
...
-  const urlParams = new URLSearchParams(location.search);
-  let tenant = urlParams.get('tenant');
-
-  if (tenant === null) {
-    tenant = 'common';
-  }
-
-  return `https://login.microsoftonline.com/${tenant}/`;
+  const authModule = new AuthenticationModule(msalApplication);
+  const authResult = await authModule.getToken();
```
ddyett, line 15:
```suggestion
  return await authModule.getToken();
```

src/app/services/graph-client/msal-service.ts
```diff
@@ -1,101 +1,31 @@
-import { AuthenticationParameters } from 'msal';
```
ddyett, line 1: how much work is it to completely get rid of this file and use AuthenticationModule directly.  The functions here are pretty thin so I don't see a ton of value other than less change.

src/app/services/graph-client/auth/authentication-module.ts
```diff
@@ -0,0 +1,102 @@
...
+  AuthenticationResult, InteractionRequiredAuthError,
+  PopupRequest, PublicClientApplication, SilentRequest
+} from '@azure/msal-browser';
+
+import { geLocale } from '../../../../appLocale';
+import { AUTH_URL, DEFAULT_USER_SCOPES } from '../../graph-constants';
+
+const defaultScopes = DEFAULT_USER_SCOPES.split(' ');
+
+export class AuthenticationModule {
```
ddyett, line 12: how did you test that this is now doing the right thing as far as PKCE.

nikithauc:
Please add doc for all the functions and comments for improving readability. https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html.

src/app/services/graph-client/auth/authentication-module.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  constructor(msalApplication: PublicClientApplication) {
+    this.msalApplication = msalApplication;
+  }
+
+  public getAccount(): AccountInfo | undefined {
+    if (this.msalApplication) {
+      const allAccounts = this.msalApplication.getAllAccounts();
+      if (allAccounts && allAccounts.length > 0) {
+        return allAccounts[0];
```
nikithauc, line 24: Is it guaranteed that the first account that is the account[0] will always be the valid or intended account?

src/app/services/graph-client/auth/login-type.ts
```diff
@@ -0,0 +1,16 @@
+import { LoginType } from '../../../../types/enums';
```
nikithauc, line 1: What do you think about adding util suffix to the class names utility classes. Example this can be renamed as `loginUtils.ts`

src/app/services/graph-client/auth/authentication-module.ts
```diff
@@ -0,0 +1,102 @@
...
+      }
+    }
+  }
+
+  private getAuthority(): string {
+    // support for tenanted endpoint
+    const urlParams = new URLSearchParams(location.search);
+    let tenant = urlParams.get('tenant');
+
+    if (tenant === null) {
```
nikithauc, line 65: Should it be if(!tenant) or just null condition ?

nikithauc:
Good work on adding a class structure and organizing the code :)

src/app/services/graph-client/auth/authentication-module.ts
thewahome: should be if(!tenant) nice catch

src/app/services/graph-client/auth/authentication-module.ts
thewahome: Thanks for pointing this out. I have seen that the MSAL.js team recommends that we track this user ourselves

nikithauc: > Is it guaranteed that the first account that is the account[0] will always be the valid or intended account?

Any idea about this? My concern is about picking up the wrong account. It would be a major issue if that happens.

src/app/services/graph-client/msal-service.ts
thewahome: I have moved the functions there to an authentication wrapper that would handle our interactions with authentication from the application

src/app/services/graph-client/auth/authentication-module.ts
thewahome: Yeah... 
I can use either:

- msalApplication.getAccountByHomeId
- msalApplication.getAccountByLocalId
- msalApplication.getAccountByUsername


This means I'd have to create a way to track either of the identifiers

src/modules/authentication/AuthenticationWrapper.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  private getAccount(): AccountInfo | undefined {
+    if (msalApplication) {
+      const allAccounts = msalApplication.getAllAccounts();
+      if (allAccounts && allAccounts.length > 0) {
+        if (allAccounts.length > 1) {
+          const homeAccountId = this.getHomeAccountId();
+          const account = (homeAccountId) ? msalApplication.getAccountByHomeId(homeAccountId) : allAccounts[0];
```
nikithauc, line 70: If the home account id is undefined,  I feel it is better to throw an error or ask the user to relogin. Returning `allAccounts[0]` feels more like an assumption that we are making.

src/modules/authentication/AuthenticationWrapper.ts
```diff
@@ -0,0 +1,154 @@
...
+
+export class AuthenticationWrapper implements IAuthenticationWrapper {
+
+  private static instance: AuthenticationWrapper;
+
+  public static getInstance(): AuthenticationWrapper {
+    if (!AuthenticationWrapper.instance) {
+      AuthenticationWrapper.instance = new AuthenticationWrapper()
+    }
+    return AuthenticationWrapper.instance
```
nikithauc, line 24:
```suggestion
    return AuthenticationWrapper.instance;
```

src/app/services/graph-client/auth/authentication-module.ts
thewahome: @ddyett 
Frankly, I have trusted that the msal team has done the right thing. My tests were geared towards whether GE was able to sign in and out effectively into all the accounts. I may need guidance on ways to test this the PKCE way.

MIchaelMainer: @thewahome Turn on Fiddler, you'll see a call to /authorize with &code_challenge parameter. That indicates that it is using PKCE.  

https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code

You'll also see the refresh token supplied in the response to /token.

Graph Explorer and MSAL are using the auth code flow as expected. The one thing that we likely aren't doing but should do is set the state token in the initial request to /authorize and then checking on the response from /token that we are getting the same state token. This helps prevent CSRF

src/app/services/graph-constants.ts
```diff
@@ -2,8 +2,7 @@ export const GRAPH_URL = 'https://graph.microsoft.com';
 export const USER_INFO_URL = `${GRAPH_URL}/v1.0/me`;
 export const USER_PICTURE_URL = `${GRAPH_URL}/beta/me/photo/$value`;
 export const AUTH_URL = 'https://login.microsoftonline.com';
-export const ADMIN_AUTH_URL = 'https://signIn.microsoftonline.com/common/adminconsent';
-export const TOKEN_URL = 'https://signIn.microsoftonline.com/common/oauth2/v2.0/token';
 export const DEFAULT_USER_SCOPES = 'openid profile User.Read';
 export const DEVX_API_URL = 'https://graphexplorerapi.azurewebsites.net';
 export const GRAPH_API_SANDBOX_URL = 'https://proxy.apisandbox.msdn.microsoft.com';
+export const HOME_ACCOUNT_KEY = 'fbf1ecbe-27ab-42d7-96d4-3e6b03682ee4';
```
ElinorW, line 8: Just wondering if this can be stored in the .env file instead

ElinorW (approved):

src/modules/authentication/AuthenticationWrapper.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    const allAccounts = msalApplication.getAllAccounts();
+    if (!allAccounts || allAccounts.length === 0) {
+      return undefined;
+    }
+
+    if (allAccounts.length > 1) {
+      const homeAccountId = this.getHomeAccountId();
+      return (homeAccountId) ? msalApplication.getAccountByHomeId(homeAccountId) || undefined : undefined;
```
nikithauc, line 85: Why use msalApplication.getAccountByHomeId(homeAccountId) || undefined?

src/modules/authentication/AuthenticationWrapper.ts
```diff
@@ -0,0 +1,193 @@
...
+import IAuthenticationWrapper from './IAuthenticationWrapper';
+import { msalApplication } from './msal-app';
+
+const defaultScopes = DEFAULT_USER_SCOPES.split(' ');
+const homeAccountKey = HOME_ACCOUNT_KEY;
+
+export class AuthenticationWrapper implements IAuthenticationWrapper {
+
+  private static instance: AuthenticationWrapper;
+  private isConsentFlow: boolean = false;
```
nikithauc, line 19: Please add a summary explaining the role of the properties and functions.

src/modules/authentication/AuthenticationWrapper.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  public static getInstance(): AuthenticationWrapper {
+    if (!AuthenticationWrapper.instance) {
+      AuthenticationWrapper.instance = new AuthenticationWrapper()
+    }
+    return AuthenticationWrapper.instance;
+  }
+
+  public getSessionId() {
```
nikithauc, line 28: Can you please specify a return type here?

src/modules/authentication/AuthenticationWrapper.ts
```diff
@@ -0,0 +1,193 @@
...
+} from '@azure/msal-browser';
+
+import { AUTH_URL, DEFAULT_USER_SCOPES, HOME_ACCOUNT_KEY } from '../../app/services/graph-constants';
+import { geLocale } from '../../appLocale';
+import { getCurrentUri } from './authUtils';
+import IAuthenticationWrapper from './IAuthenticationWrapper';
+import { msalApplication } from './msal-app';
+
+const defaultScopes = DEFAULT_USER_SCOPES.split(' ');
+const homeAccountKey = HOME_ACCOUNT_KEY;
```
nikithauc, line 14: Why not use the constant directly?

thewahome: I had placed the value there and replaced it with the constant, Nice catch! Changing it

src/modules/authentication/AuthenticationWrapper.ts
thewahome: The function `msalApplication.getAccountByHomeId(homeAccountId)` returns `AccountInfo | null`, however the references of the `getAccount()` function expects `AccountInfo | undefined`. So if the msalApp returns `null` I default it to `undefined` so that nothing breaks

src/modules/authentication/AuthenticationWrapper.ts
thewahome: @nikithauc I tried to limit using comments by instead using variable and function names that could be self-documenting; Unless when documenting a major decision eg the `clearCache()` function. Does this mean the properties and functions do not convey the meaning?

nikithauc: Yes, I didn't understand what `isConsentFlow` means?

nikithauc (approved):

PR #949, opened 2021-05-13
Set GET as default HTTP method if method specified in query params is invalid or missing 

millicentachieng:
Problem: Graph explorer crashes if HTTP method is missing in query params

Fix: Default HTTP method to 'GET' if missing or invalid

Example link: 
https://developer.microsoft.com/en-us/graph/graph-explorer?request=me&version=v1.0

### DIFF
--- a/src/app/views/App.tsx
+++ b/src/app/views/App.tsx
@@ -137,7 +137,7 @@ class App extends Component<IAppProps, IAppState> {
     const urlParams = new URLSearchParams(window.location.search);
 
     const request = urlParams.get('request');
-    const method = urlParams.get('method');
+    const method = this.validateHttpMethod(urlParams.get('method') || '');
     const version = urlParams.get('version');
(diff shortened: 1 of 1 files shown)


src/app/views/App.tsx
```diff
@@ -155,13 +155,22 @@ class App extends Component<IAppProps, IAppState> {
 
     return {
       sampleUrl: `${graphUrl}/${version}/${request}`,
-      selectedVerb: method,
+      selectedVerb: this.validateHttpMethod(method),
```
thewahome, line 158: This function can be best called from the function `getQueryStringParams()` where its usage is more uniform

thewahome (changes requested):

src/app/views/App.tsx
millicentachieng: That was the plan. Thanks for mentioning it.

thewahome (approved):

PR #952, opened 2021-05-14
Merge v4.6.1

millicentachieng:
Merge hotfix #949 into dev

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "graph-explorer-v2",
-  "version": "4.4.2",
+  "version": "4.6.1",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


thewahome (approved):

jobala (changes requested):
This defaults to 'GET' for something like 'POS' when the user wanted to make a POST request. Why are we defaulting to GET instead of letting the user know that the method is invalid?

thewahome:
This is the request that is placed in the docs that redirects to GE when clicked. It makes sense to default to GET when no method is passed. There is no opportunity for us to give feedback to the writer of the docs team at this point

jobala (approved):

PR #955, opened 2021-05-19
Feature: sovereign clouds

thewahome:
## Overview

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description


src/modules/sovereign-clouds/index.ts
```diff
@@ -0,0 +1,52 @@
+import { AUTH_URL, GRAPH_URL } from '../../app/services/graph-constants';
+import { geLocale } from '../../appLocale';
+import { ICloud } from '../../types/cloud';
+
+const storageKey = 'cloud';
+
+export const clouds: ICloud[] = [
```
thewahome, line 7: Add comment that tells people that this is where new clouds are added

src/modules/sovereign-clouds/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  return clouds.find(k => k.locale === localeValue);
+}
+
+export function getCloudProperties(cloudName: string): ICloud | undefined {
+  return clouds.find(k => k.name === cloudName) || undefined;
+}
+
+export function replaceBaseUrl(url: string): string {
+  const { origin } = new URL(url);
+  const currentCloud = getCurrentCloud() || null;
```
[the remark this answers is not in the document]
thewahome: I used `|| null` to help to default the undefined response to a null so that I can perform a null check.
I noticed that trying `if (currentCloud) { ... }` would not work when I get undefined.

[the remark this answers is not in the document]
thewahome: Just changed this to return null instead

jobala (dismissed):

PsychoData:
Hi there, I noticed that this mentions Canary and China, as well as Global - any idea if this would work with other National Clouds like US Government, US-DOD, Germany, etc - I believe this is a list - [Microsoft Graph and Graph Explorer service root endpoints](https://docs.microsoft.com/en-us/graph/deployments#microsoft-graph-and-graph-explorer-service-root-endpoints)

MIchaelMainer:
Did canary get removed as an option? Asking for a friend.

thewahome:
> Did canary get removed as an option? Asking for a friend.

@darrelmiller

src/modules/authentication/AuthenticationWrapper.ts
```diff
@@ -50,7 +50,10 @@ export class AuthenticationWrapper implements IAuthenticationWrapper {
 
   public logOut() {
     this.deleteHomeAccountId();
-    msalApplication.logoutRedirect();
+    msalApplication.logout({
+      authority: this.getAuthority()
+    });
+    // msalApplication.logoutRedirect();
```
Onokaev, line 56: is this intentional? :)

src/modules/authentication/AuthenticationWrapper.ts
```diff
@@ -50,7 +50,10 @@ export class AuthenticationWrapper implements IAuthenticationWrapper {
 
   public logOut() {
     this.deleteHomeAccountId();
-    msalApplication.logoutRedirect();
+    msalApplication.logout({
```
Onokaev, line 53: The logout function is deprecated in msal. We can use the logoutRedirect function in place of this

Onokaev (changes requested):

src/modules/sovereign-clouds/cloud-options.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    if (!this.canAccessChinaCloud()) {
+      options = options.filter(option => option.key !== 'China');
+    }
+    return options;
+  }
+
+  private canAccessChinaCloud() {
+    return geLocale === 'en-US';
```
darrelmiller, line 37: What is this for?

thewahome: I was using this when sovereign clouds in different countries was something to support. 

I would use  the locale to match the clouds I had. So that if a cloud for the region exists I would suggest to the user that they can opt in to using another cloud

thewahome: This specific line was tied specifically to the China cloud.. I used en_Us for demos

darrelmiller: I don't believe we can use locale to determine anything reliable regarding what clouds a user wants to be able to use.

PR #958, opened 2021-05-19
Task/telemetry capture English theme names only

millicentachieng:
## Overview

We have been collecting telemetry for the themes that our users are opting to use. I have noticed that some them names are captured in other languages depending on the users locale.

### Fix
I am using the `selectedTheme.key` instead which is lower case and hyphenated. I'm replacing the hyphen with a space and capitalizing the first letter. e.g. `high-contrast` => `High contrast`

### DIFF
--- a/src/app/utils/string-operations.ts
+++ b/src/app/utils/string-operations.ts
@@ -1,0 +1,11 @@
+declare global {
+  interface String {
+    toSentenceCase(): String;
+  }
+}
+
+String.prototype.toSentenceCase = function () {
+  return `${this.charAt(0).toUpperCase()}${this.toLowerCase().slice(1)}`;
+};
+
+export {};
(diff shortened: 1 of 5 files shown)


src/app/utils/string-operations.ts
```diff
@@ -0,0 +1,11 @@
+declare global {
+  interface String {
+    toSentenceCase(): String;
+  }
+}
+
+String.prototype.toSentenceCase = function () {
```
thewahome, line 7: This is nice! I didn't know this can be done

thewahome (approved):

src/app/utils/string-operations.ts
thewahome: Does it get pushed to intellisense?

millicentachieng: It does.

PR #992, opened 2021-06-15
Task: change sandbox url

thewahome:
## Overview

To get the proxy url, we first make a call to the endpoint.
Store this in state 
And use it to make subsequent calls

We fall back to the old proxy when this first call does not work

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description


package.json
```diff
@@ -37,7 +37,7 @@
     "guid-typescript": "1.0.9",
     "html-webpack-plugin": "4.0.0-alpha.2",
     "identity-obj-proxy": "3.0.0",
-    "isomorphic-fetch": "2.2.1",
+    "isomorphic-fetch": "^2.2.1",
```
ElinorW, line 40: Do we need the caret? or we can just stick to the 2.2.1 version?

ElinorW (approved):

PR #1003, opened 2021-06-23
Capture telemetry for response headers copy action

millicentachieng:


thewahome (changes requested):
We can bring the telemetry closer to where the genericCopy is so that the telemetry is local to the function.

This can happen if the genericCopy function can be extended so that it can track the activity and perform the copy action

millicentachieng:
> We can bring the telemetry closer to where the genericCopy is so that the telemetry is local to the function.
> 
> This can happen if the genericCopy function can be extended so that it can track the activity and perform the copy action

Changes made

thewahome (approved):

PR #1006, opened 2021-06-30
Highlights the previously selected query

gracet9182:
## Overview

* Couldn't previously see which query was selected
* Now highlights the previously selected query by allowing the color to be the same as when you hover over it

### Demo

https://user-images.githubusercontent.com/47228306/128296476-3eb5346b-0991-4095-8209-616c9166df81.mp4

## Testing Instructions

* Checkout this branch
* Run Graph Explorer locally
* Select any query


src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -31,10 +31,15 @@ import { sidebarStyles } from '../Sidebar.styles';
 import { isJsonString } from './sample-query-utils';
 
 export class SampleQueries extends Component<ISampleQueriesProps, any> {
+  resetCollapse: boolean;
+  groups: any[];
   constructor(props: ISampleQueriesProps) {
     super(props);
+    this.resetCollapse = true;
+    this.groups = [];
```
thewahome, line 39: Looks like `groups` is a class property and also a state property. Is there a use for the first one?

thewahome: On taking a second look, the two variables `groups` and `resetCollapse` are not necessary, since the `selectedQuery` state variable is present

src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -48,7 +53,9 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
   };
 
   public componentDidUpdate = (prevProps: ISampleQueriesProps) => {
+    const { sampleQueries } = this.state;
     if (prevProps.samples.queries !== this.props.samples.queries) {
+      this.groups = generateGroupsFromList(sampleQueries, 'category');
```
thewahome, line 58: Calculating it close to its point of need prevents having to track its location and state elsewhere in the page

src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -215,10 +226,13 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
         <div className={classes.groupHeader}>
           <DetailsRow
             {...props}
+            styles={customStyles}
             onClick={() => {
               if (!selectionDisabled) {
                 this.querySelected(props.item);
               }
+              this.resetCollapse = false;
+              this.setState({ selectedQuery: props.item.id })
```
thewahome, line 235: Passing the whole `props.item` into the `selectedQuery` instead of just the id helps get the category of the query and use it to set the expanded state of the category in the groups function

```suggestion
              this.setState({ selectedQuery: props.item})
```

src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -329,7 +343,9 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
 
     const { sampleQueries } = this.state;
     const classes = classNames(this.props);
-    const groups = generateGroupsFromList(sampleQueries, 'category');
+    if (this.resetCollapse) {
+      this.groups = generateGroupsFromList(sampleQueries, 'category');
+    }
```
thewahome, line 348: Since the category is already known in the `selectedQuery` state property, it will be pretty easy to get the category and make it expanded

```suggestion
    const groups = generateGroupsFromList(sampleQueries, 'category');
    if (this.state.selectedQuery) {
      const index = groups.findIndex(k => k.key === this.state.selectedQuery.category);
      groups[index].isCollapsed = false;
    }
```

src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -435,7 +451,7 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
             items={sampleQueries}
             selectionMode={SelectionMode.none}
             columns={columns}
-            groups={groups}
+            groups={this.groups}
```
thewahome, line 454:
```suggestion
            groups={groups}
```

thewahome (changes requested):

src/app/views/sidebar/sample-queries/SampleQueries.tsx
gracet9182: 8c67a81bfe47a5011eb99bb9edb1028dd5591c7f

src/app/views/sidebar/sample-queries/SampleQueries.tsx
gracet9182: I removed it with 8c67a81bfe47a5011eb99bb9edb1028dd5591c7f since your suggestion had it closer together to where it was needed

thewahome (approved):

PR #1007, opened 2021-07-01
Filter for DelegatedPersonal and DelegatedWork permissions

lhao03:
## Overview

I noticed when using Graph Explorer as a personal account (MSA) I still get shown DelegatedWork permissions. When I try to accept them I can't. 

delegatedPerms

### Demo
For instance, 
image
a personal account shouldn't see any permission types. 

What this bug fix shows:
image

What is currently shown:
image

## Testing Instructions

No new tests were added because the function `getProfileInfo` is already being tested and it calls the new function I added.


lhao03:
Didn't realise we changed the CICD on our end. Closing PR to clean that up.

PR #1023, opened 2021-07-14
Feature/DetailsList scrollbars only appear onHover

acchiang:
## Overview
The existing DetailsList scrollbars are the default Edge light mode scrollbars that appear when there is overflow (see right of screenshot below), and they are jarring to the user especially if they are on dark mode. 
image

https://dev.azure.com/microsoftgarage/Intern%20GitHub/_sprints/taskboard/GI21%20-%20Graph%20Explorer/Intern%20GitHub/Y21-S/06?workitem=38891

### Demo
The changes in this PR will hide the scrollbars until the user hovers over a DetailsList with overflow, then the scrollbars will appear. 
https://gyazo.com/ceaaa8648891cd6e7b672bac68b775d8

### Notes
Four scrollbars have been changed: 
History list, Sample Queries list, Permissions (panel tab) list, and Request Headers (panel tab) list. 

## Testing Instructions
* Front-end, test the four lists above (and shown in the demo recording). 
* Cross browser testing: developed in Edge and Chrome

### DIFF
--- a/src/app/views/query-runner/request/headers/RequestHeaders.tsx
+++ b/src/app/views/query-runner/request/headers/RequestHeaders.tsx
@@ -21,6 +21,7 @@ const RequestHeaders = (props: any) => {
   const [headerName, setHeaderName] = useState('');
   const [headerValue, setHeaderValue] = useState('');
   const [announcedMessage, setAnnouncedMessage] = useState('');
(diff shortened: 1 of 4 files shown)


acchiang:
I also found a custom useHover hook I could build instead of using useState in multiple places, let me know if I should update with this custom hook instead: https://usehooks.com/useHover/

thewahome (approved):

PR #1024, opened 2021-07-15
Add chat-id & membership-id

AyuAi:
The tokens added in this PR are used in our soon-to-be proposed sample queries for Teams chats. This PR will need to be merged before we can merge any sample queries into the https://github.com/microsoftgraph/microsoft-graph-devx-content repository; Otherwise, the sample queries will not work because they cannot populate the required tokens in sandbox mode.

The `chat-id` used is the first chat returned when using the `GET https://graph.microsoft.com/beta/me/chats/` query in sandbox mode, and the `membership-id` used is the first chat member id returned when using the `GET https://graph.microsoft.com/beta/me/chats/{chat-id}/members/` query in sandbox mode.

All of the newly committed IDs have been tested on `localhost`.

### DIFF
--- a/src/app/views/sidebar/sample-queries/tokens.ts
+++ b/src/app/views/sidebar/sample-queries/tokens.ts
@@ -83,6 +83,10 @@ export function getTokens(user?: any) {
       demoTenantValue: '19:d0bba23c2fc8413991125a43a54cc30e@thread.skype',
     },
     {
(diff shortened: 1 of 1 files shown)


thewahome (approved):
LGTM

millicentachieng (approved):

PR #1032, opened 2021-07-19
CORS workaround to enable downloading of content

millicentachieng:
## Overview

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/src/app/services/actions/query-action-creator-util.ts
+++ b/src/app/services/actions/query-action-creator-util.ts
@@ -1,6 +1,6 @@
 import {
   AuthenticationHandlerOptions,
-  ResponseType
+  ResponseType,
 } from '@microsoft/microsoft-graph-client';
 import { MSALAuthenticationProviderOptions } from '@microsoft/microsoft-graph-client/lib/src/MSALAuthenticationProviderOptions';
 
@@ -40,7 +40,11 @@ export async function anonymousRequest(dispatch: Function, query: IQuery) {
     ...sampleHeaders,
   };
 
-  const options: IRequestOptions = { method: query.selectedVerb, headers };
+  const options: IRequestOptions = {
+    method: query.selectedVerb,
+    headers,
+    redirect: 'manual',
+  };
 
   dispatch(queryRunningStatus(true));
 
@@ -124,6 +128,7 @@ const makeRequest = (httpVerb: string, scopes: string[]): Function => {
       .api(encodeHashCharacters(query))
       .middlewareOptions([middlewareOptions])
       .headers(sampleHeaders)
+      .option('redirect', 'manual')
       .responseType(ResponseType.RAW);
 
     let response;
--- a/src/types/request.ts
+++ b/src/types/request.ts
@@ -3,44 +3,45 @@ import { Mode } from './enums';
 import { Header, IQuery } from './query-runner';
 
 export interface IHeadersListControl {
-    handleOnHeaderDelete: Function;
-    headers?: Header[];
-    messages: any;
+  handleOnHeaderDelete: Function;
+  headers?: Header[];
+  messages: any;
 }
 
 export interface IRequestHeadersProps {
-    sampleQuery: IQuery;
-    height: string;
-    actions?: {
-        setSampleQuery: Function;
-    };
-    intl: {
-        message: object;
-    };
+  sampleQuery: IQuery;
+  height: string;
+  actions?: {
+    setSampleQuery: Function;
+  };
+  intl: {
+    message: object;
+  };
 }
 
 export interface IRequestComponent {
-    sampleQuery: IQuery;
-    mode: Mode;
-    handleOnEditorChange: Function;
-    dimensions: IDimensions;
-    headers?: Header[];
-    intl: {
-        message: object;
-    };
-    actions: {
-        setDimensions: Function;
-    };
+  sampleQuery: IQuery;
+  mode: Mode;
+  handleOnEditorChange: Function;
+  dimensions: IDimensions;
+  headers?: Header[];
+  intl: {
+    message: object;
+  };
+  actions: {
+    setDimensions: Function;
+  };
 }
 
 export interface IRequestOptions {
-    headers?: {};
-    method?: string;
-    body?: string | undefined;
+  headers?: {};
+  method?: string;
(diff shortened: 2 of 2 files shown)


ddyett:
anything blocking this moving forward?

kristen-msft:
> anything blocking this moving forward?

@ddyett @adhiambovivian @millicentachieng - we still have some work to do on the copyediting of the error messages and there is also a spacing issue in one of the error messages.

millicentachieng:
> anything blocking this moving forward?

@ddyett Sorry, I missed this comment but please review the PR when you get a chance.

millicentachieng:
> > anything blocking this moving forward?
> 
> @ddyett @adhiambovivian @millicentachieng - we still have some work to do on the copyediting of the error messages and there is also a spacing issue in one of the error messages.

@kristen-msft, could you please point me to these error messages? The screenshots in the description of the PR have all the expected messages.

src/app/services/actions/query-action-creator-util.ts
```diff
@@ -47,14 +52,76 @@ export function createAnonymousRequest(query: IQuery, proxyUrl: string) {
...
+    .responseType(ResponseType.RAW);
+
+  return client;
+}
+
+export function makeGraphRequest(scopes: string[]): Function {
+  return async (query: IQuery) => {
+    let response;
+
+    const client = createAuthenticatedRequest(scopes, query);
```
thewahome, line 102: The function name `createAuthenticatedRequest` does not match with the variable name `client`. It can be made clearer by possibly renaming `client` to `request`

src/app/services/actions/query-action-creators.ts
```diff
@@ -61,42 +99,142 @@ export function runQuery(query: IQuery): Function {
...
+      return dispatch(
+        queryResponse({
+          body: result,
+          headers: respHeaders,
+        })
+      );
+    } else {
+      if (
+        response.status === 0 &&
+        tokenPresent &&
```
thewahome, line 128: **question**: Does this mean this workaround is not available for anonymous users?

src/app/views/query-response/headers/ResponseHeaders.tsx
```diff
@@ -12,10 +13,19 @@ import { convertVhToPx, getResponseHeight } from '../../common/dimensions-adjust
...
 
+  const responseIsDownloadUrl = body['contentDownloadUrl'];
+  if (!headers && responseIsDownloadUrl) {
+    return (
+      <MessageBar messageBarType={MessageBarType.warning}>
+        <FormattedMessage id={`Missing response headers for query workaround`}/>
+      </MessageBar>
+    )
+  }
+
```
thewahome, line 28: This is nice! It hides the body so that the user doesn't see the internal payload

src/messages/GE.json
```diff
@@ -378,5 +378,10 @@
...
   "interaction_required_consent": "Please wait for the consent process to finish and verify that pop-ups are enabled.",
   "user_cancelled_consent": "Please wait for the consent process to finish.",
-  "access_denied_consent": "Your consent to this permission has been blocked by your tenant admin. Ask your admin to grant you access and then try again."
-}
\ No newline at end of file
+  "access_denied_consent": "Your consent to this permission has been blocked by your tenant admin. Ask your admin to grant you access and then try again.",
+  "This response contains unviewable content": "This response contains content that may not be viewable on this editor.",
+  "Click here to download": "Click here to download.",
+  "File response is available in original format only": "The file response is only available in the original format due to limitations of the Javascipt SDK.",
+  "Response is result of workaround": "A workaround has been used to fetch the download URL by fetching the download URL using query, await client.api('/drive/items/{item-id}').get().select('content.downloadUrl') and using the returned URL to download file contents.",
```
thewahome, line 385: **suggestion**: The message can be reworded so that the code snippet is left outside of the translation text and put into the code at the very end of the message

src/messages/GE.json
```diff
@@ -378,5 +378,10 @@
...
+  "Missing response headers for query workaround": "Some headers, typically returned for this request, may be missing due limitations of the Javascript SDK."
```
thewahome, line 386:
```suggestion
  "Missing response headers for query workaround": "Some headers, typically returned for this request, may be missing due to limitations of the Javascript SDK."
```

src/app/services/actions/query-action-creator-util.ts
millicentachieng: I agree

src/app/services/actions/query-action-creators.ts
millicentachieng: It's not necessary. There are no redirects happening.

src/messages/GE.json
millicentachieng: Kristen has a few more suggestions. I'll wait to merge.

millicentachieng: Kristen has a few more suggestions. I'll wait to merge.

jasonjoh:
Should we present this as a limitation of the JavaScript SDK? CORS is CORS right, no SDK can bypass it by design. I think this is more accurately described as an issue with the File API ([good summary here](https://stackoverflow.com/a/35583833/3307985)). Basically, the Files API doing a 302 redirect is fine, but the server it redirects to does NOT support CORS, so the redirect fails in JavaScript. This would fail even without the SDK.

The OneDrive team patched it up to allow simple GET requests on the `downloadUrl` (these don't trigger CORS), hence the workaround that's required.

I think your `A workaround has been used` messages should give a little more info on the problem:

- That the redirected download URL is not CORS-compliant
- A link to https://developer.mozilla.org/docs/Web/HTTP/CORS for more information on CORS

We may want to add this to https://docs.microsoft.com/graph/api/driveitem-get-content?view=graph-rest-1.0&tabs=http and just have GE link to that doc for more info.

I also think your `file response is only available...due to limitations of the JS SDK` message should instead say something like `due to limitations of the workaround, the file is only available...`.

millicentachieng:
Thanks @jasonjoh, for these insights. I have updated the contents of the messages to reflect what you've highlighted. I've also talked to Jeremy Kelly about updating the API documentation to include details on the workaround.

PR #1041, opened 2021-08-02
Fix: Accessibility bugs - Aug 

ElinorW:
## Overview
- [x]  - Add aria-labels for 'Copy' button on 'Code Snippets' and 'Request Headers' tab 
- [x] - Ensure elements that have scrollable content must be accessible by keyboard. (Sample Queries list)
- [x] - Use clearer focus indicator on "GET" option.
- [x] - Ensure keyboard focus does not move to non-interactive table headers/rows under the 'Modify permissions' tab item, and should only traverse on the interactive controls of the page
- [x] - Fix message bar getting cropped/overlapped at 200% zoom.
- [x] - Ensure screen reader announces correct count of history query items after activating the delete button under the 'Today' button
- [x] - Add screen reader instructions to navigate to the 'More Actions' button of query item in the History Tab options.
- [x] - Fix keyboard focus moving to some hidden element after activating the last "remove request header"(Delete icon) button.

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description


src/app/views/query-runner/query-runner.scss
```diff
@@ -7,3 +7,8 @@
   }
 }
 
+ #Dropdown17:focus {
```
thewahome, line 10: Does this id get generated each time as dropdown17? or would a new build make it dropdown20?

ElinorW: Great question! 
Seems like it could actually change. Guess I'll have to find another way to manipulate it on focus 🤦🏾‍♀️

ElinorW: Great question! 
Seems like it could actually change. Guess I'll have to find another way to manipulate it on focus 🤦🏾‍♀️

thewahome (approved):

PR #1043, opened 2021-08-03
Fix: change hover scroll behaviour

thewahome:
## Overview

Enhances #1023 to use CSS instead of javascript functions to show the hover

### DIFF
--- a/src/app/views/sidebar/sample-queries/SampleQueries.tsx
+++ b/src/app/views/sidebar/sample-queries/SampleQueries.tsx
@@ -423,26 +423,28 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
         <Announced
           message={`${sampleQueries.length} search results available.`}
         />
-        <DetailsList
-          className={classes.queryList}
-          cellStyleProps={{
-            cellRightPadding: 0,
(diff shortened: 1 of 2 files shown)


src/app/views/sidebar/sample-queries/SampleQueries.tsx
```diff
@@ -437,31 +423,26 @@ export class SampleQueries extends Component<ISampleQueriesProps, any> {
...
-            groupProps={{
-              showEmptyGroups: true,
-              onRenderHeader: this.renderGroupHeader,
-            }}
-            onRenderRow={this.renderRow}
-            onRenderDetailsHeader={this.renderDetailsHeader}
-            onItemInvoked={this.querySelected}
-          />
-        </div>
+        <DetailsList
```
ElinorW, line 426: We need the div with the navigation role for accessibility purposes

ElinorW (changes requested):

src/app/views/sidebar/sample-queries/SampleQueries.tsx
thewahome: aaah I see

ElinorW (approved):

PR #1044, opened 2021-08-03
Feature: Add Graph explorer documentation

### LINKED ISSUE #881: Add a link to Graph Explorer documentation on GE's UI
We now have Graph Explorer documentation on Graph docs and therefore a need to surface it on the tool so users can access it more easily. This task is a proposal to add a link on the tool, preferably as one of the options under settings gear, but we can move this later when we add a banner on the tool at the very bottom with the help of a designer.

ElinorW:
## Overview
Add a link to Graph Explorer documentation in the 'More Actions' menu.
Fixes #881 

### Demo
image

### DIFF
--- a/src/app/views/query-runner/request/auth/Auth.tsx
+++ b/src/app/views/query-runner/request/auth/Auth.tsx
@@ -5,7 +5,7 @@ import { FormattedMessage } from 'react-intl';
 import { useSelector } from 'react-redux';
 import { authenticationWrapper } from '../../../../../modules/authentication';
 
-import { componentNames, eventTypes, telemetry } from '../../../../../telemetry';
+import { componentNames, telemetry } from '../../../../../telemetry';
 import { IRootState } from '../../../../../types/root';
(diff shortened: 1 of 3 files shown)


ddyett:
would be great to have some designer feedback it may make more sense for the stuff that is an action with graph explorer to be on top and the stuff that links outside of graph explorer to be at the bottom.

ElinorW:
@ddyett yeah that would provide a better experience, though I am concerned about the 'Sign out' option, would this still remain as the last option, or would this now move to the middle?

adhiambovivian:
@ElinorW  @ddyett - my thinking is, that part of GE is like a "settings" section. If that's the case, "report an issue", "GE documentation" should possibly be moved to the bottom of the page. That "Settings" section should be then reserved for "settings tweaks that improve user experience eg dark mode". The sign-out should be the last option on the drop-down based on applying the logic of "the power of familiar UI design" to enable users to use the product even in auto-pilot mode.

PR #1060, opened 2021-08-17
Enhancement/error hints

Onokaev:
## Overview

Solves the following authentication errors: null_or_empty_id_token,authorization_code_missing_from_server_response, no_tokens_found, invalid_request, user_login_error,nonce_mismatch_error, login_progress_error, interaction_in_progress,
'interaction_required, invalid_grant, endpoints_resolution_error, monitor_window_timeout.

Shows the user hints in case of common user mistakes like cancelling authentication popups,  more information when scope consent is blocked and general hints on what to do when an authentication error occurs. 

### Demo
image
Notice the tip displayed after a user cancels the popup

image
Sample error when consent to access client applications with an accont is restricted by a tenant admin.

## Testing Instructions

-Try signing in then cancel the window pop-up. A hint will be displayed on a messagebar on what to do.

-Sign in with your account and attempt to consent to permissions restricted by your admin. A hint will be displayed on what you can do

- Update the redirect URI on App Registrations to 'uri + blank.html' i.e 'http://localhost:3000/blank.html'


src/app/views/authentication/Authentication.tsx
```diff
@@ -38,69 +44,112 @@ const Authentication = (props: any) => {
...
       setLoginInProgress(false);
       telemetry.trackException(
         new Error(errorTypes.OPERATIONAL_ERROR),
         SeverityLevel.Error,
         {
           ComponentName: componentNames.AUTHENTICATION_ACTION,
-          Message: `Authentication failed: ${errorCode ? errorCode.replace('_', ' ') : ''}`,
-        });
+          Message: `Authentication failed: ${
+            errorCode ? errorCode.replace('_', ' ') : ''
```
thewahome, line 67: This will replace only the first occurrence of the `_` character. Try doing it like this

```suggestion
            errorCode ? errorCode.replace(/_/g, ' '): ''
```

src/app/views/authentication/Authentication.tsx
```diff
@@ -38,69 +44,112 @@ const Authentication = (props: any) => {
...
+        deleteHomeAccountId();
+        window.sessionStorage.clear();
+      }
+    }
+    return errorCode.replace('_', ' ') + '      ' + translateMessage('Tip') + ':  ' + errorMessageHint;
+  };
+
+  const deleteHomeAccountId = () : void => {
+    localStorage.removeItem(HOME_ACCOUNT_KEY);
   }
```
thewahome, line 89: This method has already been declared in the `modules\authentication\AuthenticationWrapper.ts` file as a private method. Changing it to a public function enables access to it from outside of the class

src/app/views/authentication/Authentication.tsx
```diff
@@ -38,69 +44,112 @@ const Authentication = (props: any) => {
...
   };
 
-  const showProgressSpinner = (): React.ReactNode => {
-    return <div className={classes.spinnerContainer}>
-      <Spinner className={classes.spinner} size={SpinnerSize.medium} />
-      {!minimised && <Label>
-        <FormattedMessage id={`Signing you ${loginInProgress ? 'in' : 'out'}...`} />
-      </Label>}
-    </div>;
+  const getErrorAndHint = (errorCode: string): string => {
```
thewahome, line 74: A function should do one thing only. Consider breaking down the functionalities here so that the `getErrorAndHint()` function gets the error with the hint (if a hint exists) and clearing the cache happens in an appropriately named function that is called elsewhere

src/modules/authentication/AuthenticationWrapper.ts
```diff
@@ -113,13 +129,28 @@ export class AuthenticationWrapper implements IAuthenticationWrapper {
...
     }
   }
 
+  private checkIfAuthResultError(errorCode: string): boolean {
+    for(const errorItem of AuthErrorList){
+      if( errorItem.trim() === errorCode.trim() ){
+        return true;
+      }
+    }
+    return false;
```
thewahome, line 151: This could give the same result

```suggestion
    return AuthErrorList.includes(errorCode.trim());
```

src/modules/authentication/authUtils.ts
```diff
@@ -20,6 +20,6 @@ export function getLoginType(): LoginType {
  * -js/blob/9274fac6d100a6300eb2faa4c94aa2431b1ca4b0/lib/msal-browser/src/utils/BrowserUtils.ts#L49
 */
 export function getCurrentUri(): string {
-  const currentUrl = window.location.href.split('?')[0].split('#')[0];
-  return currentUrl.toLowerCase();
+  const currentUrl = window.location.href.split('?')[0].split('#')[0] + 'blank.html';
```
thewahome, line 23: **question:** what is the effect of adding the blank.html to the URL?

src/modules/authentication/AuthenticationWrapper.ts
```diff
@@ -155,11 +186,48 @@ export class AuthenticationWrapper implements IAuthenticationWrapper {
...
+        const { errorCode } = error;
+        if(this.checkIfAuthResultError(errorCode) && !this.consentingToNewScopes){
+        this.clearCache();
+        this.deleteHomeAccountId();
+        window.sessionStorage.clear();
+        this.eraseInteractionInProgressCookie();
+        const result = await msalApplication.loginPopup(popUpRequest);
+        this.storeHomeAccountId(result.account!);
+        return result;
+      }
```
thewahome, line 199: **nitpick:** Check the indentation in this section

**suggestion:** There is a chance for a situation where the user will be in a popup loop since the error doesn't self resolve if the app tries to re-request a login popup. It could be safer to throw the error after clearing the cache since the user will get a hint about what went wrong

src/app/views/authentication/Authentication.tsx
```diff
@@ -38,69 +44,112 @@ const Authentication = (props: any) => {
...
+    const errorMessageHint : string = translateMessage(errorCode);
+
+    for(const errorItem of AuthErrorList){
+      if(errorItem === errorCode){
+        authenticationWrapper.clearCache();
+        deleteHomeAccountId();
+        window.sessionStorage.clear();
+      }
+    }
+    return errorCode.replace('_', ' ') + '      ' + translateMessage('Tip') + ':  ' + errorMessageHint;
```
thewahome, line 84: This should return the tip message if the error message hint exists

src/app/services/actions/permissions-action-creator.ts
```diff
@@ -115,7 +115,7 @@ export function consentToScopes(scopes: string[]): Function {
       dispatch(
         setQueryResponseStatus({
           statusText: translateMessage('Scope consent failed'),
-          status: errorCode,
+          status: errorCode + ' ' + translateMessage('Tip') + ': ' + translateMessage('Access to permission denied'),
```
thewahome, line 118: **question:** Is this the only situation that consenting to permissions fails?

src/app/views/authentication/Authentication.tsx
Onokaev: Yes. Let me fix that

src/app/views/authentication/Authentication.tsx
Onokaev: Okay. Let me fix this also

src/app/views/authentication/Authentication.tsx
Onokaev: I agree. Let me break it down.

src/modules/authentication/AuthenticationWrapper.ts
Onokaev: Yeah. I've seen it. Shorter and cleaner. Let me fix this also

src/modules/authentication/authUtils.ts
Onokaev: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/errors.md#monitor_window_timeout

We will no longer be getting monitor_window_timeout errors according to the msal documentation. I could use your input on this

src/modules/authentication/AuthenticationWrapper.ts
Onokaev: That's a good idea. I missed that. I'll throw the error after clearing the cache.

src/app/views/authentication/Authentication.tsx
Onokaev: Let me add a check for that also. Nice observation

src/app/services/actions/permissions-action-creator.ts
Onokaev: They are two situations. The first one is an authentication error (access_denied) when a user is signing in to GE for the first time : If the user's tenant admin has restricted consent to client applications, they'll get the access denied error (Authentication failed: access_denied)
image

The second one is when a user is signed in to GE but they cannot consent to certain permissions because the tenant admin has blocked that. For this case, the user gets a 'Scope consent failed: access_denied
image

Is there a different scenario I have left out?

package.json
```diff
@@ -6,7 +6,7 @@
     "@azure/msal-browser": "2.12.0",
     "@babel/core": "7.12.13",
     "@babel/eslint-parser": "7.12.13",
-    "@fluentui/react": "8.28.0",
+    "@fluentui/react": "^8.28.0",
```
thewahome, line 9: Let's keep this one at a constant.. So we can remove ^

src/types/status.ts
```diff
@@ -6,4 +6,5 @@ export interface IStatus {
   status: number;
   statusText: string;
   duration?: number;
+  hint?: string;
```
thewahome, line 9: NICE!!

PR #1098, opened 2021-09-06
Capture telemetry for Graph Proxy API calls

millicentachieng:
## Overview

The current configuration only whitelists calls to the endpoint  for fetching the proxy URL. 

We need to capture calls to the proxy API .

## Testing Instructions
- Open developer tools on browser and go to network tab
- Run query as an anonymous user
- Check to see if telemetry data has been sent to Application Insights
image


thewahome (approved):
Happy that we now capture the dynamic endpoint in telemetry

PR #1105, opened 2021-09-10
Capture GE mode for all telemetry items

millicentachieng:
## Overview

Capture GE mode for all telemetry items sent to Application Insights.

This is to help us properly distinguish between Try It and main GE telemetry.

### Demo
Load Graph Explorer and click on the Run query button while on the network tab of developer tools.


thewahome (approved):
Good stuff! 

Also, since we now have access to the state in the filter, can we clean up the other telemetry instrumentations that submit state info and collect state info in the filters instead?

millicentachieng:
> Good stuff!
> 
> Also, since we now have access to the state in the filter, can we clean up the other telemetry instrumentations that submit state info and collect state info in the filters instead?

I'll explore that, @thewahome.

Issue #1141, opened 2021-09-30
Deliver 'Advanced' GE tour targeting pro/citizen GE/Graph users


adhiambovivian:

ddyett:
feature request.  placing under PM review for when/if this should be pursued.

lehrt:
Hi all, this issue is almost 4 years stale. Graph Explorer is no longer in active development and has moved to maintenance mode. No more major features like this are funded. If you are an Microsoft employee, you can reach out to internal Graph Explorer team about new tasks/feature requests. Closing this issue.

PR #1159, opened 2021-10-07
Feature: Ge tour

Onokaev:
## Overview
Gives a tour of Graph Explorer features for both getting started and advanced experiences. 

### Demo
image
image
A coachmark appears when GE is first loaded. This only appears once

image
Sample teachingbubble

image
After completing the tour, a message like this pops up with a link to the docs

image
If a user opts out of the tour mid-way, this message pops up

image
The Tour button is under the More Actions menu button

image
When the tour is launched from the 'More Actions' button, this dialog pops up.

image
Start a tour by selecting your preferred tour type

image
The tour adds an overlay over the rest of the page to maintain focus on the component being demonstrated

image
A context menu has been added on the response pivots so that you can start the tour from the response components

### Notes
#1140 #1141 
## Testing Instructions
On the more actions button, click Graph Explorer Tour


Onokaev:
> When the tour dialog pops up, it has the first section selected by default. Let's change this to have no options selected so that a user can choose for themselves which one they'd like.
> 
> We can also have a button that says "start tour" which closes the dialog and starts the tour

Sure. Let me add that

Onokaev:
> image
> 
> This is awesome!
> 
> _question_: How do users know that the items are right-clickable?

I can add a small note on the initial dialog to notify users of this

src/app/views/query-response/QueryResponse.tsx
```diff
@@ -101,16 +104,48 @@ const QueryResponse = (props: IQueryResponseProps) => {
     }
   };
 
+  const selectContextItem = (e: any, item: any, link: any) => {
```
jobala, line 107: Consider using the appropriate type instead of **any**

Onokaev: Thank you. 
Fixed!

Onokaev:
Will reopen when feature is planned for

Issue #1142, opened 2021-09-30
Updating/adding new features to GE tour. For this section, we can use content repo to track new features or feature updates that go towards GE tour


adhiambovivian:

ddyett:
closing this as not actionable and feature is not being worked on.

PR #1200, opened 2021-11-08
Task: fetch steps from blob

Onokaev:
## Overview
Fetch tour content from devx api '/tourSteps' endpoint
Added local steps as a redundancy in case calls to devX fail

#1142 

### Demo

### Notes


src/app/views/query-response/pivot-items/pivot-items.tsx
```diff
@@ -56,13 +57,14 @@ export const getPivotItems = () => {
       const { itemKey } = link;
       const itemKeyString: string = !itemKey ? '' : itemKey.toString();
       const target = findTarget(itemKeyString);
-      const targetStepIndex = getTargetStepIndex(target, item.key)
+      const targetStepIndex = getTargetStepIndex(target, item.key, tour.tourSteps, tour.beginner)
```
thewahome, line 60: This function `getTargetStepIndex` has 4 parameters. It looks like the function could be doing more than one thing. If so, it can be split into smaller functions and they can be called in succession as they resolve to create the parameters for the next function.

Another option would be to create an object and pass the object into the function if the function truly requires all the parameters.

Alternatively specifically for this case,  `tour.tourSteps` and `tour.beginner` can just be passed in as one and destructured within the function.

src/app/views/tour/utils/blob-steps/getSteps.tsx
```diff
@@ -0,0 +1,55 @@
...
+      content: tourContent,
+      disableBeacon,
+      directionalHint,
+      autoNext,
+      title,
+      hideCloseButton,
+      expectedActionType,
+      spotlightClicks,
+      query
+    }
```
thewahome, line 35: Using the spread operator, all the properties of an object can be copied and the new properties added to it. There'd then be no need to fully destructure the `step` object so that it can fit into the new object.

```suggestion
    const actualStep: ITourSteps = {
      ...step,
      content: tourContent,
    }
```
```suggestion
    const actualStep: ITourSteps = {
      target,
      content: tourContent,
      disableBeacon,
      directionalHint,
      autoNext,
      title,
      hideCloseButton,
      expectedActionType,
      spotlightClicks,
      query
    }
```

src/app/views/tour/utils/blob-steps/getSteps.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+
+    if(!!query){
+      const objectCount = Object.keys(query).length;
+      if(objectCount === 0) {
+        delete actualStep.query
+      }
+
+    }
```
thewahome, line 44: *Question*: Not really sure what this step means. Can we document it in a well named function?

src/app/views/tour/utils/blob-steps/getSteps.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    if (advanced === true) {
```
thewahome, line 46:
```suggestion
    if (advanced) {
```
```suggestion
    if (advanced === true) {
```

src/app/views/tour/utils/blob-steps/getSteps.tsx
```diff
@@ -0,0 +1,55 @@
...
+      advancedSteps.push(actualStep);
+    }
+    else {
+      beginnerSteps.push(actualStep);
+    }
+  }
+
+  return (beginnerTour === true ? beginnerSteps : advancedSteps);
```
thewahome, line 54:
```suggestion
  return (beginnerTour ? beginnerSteps : advancedSteps);
```
```suggestion
  return (beginnerTour === true ? beginnerSteps : advancedSteps);
```

src/app/views/tour/Tour.tsx
```diff
@@ -6,21 +6,23 @@ import { IRootState } from '../../../types/root';
...
 
 const GETour = () => {
 
   const { tour } = useSelector((state: IRootState) => state);
-  const {step, beginner, isRunning} = tour;
+  const { step, beginner, isRunning, tourSteps, pending } = tour;
   const dispatch = useDispatch();
   const [run, setRun] = useState(isRunning);
-  const [steps] = useState(beginner === true ? BEGINNER_TOUR : ADVANCED_TOUR);
+  const [steps, setSteps] = useState(BEGINNER_TOUR);
```
thewahome, line 25: *Question*: This `BEGINNER_TOUR` is a constant set to an empty array. Is there a reason why we shouldn't just pass an empty array instead of the constant?

src/app/views/tour/utils/contextHelpers.ts
```diff
@@ -1,9 +1,13 @@
...
+import { getTourSteps } from './blob-steps/getSteps';
 import { componentToTargetMap } from './mapToComponent';
-import { ADVANCED_TOUR, COMPONENT_INFO } from './steps';
-import { ITourContextMenu } from './types';
+import { COMPONENT_INFO } from './steps';
+import { ITourContextMenu } from './types'
+
+
+export const getTargetStepIndex = (target: string, menuType: string, tourSteps: any, beginner: boolean): number => {
+  const ADVANCED_TOUR = getTourSteps(tourSteps, beginner)
```
thewahome, line 9: *Nitpick*: The naming convention according to how the rest of the code works depicts a constant we can use in a global scope. e.g actions / URLS.. 

It could then read 
```suggestion
  const advancedTour = getTourSteps(tourSteps, beginner);
```
```suggestion
  const ADVANCED_TOUR = getTourSteps(tourSteps, beginner)
```

src/app/views/tour/utils/contextHelpers.ts
```diff
@@ -1,9 +1,13 @@
...
 
-export const getTargetStepIndex = (target: string, menuType: string): number => {
   const tourStep = menuType === 'info' ? COMPONENT_INFO : ADVANCED_TOUR;
   if (target === '') {
```
thewahome, line 12:
```suggestion
  if (!target) {
```
```suggestion
  if (target === '') {
```

src/app/views/App.tsx
```diff
@@ -80,6 +81,7 @@ class App extends Component<IAppProps, IAppState> {
   private currentTheme: ITheme = getTheme();
   private statusAreaMobileStyle = appStyles(this.currentTheme).statusAreaMobileScreen;
   private statusAreaLaptopStyle = appStyles(this.currentTheme).statusAreaLaptopScreen;
+  private displayNewFeatureCoachmark = localStorage.getItem('tour');
```
thewahome, line 84: It would make sense to abstract this implementation to a function that returns a boolean. 
```suggestion
  private displayNewFeatureCoachmark = shouldDisplayNewFeatureCoachMark();
```

Subsequently it would change the check to something truthy.

`{this.displayNewFeatureCoachmark && <NewFeature />}`

or even 
`{shouldDisplayNewFeatureCoachMark() && <NewFeature />}`

src/app/views/tour/Tour.tsx
Onokaev: The setSteps function is expecting an array of type ITourSteps[]. So even if the array is empty, it still has to be of type ITourSteps[]

Onokaev:
Will re-open when feature is planned for

PR #1201, opened 2021-11-09
fix: request and response height on try-it

Onokaev:
## Overview
Fixes request and response heights on try-it experience and nps. 

### Demo

### Notes
Assigning classname dynamically to the response pivot was breaking experiences that rely on the classname being static.
The responsiveness that relied on the dynamic classname has been moved to the scss file as a media query.

## Testing Instructions

* How to test this PR
Set graphexplorermode to Mode.TryIt: The response height is maintained above a set minimum.
Launch nps feedback through the button: The pop-up does not push the request height to the top


thewahome (approved):
LGTM

PR #1204, opened 2021-11-10
Fix: Remove hover functions on history tab

### LINKED ISSUE #1156: Improve overall user navigation on sample query history
image

When using the history tab on GE, after running one of the historical queries, I would like to be able to navigate easily to the point where I had my initial historical query for example, when I run one of the queries from the 'yesterday' section GE takes you back to 'today' section instead of the current historical section that the user is exploring and in my case that's 'yesterday' section. This forces a user to scroll through the historical tab multiple times especially if the user has several historical query groupings such as 'today',' yesterday' etc. and several historical queries.

ElinorW:
## Overview

The 'Older' and 'Yesterday' group collapse when the mouse focus is moved elsewhere after a user runs a query from either of these groups.
This fix ensures the focus remains on the query selected even if the mouse focus moves elsewhere.

Fixes #1156 

### Demo

### Notes

## Testing Instructions

* Load Graph Explorer
* Navigate to the History Tab 
* Select a query from the "Yesterday' or 'Older' group
* Move mouse focus to anywhere else on the page
* Notice the group doesn't collapse, and highlight remains on the selected query.

### DIFF
--- a/src/app/views/sidebar/history/History.tsx
+++ b/src/app/views/sidebar/history/History.tsx
@@ -527,12 +527,7 @@ export class History extends Component<IHistoryProps, any> {
             message={`${items.length} search results available.`}
           />
           {items.length > 0 &&
-            <div
-              onMouseEnter={() => this.setState({ isHoverOverHistoryList: true })}
-              onMouseLeave={() => this.setState({ isHoverOverHistoryList: false })}>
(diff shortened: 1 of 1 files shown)


Onokaev (approved):
Looks good

PR #1214, opened 2021-11-16
Enhancement/hint when unauthenticated

### LINKED ISSUE #974: Hint users to login when they are not getting results
Sometimes users will try APIs immediately and get no results.  It's not always clear that they can/need to login on the left-hand side to get "real" data.  Even though it says so on the sample account, there might be some UX optimizations that can be made to make this more intuitive.

Maybe a hint when no results are received, or when requests are made to APIs outside of the sample's well-known set of responses.  "Hey you can login for real results."  
[AB#9661](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9661)

Onokaev:
## Overview
Fixes #974 

After running more than 8 queries when unauthenticated, the coachmark below appears
image
Hovering over the coachmark displays this bubble content:
image

### Notes

## Testing Instructions
Sign out of GE.
Run 9 sample queries
A coachmark appears which prompts the user to sign in to see real data


ElinorW:
Hey , could you update the testing instructions...
I was confused at the "Click on any sample queries other than GET" step...
Then I realised you meant that we should click on the GET request more than 5 times

Onokaev:
Sure,
I’ve made the update

jobala (changes requested):
@Onokaev  from the issue description the problem is that user tries to run samples **immeadiately** without knowing that they are getting sample data, showing this to the user eight requests later doesn't solve this problem. Also this issue is the reason we had the **sign in to access your own data** label on top of the login button and should be considered solved,  why are we doing this work again?

jobala:
@Onokaev also note that this issue was resolved by [this change](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/blob/28a58fba5109b505e73ffcd2a9264a39bb3d3d18/src/app/views/authentication/Authentication.tsx#L83) but maybe we forgot to close it then. This change was added around two days after the issue was created.

jobala:
@Onokaev @adhiambovivian  It is possible that this issue was solved by [this change](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/blob/28a58fba5109b505e73ffcd2a9264a39bb3d3d18/src/app/views/authentication/Authentication.tsx#L83) because it made it clear that a users need to sign in to access their data.

Before the change
image

After the change
image

Onokaev:
> @Onokaev @adhiambovivian It is possible that this issue was solved by [this change](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/blob/28a58fba5109b505e73ffcd2a9264a39bb3d3d18/src/app/views/authentication/Authentication.tsx#L83) because it made it clear that users need to sign in to access their data.
> 
> Before the change image
> 
> After the change image

Hi @jobala. Thanks for pointing this out. Yes, this change you've highlighted does address the issue. I am closing this PR

Issue #1078, opened 2021-08-26
Status messages missing for unauthenticated users


ElinorW:
**Describe the bug**
When an unauthenticated user runs a query and it returns a successful or error response, the status messages are missing. E.g OK, Bad Request, Unauthorized... etc

**To Reproduce**
1. Go to [Graph Explorer](https://developer.microsoft.com/en-us/graph/graph-explorer)
2. Click on 'Run Query' button
3. Observe -200 message (Only status code is displayed)
4. See missing 'OK' message

**Expected behavior**
The status message should be displayed and not just the status code. 

**Screenshots**
![image](https://user-images.githubusercontent.com/18407044/130943513-55068dda-84ac-4f0f-ab4c-6b7c6b52688b.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: All

**Additional context**
This problem is not reproduced in cases where a user is authenticated.
 
[AB#10874](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/10874)

adhiambovivian:
To be fixed on API-Sandbox side

PR #1278, opened 2021-12-01
Task: add status message for anonymous queries

Onokaev:
## Overview
Solves #1078 

### Demo
On production, the status message is missing
image

This PR adds the status message to the response if it's unavailable
image

## Testing Instructions
Start Graph Explorer
Click on any sample queries when you are **not signed in.** You should see the status messages for all GET sample queries

### DIFF
--- a/src/app/utils/status-message.ts
+++ b/src/app/utils/status-message.ts
@@ -37,3 +37,52 @@ export function getMatchesAndParts(message: string) {
   const parts: string[] = message.split(numberPattern);
   return { matches, parts };
 }
+
+const responseStatusMap = new Map([
+  [100, 'Continue'],
+  [101, 'Switching Protocols'],
+  [102, 'Processing'],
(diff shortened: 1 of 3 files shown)


src/app/utils/status-message.ts
```diff
@@ -37,3 +37,52 @@ export function getMatchesAndParts(message: string) {
   const parts: string[] = message.split(numberPattern);
   return { matches, parts };
 }
+
+const responseStatusTextMap = new Map([
```
jobala, line 41: We don't need to specify the type in the variable name because the editor gives us that information. Consider renaming `responseStatusTextMap` to `repsonseStatusText`. On the same note, `httpStatusDescription` is a more appropriate variable name.

Onokaev: On it

ElinorW (approved):

PR #1307, opened 2021-12-09
Fix: Expand/Collapse button on Response Preview tab

ElinorW:


src/app/views/common/monaco/Monaco.tsx
```diff
@@ -1,14 +1,14 @@
 import { FocusZone } from '@fluentui/react';
 import React from 'react';
-import MonacoEditor, { ChangeHandler } from 'react-monaco-editor';
```
millicentachieng, line 3: Does this package`react-monaco-editor` need to be uninstalled since it's no longer being used.

ElinorW: Yeah, let me uninstall it

ElinorW:
Closing this as the package-lock got an issue.

PR #1357, opened 2022-01-19
Fix version validation for autocomplete

### LINKED ISSUE #1335: Input validation on version for the autocomplete endpoint
**Describe the bug**
Version information is parsed as the 4 characters following https://graph.microsoft.com/, so if a user types https://graph.microsoft.com//, the following request is sent to the backend:
```
GET graphexplorerapi.azurewebsites.net/openapi?url=/&style=geautocomplete&graphVersion=/
```
which is guaranteed to fail with unsupported graphVersion `/`

**Expected behavior**
Version information is validated client-side.

**Additional context**
We are notified about this by the automated AppInsight alert: https://portal.microsofticm.com/imp/v3/incidents/details/282681669/home

millicentachieng:
## Overview
The changes include:
- Removing extra slashes from URL before calling autocomplete endpoint
- Validating Graph API version before calling autocomplete endpoint
- Display warning if the version specified for the query is invalid

With these changes, we are reducing the number of calls to the autocomplete endpoint that will definitely fail.

Closes #1335


millicentachieng:
@Onokaev, @ElinorW Please review this PR when you get the chance

Onokaev (approved):
Nice work @millicentachieng!
Suggestion: What do you think about having a message under the query text input that warns a user when the version information is invalid (the same way a message is displayed when the query has invalid white spaces)?

ElinorW (approved):

ElinorW:
> Nice work @millicentachieng! Suggestion: What do you think about having a message under the query text input that warns a user when the version information is invalid (the same way a message is displayed when the query has invalid white spaces)?


Yeah, I also think it would be great to have the message as well because if it's a learning tool, we want to encourage dev's to practice writing queries correctly. 
what do you think?

millicentachieng:
I agree. We should let a user know if the version in their input is wrong.

PR #1367, opened 2022-01-25
Show Resource Explorer HTTP methods next to resource

### LINKED ISSUE #1322: [Resource Explorer] Display HTTP methods in Resource explorer like they are shown in Samples Queries.
image

This will allow "selecting" a resource by just clicking on the method.  No need to use context menu.
Also the context menu should allow "add to collection" for specific HTTP methods.

millicentachieng:
## Overview

Closes #1322

### Demo
Resource Explorer new look
image

 
Postman collection preview
image


ElinorW (commented):
Wow, I really like the added changes! 👍🏾👏🏾

A question though, why do we use the plug disconnected icon? 
image

ElinorW:
Also, previously on selecting 'Set Query' the query would autofill...
I'm not experiencing the same behaviour here...

image
_Image show's when a user selects GET, the Method, Version and Query are auto-filled at the top _

millicentachieng:
> Wow, I really like the added changes! 👍🏾👏🏾
> 
> A question though, why do we use the plug disconnected icon?
 It's to indicate that a node has no children.

millicentachieng:
> Also, previously on selecting 'Set Query' the query would autofill... I'm not experiencing the same behaviour here...
> 
> _Image show's when a user selects GET from the resource..., the Method, Version and Query are auto-filled at the top_

Clicking on the HTTP method sets the query on the run query box. I agree that it may not be that intuitive to a user and we should instead look into setting the query when the list item is clicked.

Onokaev:
> > Also, previously on selecting 'Set Query' the query would autofill... I'm not experiencing the same behaviour here...
> > _Image show's when a user selects GET from the resource..., the Method, Version and Query are auto-filled at the top_
> 
> Clicking on the HTTP method sets the query on the run query box. I agree that it may not be that intuitive to a user and we should instead look into setting the query when the list item is clicked.

Nice work @millicentachieng! 
Good idea! We can maintain the methods and allow for setting query when a list item is clicked

millicentachieng:
> Also, there is a small issue with the localization:
> 
> In PathsReview.tsx file line 84, the id passed to FormatMessage has been truncated (It's in 2 lines). Having it on one line will solve the localization error.

Good catch, @Onokaev!

ElinorW:
> > Wow, I really like the added changes! 👍🏾👏🏾
> > A question though, why do we use the plug disconnected icon?
> 
> It's to indicate that a node has no children.

Oh I see... I'm not sure it's very intuitive.. for instance I didn't get that. Can't it just stay without the icon?

millicentachieng:
> > > Wow, I really like the added changes! 👍🏾👏🏾
> > > A question though, why do we use the plug disconnected icon?
> > 
> > 
> > It's to indicate that a node has no children.
> 
> Oh I see... I'm not sure it's very intuitive.. for instance I didn't get that. Can't it just stay without the icon?

That's a good point. It might cause more confusion to our users than being informative. @thewahome can give us his reasons but I'll remove the icons for now.

ElinorW:
Also some suggestions, 
1) could there be a confirmation that an item has been added to the collection. I believe it would help users who navigate with a screen reader for example. 
2) When a user presses enter on the resource e.g. GET admin, could it also set the query the same way it does when you click on it? This would make it easier for user's who use keyboard for navigation.

millicentachieng:
> Also some suggestions,
> 
> 1. could there be a confirmation that an item has been added to the collection. I believe it would help users who navigate with a screen reader for example.
> 2. When a user presses enter on the resource e.g. GET admin, could it also set the query the same way it does when you click on it? This would make it easier for users who use a keyboard for navigation.

Good insights, @ElinorW. I have created a [new issue](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/1394) to track the first item. The second item's done.

ElinorW (approved):

Issue #1063, opened 2021-08-23
Request header ConsistencyLevel eventual is not sent correctly


Licantrop0:
Sending [Advanced queries](https://docs.microsoft.com/en-us/graph/aad-advanced-queries) with graph explorer sometimes fails because `ConsistencyLevel: eventual` header is not sent:

![image](https://user-images.githubusercontent.com/2388430/130453483-af7b201f-264a-48c6-bf4c-17edd65147e3.png)

**To Reproduce**
Try here: https://developer.microsoft.com/en-us/graph/graph-explorer?request=applications%3F%24search%3D%22displayName%3ABrowser%22&method=GET&version=v1.0&GraphUrl=https://graph.microsoft.com&headers=W3sibmFtZSI6IkNvbnNpc3RlbmN5TGV2ZWwiLCJ2YWx1ZSI6ImV2ZW50dWFsIn1d

Licantrop0:
yes, it happen when I'm logged in.

Onokaev:
This issue was consistent with the previous API sandbox. 
It is currently being planned for as a feature request

Licantrop0:
@Onokaev now it happens when I'm **NOT logged in**, see (missing @odata.count):
![MicrosoftTeams-image](https://user-images.githubusercontent.com/2388430/143865436-1dd3b0ac-959c-42e5-a833-9cae527a96be.png)

Instead, when I'm logged in I can see the count:
![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/2388430/143865495-9bab0bf6-dece-4420-953b-399375aa3b33.png)

Before July/August, Graph Explorer was working fine with the [Advanced Query Capabilities](https://docs.microsoft.com/en-us/graph/aad-advanced-queries)

PR #1391, opened 2022-01-31
Fix: Selectively add headers to fetch call

Onokaev:
## Overview
When running advanced queries in an anonymous experience, all GET responses are cached temporarily. This means that if the returned response is an error response, any subsequent API calls will hit the cache and not the API sandbox.
Unless the url is changed, any API call with the same url will hit the cache until the cache expires.

This fix forces a call to the API Sandbox in a subsequent request only if the previous request returned an error response
Solves #1063 

### Demo
image
Running an advanced query without the ConsistencyLevel header

image
Response returned after adding the correct header and making the same API call

### Notes

## Testing Instructions
Run this query: https://developer.microsoft.com/en-us/graph/graph-explorer?request=users%2F%24count&method=GET&version=v1.0&GraphUrl=https://graph.microsoft.com
It doesn't have the correct request header

Notice the error response

Now add the header: ConsistencyLevel = eventual

Run the query

Notice the successful response.

### DIFF
--- a/src/app/services/actions/query-action-creator-util.ts
+++ b/src/app/services/actions/query-action-creator-util.ts
@@ -48,6 +48,8 @@ export function createAnonymousRequest(query: IQuery, proxyUrl: string) {
 
   const authToken = '{token:https://graph.microsoft.com/}';
   const headers = {
+    'cache-control': 'no-cache',
+    pragma: 'no-cache',
     Authorization: `Bearer ${authToken}`,
     'Content-Type': 'application/json',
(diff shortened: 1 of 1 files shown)


RabebOthmani:
@darrelmiller  I'd love your opinion please. As a user, it is hard for me to guess from the error there that I need to specifically add the ConsistencyLevel header. Is there something we can do to refine the error message or even identify the call as needing ConsistencyLevel then we can provide the user with the right message?

ElinorW (approved):

PR #1410, opened 2022-02-02
Revert "Show Resource Explorer HTTP methods next to resource"

darrelmiller:
Reverts microsoftgraph/microsoft-graph-explorer-v4#1367 to get PM sign off before releasing.


RabebOthmani:
Can we please revert all PRs on all tickets I didn't sign off on. Thank you

millicentachieng:
After a discussion with @RabebOthmani, we've decided not to revert the 8 PRs related to Resource Explorer that have been merged into `dev` branch in the past month. This is because the process is going to introduce breaking changes with some of the features we want to go out building on top of some changes that were made.

The Resources tab is going to be hidden for now as it already is in production. We do realize that this should not have happened and the Resource Explorer work should have remained unmerged until there was approval from the stakeholders.

Going forward, we will be introducing a branch that will hold changes coming from multiple PRs related to a feature. This will allow for devs to work on tasks that build on top of each other without waiting for them to be merged into dev.

PR #1434, opened 2022-02-08
Task: Improve test coverage

Onokaev:
## Overview
Adds tests for all files

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description


thewahome (commented):
*question*: is there a different way to test QueryInput.tsx and Request.tsx without redoing it in the _mocks_ folder?

Onokaev:
> _question_: is there a different way to test QueryInput.tsx and Request.tsx without redoing it in the _mocks_ folder?

The mocks are actually for the QueryRunner.tsx. The two components are used in QueryRunner component and not all the props required by them are passed to them from the QueryRunner component. 
          
These are the props supplied to the Request component for example.
image

The props required by the Request component are: 
image

Jest breaks when I try to render QueryRunner without the mocks

Onokaev:
> It's possible we are passing in properties we do not need. Should we remove the extras in this PR or open a new issue that changes the component signature?

Yes. It's a good idea to change the component signature. In the meantime, I can remove the mocks and stub divs in places were they are needed

thewahome (approved):

Issue #1427, opened 2022-02-04
[Resource Explorer] Search box style 


RabebOthmani:
Currently the search box looks like it has double border , interior border should go. 
<img width="272" alt="search box" src="https://user-images.githubusercontent.com/5781590/152584979-ed47a98b-9060-4dff-96cf-392914c5794f.png">

RabebOthmani:
@Onokaev tested on sample queries using the test instance provided , it looks fine. I'm assuming it would behave in the same way with Resource Explorer.

Onokaev:
Yes @Rabeb. The searchbox is the same on the sample queries and resources as well

PR #1438, opened 2022-02-08
Fix: Remove search box outline

Onokaev:
## Overview
Solves #1427 

### Demo
image

image

image

### Notes


src/app/views/sidebar/history/History.tsx
```diff
@@ -101,9 +102,13 @@ export class History extends Component<IHistoryProps, any> {
...
+  }
+
+  private sortItems = ( items: any ) => {
     items
       .sort(dynamicSort('createdAt', SortOrder.DESC))
-      .forEach((value, index) => {
+      .forEach((value : any, index: any) => {
         value.index = index;
       });
     return items;
```
thewahome, line 114: Let's use the types here instead of any especially since we know what we are sorting

Onokaev: Making the changes

src/app/views/app-sections/ResponseMessages.tsx
```diff
@@ -49,7 +49,7 @@ export function responseMessages(graphResponse: IGraphResponse, sampleQuery: IQu
   if (odataLink) {
     return (
       <MessageBar messageBarType={MessageBarType.info}>
-        <FormattedMessage id={'This response contains an @odata property'} />: @odata.{odataLink!.name}
+        <FormattedMessage id={'This response contains an @odata property.'} />: @odata.{odataLink!.name}
```
ElinorW, line 52: Why was the full stop added? I don't think we should have it before a colon

thewahome: Looks like it is how it went into the all of the localization files

ElinorW (approved):

Issue #1454, opened 2022-02-11
When doing "Enter" in the query textfield, we get a whitespace error


sebastienlevert:
**Describe the bug**

**To Reproduce**
1. Go to https://aka.ms/gr
2. Click on the query field
3. Use a "long" query that would span on more than one line like : https://graph.microsoft.com/v1.0/users/02b1868e-37e7-4c0e-a956-846dadaab298/photo/$value
3. Hit the enter key
4. See error "Invalid whitespace in URL"

**Expected behavior**
We should not add a "enter" in the text field and just process the query as is

**Screenshots**
![image](https://user-images.githubusercontent.com/7620955/153600736-c3a3d61d-a86d-40b5-bcad-e946870ddd10.png)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: Latest

PR #1464, opened 2022-02-17
Disable invalid space error on url on carriage returns

Onokaev:
## Overview
Solves #1454 

### Demo
image

image

## Testing Instructions
Type a long url on the url textbox. Press enter a couple of times. Notice that the error is not displayed


millicentachieng (commented):
Could we also auto-remove leading whitespaces instead of notifying the user?

src/tests/utils/sample-url-generation.spec.tsx
```diff
@@ -80,14 +80,15 @@ describe('Sample Url Generation', () => {
 
 describe('hasWhiteSpaces should', () => {
   const invalidUrls = [
-    {url: ' https://graph.microsoft.com/v1.0/me', output: true},
+    {url: ' https://graph.microsoft.com/v1.0/me', output: false},
```
thewahome, line 83: *question*: Why should we allow the whitespace before the string?

src/app/utils/sample-url-generation.ts
```diff
@@ -82,6 +82,33 @@ export function removeExtraSlashesFromUrl(url: string): string {
...
 export function hasWhiteSpace(url: string): boolean {
   const parts = url.split('?');
-  const whitespaceChars = [' ', '\t', '\n', '%20'];
-  return whitespaceChars.some((char) => parts[0].includes(char));
+  const whitespaceChars = [' ', '\t', '\n', '%20', '\r'];
+  const allParts = url.split(/(%20|\s)/);
+  return getWhiteSpace(parts, allParts, whitespaceChars);
+}
+
+export function getWhiteSpace(parts: string[], allParts: string[], whitespaceChars: string[]): boolean {
```
thewahome, line 90: *Nitpick*: The function is checking the presence of whitespaces. The name `getWhiteSpace` suggests we are going to do something with the whitespaces. This can be renamed to something clearer or of a boolean nature. Like `whiteSpaceExists`

src/app/utils/sample-url-generation.ts
```diff
@@ -82,6 +82,33 @@ export function removeExtraSlashesFromUrl(url: string): string {
...
+  const urlHasArgs = parts.length > 1;
+  if (urlHasArgs) {
+    const hasWhiteSpaceBeforeArgs = whitespaceChars.some((char) => parts[0].trimStart().includes(char));
```
thewahome, line 93: Looks like trimming the start of the content might remove existing spaces and cause the URL to fail the whitespace test

src/tests/utils/sample-url-generation.spec.tsx
Onokaev: This was a suggestion by @millicentachieng. The leading spaces do not affect the url. You can still successfully run a query with leading spaces.

thewahome:
Noticed that enter still moves to a new line inside the text area. From the issue, the user expects that it should just run the query

thewahome:
> Noticed that enter still moves to a new line inside the text area. From the issue, the user expects that it should just run the query

Have we considered adding `event.preventDefault()` to [this line](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/blob/87c22281d01d6aac3b6c2764a4ccee1dbf834873/src/app/views/query-runner/query-input/auto-complete/AutoComplete.tsx#L148)?

src/tests/utils/sample-url-generation.spec.tsx
thewahome: ooh that's alright

Onokaev:
> Have we considered adding `event.preventDefault()` to [this line](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/blob/87c22281d01d6aac3b6c2764a4ccee1dbf834873/src/app/views/query-runner/query-input/auto-complete/AutoComplete.tsx#L148)?

Thank you! This fixes it

thewahome (approved):

PR #1481, opened 2022-02-22
Chore: reorganise test files

thewahome:
## Overview

We should keep component test files next to the component under test. Tests can be moved to a more global location if they are testing more than one function. e.g a scenario where more than one component is responding.

Code that changes together should stay together


Onokaev (approved):
Looks good

PR #1495, opened 2022-02-28
Fix: try-it experience layout

thewahome:
## Overview

The layout should take the entire space in the iframe when the app is loaded in try-it


Onokaev (commented):
Nice catch
The sign in button and the message bar at the top of the page have been pushed to the far left of the page. We can increase the left margin to push them to the right so that they are all aligned
image

Onokaev (approved):
Looks good!

PR #1497, opened 2022-03-01
Tests: move spec files closer to file under test

thewahome:
## Overview

Files that change together should stay together


Onokaev (approved):
Looks good

Issue #1456, opened 2022-02-14
Modify permissions tab shows incorrect information when the full permissions panel is open


millicentachieng:
**Describe the bug**

**To Reproduce**
1.  Have `https://graph.microsoft.com/v1.0/me` in the query textbox
2. Click on 'Modify permissions (preview)' tab
3. You should see the list of permissions to be consented
4. Go to 'Select Permissions' for the full list of permissions 
5. Once the permissions panel appears, the information in the 'Modify permissions (preview)' tab reads 'Permissions for the query are missing on this tab...'

**Expected behavior**
If there's a list of permissions to be consented in the 'Modify permissions (preview)' tab, they should still be displayed even after the permissions panel appears

**Screenshots**
 'Modify permissions (preview)' tab before permissions panel appears
<img width="824" alt="image" src="https://user-images.githubusercontent.com/25274795/153932092-763d0fc3-2ddf-4586-8738-83f127a3ba75.png">

 'Modify permissions (preview)' tab after permissions panel appears
<img width="832" alt="image" src="https://user-images.githubusercontent.com/25274795/153933209-342bdf36-9033-4bea-ac6e-1a48aef5b3bf.png">

PR #1498, opened 2022-03-01
Fix: Show correct information on Modify Permissions tab

Onokaev:
## Overview
Solves #1456 

### Demo
image

image

## Testing Instructions
Sign in to Graph Explorer
Click on Modify Permissions tab
Notice that permissions required for the current query are loaded
On the settings menu (...) click on Select Permissions
Notice that the modify permissions tab is still displaying correct information even with the permissions panel open


src/app/services/actions/permissions-action-creator.ts
```diff
@@ -81,13 +89,17 @@ export function fetchScopes(): Function {
...
         const scopes = await response.json();
-        return dispatch(
-          fetchScopesSuccess({
-            hasUrl,
-            scopes
-          })
-        );
+
+        return permissionsPanelOpen ? dispatch(fetchFullScopesSuccess({
+          hasUrl: false,
```
thewahome, line 94: since there is an explicit name for the scopes, there would be no need for the hasUrl flag

thewahome (approved):

PR #1555, opened 2022-03-14
Fix: Parse whitespace in query parameters

### LINKED ISSUE #1544: Snippets don't work with urls containing spaces even if escaped
**Describe the bug**
Urls that contain spaces, such as most that use $filter.

**To Reproduce**
1. Enter a url like `https://graph.microsoft.com/v1.0/groups?$filter=displayName eq 'All Company'`.
2. Click to view a snippet in the language of your choice.
3. Notice that no snippet is generated.
4. Examine dev tools, notice that a devxapi returns a 400 error, as the posted HTTP message is not valid, because spaces were not escaped.
5. Try to use enter a properly escaped url.
6. Notice that no snippet is generated.
7. Notice that the http message actually sent to devxapi is still not escaped.
8. Try manually posting a correctly escaped version to devxapi.
9. Notice that a snippet is actually generated.

**Expected behavior**
A correct snippet is generated.

Obviously the URL is correctly escaped when the query is executed, but is not escaped properly for posting to the snippet API. Furthermore it seems that some characters are even getting unescaped prior to posting to the snippet api, which seems really odd.

ElinorW:
## Overview
Replaces whitespace in query parameters with a '+' sign to enable a successful code snippets fetch.
Fix #1544 

### Demo
**Before fix**: 
image

**After fix**: 
image

## Testing Instructions

* Load GE
* Paste `https://graph.microsoft.com/v1.0/groups/?$filter=displayName eq 'All Company'` and Run query
* Select 'Code Snippets' tab and view results.

### DIFF
--- a/src/app/utils/sample-url-generation.spec.ts
+++ b/src/app/utils/sample-url-generation.spec.ts
@@ -34,13 +34,15 @@ describe('Sample Url Generation', () => {
   it('destructures sample url with % sign', () => {
     const name = 'DiegoS%40m365x214355.onmicrosoft.com';
     const search = `?$select=displayName,mail&$filter=mail eq '${name}'`;
+    const parsedSearch = `?$select=displayName,mail&$filter=mail+eq+'${name}'`;
+
     const url = `https://graph.microsoft.com/v1.0/users${search}`;
 
     const expectedUrl = {
       requestUrl: 'users',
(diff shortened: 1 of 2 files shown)


Onokaev (approved):
Looks good to me.

thewahome (approved):

PR #1557, opened 2022-03-15
Fixes failing snippet calls

Onokaev:
## Overview
Fixes failing code snippet calls

### Demo
image

### Notes
image
The request body is incorrect. The http version information is on a new line.
It should be on the same line with the method and request url like so:
image

## Testing Instructions
Ensure there's a query on the query input
Open code snippets tab
Click on any code snippet
Notice that snippet is fetched successfully


Onokaev:
The fix is in the new release

PR #1603, opened 2022-04-04
Chore: Dependabot upgrades

thewahome:
## Overview

Changes target branch from dev to dependabot-upgrades to:
- reduce noise in the pipelines. All workflows and pipelines are triggered whenever there is a PR
- ensure there is an isolated branch that can be used to test out all these new changes coming in from the PR

### Notes

Future adjustments

- ensure that the target branch is always upto date with the latest changes from dev. Possibly add a github workflow triggered when a PR from dependabot is made. To checkout the branch and update from the latest dev
- refine the PR and workflow triggers to be more specific so that they are not all running when dependabot does its thing

### Package not updated
- React-intl has a breaking change. It should be updated in isolation


package.json
```diff
@@ -48,21 +48,21 @@
                 "react": "16.14.0",
                 "react-app-polyfill": "0.2.1",
                 "react-dom": "16.8.2",
-                "react-intl": "2.8.0",
+                "react-intl": "5.24.8",
```
ElinorW, line 51: this package introduces some breaking changes

ElinorW (approved):

PR #1639, opened 2022-04-08
Fix: Send proper request body type for code snippets

### LINKED ISSUE #1608: PowerShell tab under Code snippets often fails to render *-Mg* cmdlet usage
**Describe the bug**
When any action is successfully performed in Microsoft Graph Explorer the user performing the action should be able to click **Code snippets** and then **PowerShell** in the response section and see the *-Mg* cmdlet that could be used to perform the exact same action. It should also show how to craft the body of a request if PATCH or POST is performed. Instead, the **PowerShell** tab almost always returns **Snippet not available**, especially for more complex actions like PATCH or POST.

**To Reproduce**
Steps to reproduce success and failure behaviors for comparison:
1. Go to https://aka.ms/ge and sign-in as a global administrator.
2. Grant the permission required to create a domain https://docs.microsoft.com/en-us/graph/api/domain-post-domains?view=graph-rest-beta

NOTE: This failure issue is not specific to domain creation, it happens for almost all API calls and failure rate increases as the calls get more complex.

3. Use this simple call to list the domains by clicking **Run query**.

GET https://graph.microsoft.com/beta/domains

4. In the response section of a successful call, click **Code snippets** and then click **PowerShell** and it typically shows an expected result like this:

Import-Module Microsoft.Graph.Identity.DirectoryManagement

Get-MgDomain

**NOTE**: A call as simple as this has a better chance of showing how to craft the PowerShell call.

5. Click away from the **PowerShell** tab to the **Response preview** tab.
6. Create a new domain using this call and clicking **Run query**.

POST https://graph.microsoft.com/beta/domains

{
  "id": "contoso.com"
}

7. In the response section of a successful call, click **Code snippets** and then click **PowerShell** and it should show a result as the GET call did, but it fails with **Snippet not available**.  This should always work so organizations can script the same complex action using Microsoft.Graph PowerShell against thousands of objects.

**Expected behavior**
See #7 in the **To Reproduce** section above.

**Screenshots**

 - OS: [Windows 11]
 - Browser [Chrome]
 - Version [99.0.4844.84 (Official Build) (64-bit)]

ElinorW:
## Overview

1. Fixes an issue where the Go and Powershell snippets wouldn't display snippets when there was a request body. E.g POST and PATCH requests
This was because the request body was being sent as a string instead of a JSON 

image

2.  This PR also adds validation to ensure we do not send a request body when there's a GET request

Fixes #1608 

### Demo
Powershell tab
image

Go Tab
image

### Notes
- Even though the other snippets tab would display info, it wasn't always correct... for example the CSharp snippet was missing the highlighted section 
image
 - Sometimes the snippets may still be unavailable due to a null reference on the API side or the resource may not be available in the schema used.

## Testing Instructions

* Open Graph Explorer
* Select version POST, version 'Beta, and input query as 'https://graph.microsoft.com/beta/domains'
with a body similar to this.
image
* Select Code Snippets tab
* Observe Powershell and Go snippets are displayed
* Try with other POST/PATCH queries as well

### DIFF
--- a/src/app/services/actions/snippet-action-creator.ts
+++ b/src/app/services/actions/snippet-action-creator.ts
@@ -30,6 +30,8 @@ export function getSnippetPending(): any {
 export function getSnippet(language: string): Function {
   return async (dispatch: Function, getState: Function) => {
     const { devxApi, sampleQuery } = getState();
+    let body: string = '';
+
     try {
       let snippetsUrl = `${devxApi.baseUrl}/api/graphexplorersnippets`;
 
@@ -56,13 +58,17 @@ export function getSnippet(language: string): Function {
 
       const requestBody =
         sampleQuery.sampleBody &&
-          Object.keys(sampleQuery.sampleBody).length !== 0 && // check if empty object
-          sampleQuery.sampleBody.trim() !== ''
+          Object.keys(sampleQuery.sampleBody).length !== 0
           ? JSON.stringify(sampleQuery.sampleBody)
           : '';
 
-      // eslint-disable-next-line max-len
-      const body = `${sampleQuery.selectedVerb} /${queryVersion}/${requestUrl + search} HTTP/1.1\r\nHost: graph.microsoft.com\r\nContent-Type: application/json\r\n\r\n${requestBody}`;
+      if(sampleQuery.selectedVerb === 'GET'){
+        // eslint-disable-next-line max-len
+        body = `${sampleQuery.selectedVerb} /${queryVersion}/${requestUrl + search} HTTP/1.1\r\nHost: graph.microsoft.com\r\nContent-Type: application/json\r\n\r\n`;
+      } else {
+        // eslint-disable-next-line max-len
+        body = `${sampleQuery.selectedVerb} /${queryVersion}/${requestUrl + search} HTTP/1.1\r\nHost: graph.microsoft.com\r\nContent-Type: application/json\r\n\r\n${requestBody}`;
+      }
 
       const options: IRequestOptions = { method, headers, body };
       const obj: any = {};


src/app/services/actions/snippet-action-creator.ts
```diff
@@ -56,13 +58,17 @@ export function getSnippet(language: string): Function {
...
 
-      // eslint-disable-next-line max-len
-      const body = `${sampleQuery.selectedVerb} /${queryVersion}/${requestUrl + search} HTTP/1.1\r\nHost: graph.microsoft.com\r\nContent-Type: application/json\r\n\r\n${requestBody}`;
+      if(sampleQuery.selectedVerb === 'GET'){
+        // eslint-disable-next-line max-len
+        body = `${sampleQuery.selectedVerb} /${queryVersion}/${requestUrl + search} HTTP/1.1\r\nHost: graph.microsoft.com\r\nContent-Type: application/json\r\n\r\n`;
+      } else {
+        // eslint-disable-next-line max-len
+        body = `${sampleQuery.selectedVerb} /${queryVersion}/${requestUrl + search} HTTP/1.1\r\nHost: graph.microsoft.com\r\nContent-Type: application/json\r\n\r\n${requestBody}`;
+      }
```
thewahome, line 71: The body string described when the verb is GET looks similar to what is in the next section except for the addition of the `requestBody` variable.

```suggestion
        // eslint-disable-next-line max-len
        const body: string = `${sampleQuery.selectedVerb} /${queryVersion}/${requestUrl + search} HTTP/1.1\r\nHost: graph.microsoft.com\r\nContent-Type: application/json\r\n\r\n`;
       
      if (sampleQuery.selectedVerb !== 'GET') {
        body += `${requestBody}`;
      }
```

ElinorW: Yeah, a GET request shouldn't have a request body...
I am working on refactoring this though and seeing how we can pass more sampleHeaders... and not hardcoding the Content-Type...

millicentachieng (approved):

PR #1658, opened 2022-04-12
Fixes sidebar and query response components on zoom

Onokaev:
## Overview
The layout should remain consistent on all zoom levels. Zooming beyond 125% distorts the app as shown:
image

### Demo
image
App remains consitent at 200% zoom

### Notes
The height property on the sidebar has been removed. The whole sidebar grows to full length at all times due to the 'align-items: stretch' css property which stretches flex items to the full width of the parent flexbox 

## Testing Instructions

* **How to test this PR**
* Check out this branch and run it locally
* Zoom the app beyond 125%
* Notice that the sidebar and content area are not distorted

### DIFF
--- a/src/app/views/App.tsx
+++ b/src/app/views/App.tsx
@@ -404,6 +404,9 @@ class App extends Component<IAppProps, IAppState> {
     } else if (minimised) {
       sidebarWidth = classes.sidebarMini;
     }
+    else{
+      sideHeight = ''
+    }
 
     this.removeFlexBasisProperty();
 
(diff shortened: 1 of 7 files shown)


thewahome (approved):

src/app/views/App.tsx
```diff
@@ -491,13 +502,11 @@ class App extends Component<IAppProps, IAppState> {
                   </div>
                 </div>
 
-                <div style={this.queryResponseStyle}>
+                <div style={{}}>
```
ElinorW, line 505: I guess you can remove the style prop if it's empty

src/app/views/App.tsx
```diff
@@ -376,7 +389,7 @@ class App extends Component<IAppProps, IAppState> {
     let sidebarWidth = classes.sidebar;
     let layout = '';
     let sideWidth = sidebar.width;
-    let sideHeight = sidebar.height;
+    const sideHeight = '';
```
ElinorW, line 392: Is this being used anywhere?

Onokaev: Yes, on the resizable component.
image
The value however doesn't change. Let me remove it and pass an empty string to the height property

ElinorW (approved):

Issue #1647, opened 2022-04-11
[Overall layout] Show all features on the left pane on the default view


RabebOthmani:
As a user, when I first land on Graph Explorer
I would like the left side pane to show all tabs
So, I’m presented with all features available to me 

**Acceptance Criteria**

- Responsive design 
- Accessibility
- When I first land on Graph Explorer, I can see Sample queries, Resource Explorer, History tabs with sample queries tab being selected and in view
- When I resize the pane (expand/collapse), the tab selection should not change

PR #1664, opened 2022-04-13
Task/persist sidebar selection

Onokaev:
## Overview
Implements #1647 

### Notes

## Testing Instructions

Check out this branch locally
Click on the history tab
Now click on the hamburger menu on the top left to hide the sidebar
Click on the hambugger menu again to restore the sidebar
Notice that the selection stays on the history tab


src/app/views/sidebar/Sidebar.tsx
```diff
@@ -6,10 +6,25 @@ import { translateMessage } from '../../utils/translate-messages';
...
+  currentTab: string;
+  setSidebarTabSelection: Function;
+}
+export const Sidebar = (props: ISidebar) => {
+
+  const onPivotItemClick = (item?: PivotItem) => {
+    if (!item) { return; }
+    const key = item.props.itemKey;
+    if (key) {
+      props.setSidebarTabSelection(key);
```
thewahome, line 20: *question:* Why can't the sidebar store this information. It looks like it's the only component that should care about which pivot item was selected

Onokaev: The Sidebar is re-rendered a fresh each time it is collapsed. This means that all local state is 're-set'. This is why I lifted the state to the App component

thewahome (approved):

PR #1687, opened 2022-04-28
Fix: Feedback rating

### LINKED ISSUE #1667: Fix the NPS scale to align with OCV scale
image

Onokaev:
## Overview
Fixes #1667 

### Demo
image

### Notes

## Testing Instructions

* How to test this PR
Check out this branch
Use the app for a minute
A feedback popup will appear with the new ratings


src/app/views/query-runner/request/feedback/uiStrings.ts
```diff
@@ -6,7 +6,8 @@ export const uiStringMap: any = {
   Prompt_YesLabel: translateMessage('Sure'),
   Prompt_NoLabel: translateMessage('Sure'),
   Graph_Explorer_Rating_Question: translateMessage('Graph Rating Question'),
-  Rating_Values_1: `1 - ${translateMessage('Not at all likely')}`,
```
ElinorW, line 9: I know I had suggested removing the strings, this was because of the wrong interpretation on the OCV side. Now that everything is working as expected I think you can just return them
(i.e 0 - not at all likely and 10 - Extremely likely)

Onokaev: Awesome. Done

PR #1714, opened 2022-05-05
Fix: panel overlay

Onokaev:
## Overview
Fixes overlay problem when the permissions panel is opened

### Demo
image

image

image

### Notes

## Testing Instructions

* How to test this PR
Check out this branch
Start the local instance and sign in
Open permissions panel
Notice the fix


src/app/views/authentication/profile/Profile.tsx
```diff
@@ -160,6 +162,51 @@ const Profile = (props: any) => {
...
+  }
+
+  const menuProperties = {
+    shouldFocusOnMount: true,
+    alignTargetEdge: true,
+    items
+  };
+
+  const panelOverlayProps : IOverlayProps = {
+    isDarkThemed: true
```
thewahome, line 206: *Question*
What is the effect on when the theme is not dark?

PR #1740, opened 2022-05-17
Chore: Consolidate github workflows

### LINKED ISSUE #1513: Speed up tests
**Describe the bug**
The test suites are running slowly. 
The tests need to work faster in local and pipeline environments.

thewahome:
## Overview
Consolidates related pipeline actions together so that they can share resources.
Fixes #1513
Makes things look pretty 🥇 

image

## Notes
Two concepts were used in reducing the time it takes to run tests from +25 minutes to ~12 minutes
1. Caching
2. Sharding

**Caching and reuse**
Some of the steps were being re-run for a couple of the test steps. Each time the steps were running afresh they'd take a long time. Now we can cache the results of the first step and reuse the output in the next step. Helping us to reduce the need to run the steps each time. 

The caching should last across runs so subsequent runs will benefit from the cache as long as the package-lock.json file does not change.

**Tests sharding**
Dividing the tests into batches and having them run in parallel helps us reduce the time taken to run the tests. It also adds the chance for us to see which batch is slowing things down and can allow us to know which specific test to improve to speed up time.


.github/workflows/linter.yml
```diff
@@ -7,43 +7,196 @@ on:
...
+        id: cache-node-modules
+        with:
+          path: |
+            node_modules
+            packages/*/node_modules
+          key: modules-${{ hashFiles('package-lock.json') }}
+
+      - name: Install dependencies
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
+        run: npm ci --no-audit
```
Onokaev, line 47: This step is a bit unclear. Does it install dependencies or run accessibility checks or both?

ElinorW (approved):

PR #1761, opened 2022-05-30
Fix: Restore permissions styles

Onokaev:
## Overview
Restores styling on the permissions panel:
image

image

Fixes alignment issue on the Graph explorer label on mobile screen
image

image

### Demo

### Notes

## Testing Instructions

* How to test this PR
Check out this branch
Run it locally
Resize screen from mobile to full screen view. Notice fix

### DIFF
--- a/src/app/views/authentication/auth-util-components/UtilComponents.styles.ts
+++ b/src/app/views/authentication/auth-util-components/UtilComponents.styles.ts
@@ -1,0 +1,18 @@
+import { ITheme } from '@fluentui/react'
+
+export const utilComponentStyles = (theme: ITheme) => {
+  return {
+    actionButtonStyles: {
+      root: {
+        height: 50,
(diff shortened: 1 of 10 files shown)


src/app/views/authentication/auth-util-components/UtilComponents.styles.ts
```diff
@@ -0,0 +1,18 @@
+import { ITheme } from '@fluentui/react'
+
+export const utilComponentStyles = (theme: ITheme) => {
```
thewahome, line 3: Though we have always called it `UtilComponent`, can we rename it to something that says exactly what it does?

Onokaev: Sure. Let me rename it to ProfileButton

ElinorW (approved):

Issue #1774, opened 2022-05-31
[Auto-Complete] Suggest IDs for auto-complete


RabebOthmani:
To be able to suggest IDs, we will use this extension https://docs.microsoft.com/en-us/connectors/custom-connectors/openapi-extensions#use-dynamic-values

This works depends on the CSDL to OpenAPI conversion tool, so until the extensions are generated on that side, this ticket is on hold.

thewahome:
Should also close https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/682

darrelmiller:
@thewahome The dynamic-values extension allows a client to make a secondary call to return a list of actual Id values that are value to use in the call for the current tenant based on the the current URL.

What I am seeing in the PR you referenced is the Id variable name being displayed in the dropdown.

![image](https://user-images.githubusercontent.com/447694/174218217-a69fbc3b-998e-42ca-8879-746dca28d48f.png)

That was not my intention with this feature.  Sorry for not including more clarity in the original issue, but this issue was put on hold because the GE work cannot be done until we have the dynamic-values extension in the OpenAPI description.  

@adhiambovivian  What was the reason for putting this issue that was OnHold in the June milestone?

/cc @ddyett

thewahome:
@darrelmiller 
I proposed that it goes into this milestone because I had found a way to suggest IDs while using the open API tree.

It didn't occur to me to think of it as ID values so it seemed as though showing that an ID was needed as the next path segment would suffice.

thewahome:
@RabebOthmani / @darrelmiller 
Since we might not know when the extensions will be complete and there was this issue #682 that I had opened 2 years ago that is fulfilled. I can keep this current one open and on hold while closing the other.

My thoughts are there is value in knowing that the path should have an ID at that point and I assume we will be replacing this ID suggestion with a real value when that extension is done

darrelmiller:
@thewahome Keeping this issue open and closing the other one seems like the right move to me.  Yes, this has been a long term "dream feature".  It resurfaced recently because of the work that Koros and Sunday have been doing in Graph Connectors that requires the use of these dynamic extensions and it seemed more likely that we would get to work on adding those extensions to our OpenAPI in the coming semester.

[...]

PR #1802, opened 2022-06-02
Feature: autocomplete typeahead

### LINKED ISSUE #1465: Speed up autocomplete suggestions
For the resource explorer to work, the application fetches a links only subset of the open-API description. We can leverage this since it is local and can reduce the calls made to the api

Only check the API for query parameters since they do not come in with the open-API description loaded on page load. Since the links are already available, simply check in state to get the next part of the url. e.g. `me/` should yield a list straight from the state and not from a network call

Keeping it as is. No complaints have been made so far about the number of network calls

**Additional context**
Reducing network dependency would make autocomplete search faster

thewahome:
## Overview

Closes #1465 

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description


src/app/views/query-runner/query-input/auto-complete/AutoComplete.tsx
```diff
@@ -72,373 +75,182 @@ class AutoComplete extends Component<IAutoCompleteProps, IAutoCompleteState> {
...
-      } else {
-        if (!this.props.autoCompleteOptions || `${requestUrl}` !== this.props.autoCompleteOptions.url) {
-          this.props.actions!.fetchAutoCompleteOptions(requestUrl, queryVersion);
-        } else {
-          this.performLocalSearch(url);
-        }
+    const theSuggestions = (shouldSuggestVersions) ? GRAPH_API_VERSIONS : getSuggestions(url, autoCompleteOptions!);
+    if (theSuggestions.length > 0) {
+      const filtered = (searchText) ? getFilteredSuggestions(searchText, theSuggestions) : theSuggestions;
+      if (filtered[0] !== searchText) {
```
Onokaev, line 171: What's the effect of showing suggestions when the searchText is the first index?
I don't understand what this is achieving

src/app/views/query-runner/query-input/auto-complete/auto-complete.util.ts
```diff
@@ -13,60 +14,38 @@ export function cleanUpSelectedSuggestion(compare: string, userInput: string, se
...
-  const max = availableSymbols.reduce((prev, current) => (prev.value > current.value) ? prev : current);
-  return max;
-}
-
 // Filter out suggestions that don't contain the user's input
-export function getFilteredSuggestions(compareString: string, suggestions: string[]) {
+function getFilteredSuggestions(compareString: string, suggestions: string[]) {
   return suggestions.filter((suggestion: string) => {
     return suggestion.toLowerCase().indexOf(compareString.toLowerCase()) > -1;
   });
```
Onokaev, line 25: Suggestion: We can sort the suggestions so that the ones that match the search text appear first on the list as Sunday suggested. We can achieve this with a custom sort that checks how close a suggestion is to the searchtext.
This is not within the scope of this task, but it's something we can implement later

src/app/views/query-runner/query-input/auto-complete/AutoComplete.tsx
thewahome: This fixes an edge case where the query parameters ($select .. etc) even after being selected would still show up as suggestions instead of showing the next round of suggestions after the parameters

Onokaev (dismissed):

ElinorW:
For some reason, adding a slash causes the suggestions box to really expand 
image
I am using the app in half screen here...

ElinorW:
An accessibility issue I noticed. 
When I select my autocomplete suggestion, I am unable to navigate out of the textbox using keyboard navigation. I should be able to tab and move focus onto the 'Run Query' Button

darrelmiller:
Regarding this conversation https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/1774#issuecomment-1158904910  

Do we believe that inserting the Id token is a better experience?  Do we have any evidence to support it? Do we have evidence that customers were having difficulty knowing that they need to provide an Id?

I'm not sure.  I think the token might be helpful, but I am also reluctant to change the current behavior without some level of confidence that it is a good change.

RabebOthmani:
@thewahome regarding the public versions, If I select beta from the drop down list, the query automatically updates to https://graph.microsoft.com/v1.0/meagreementAcceptances and when I select v1 , it shows https://graph.microsoft.com/v1.0/meadmin

thewahome:
@RabebOthmani does that happen when the url is basic, e.g https://graph.microsoft.com or when it has content after like https://graph.microsoft.com/v1.0/admin ?

RabebOthmani:
@thewahome  basic url , not sure I understand the part about when the url has content. This is related to the dropdown so when it has content and I click delete until I'm back to the version level I get the same issue

thewahome:
> @thewahome basic url , not sure I understand the part about when the url has content. This is related to the dropdown so when it has content and I click delete until I'm back to the version level I get the same issue

I've just fixed it right now. It should work fine...

ElinorW (commented):
Noticed a few more details
1) when I invoke the suggestions list e.g. by adding a slash after me, then I proceed to click on a sample query, the suggestions list doesn't close 
2) clicking away doesn't close the list either, I had to click on the textfield then click away for the list to disappear

ElinorW (approved):
This is great work! LGTM! 🚀🙌🏾

PR #1810, opened 2022-06-06
Add headers to snippet request

### LINKED ISSUE #1657: Request headers info isn't present on code snippets
**Describe the bug**
In cases where queries have some request headers, the code snippets don't display some header information

**To Reproduce**
1. Go to [Graph Explorer](https://developer.microsoft.com/en-us/graph/graph-explorer)
2. Click on the **Applications** group
3. Scroll down to 
image
5. Click on the **code snippets** tab and view _CSharp_ tab
6. See missing header information

**Expected behavior**
Some of the snippets e.g the CSharp one should show information about the headers

**Screenshots**
This is how it appears on the snippets
image

This is how it should appear (according to the [docs](https://docs.microsoft.com/en-us/graph/api/application-list?view=graph-rest-1.0&tabs=csharp#code-try-13)) --> **Notice the missing header**
image

 - OS: Windows
 - Browser: MS Edge

**Additional context**

Onokaev:
## Overview
Fixes #1657 

### Demo

### Notes

- It is incorrect to send content-type: application/json for GET requests, so we skip adding this content-type if the request method is GET. 

- We are sending over content-type: application/json by default for all requests as we currently only support JSON request body.
  If we start supporting other content-types in future, we won't need to change the way we handle content-types as that is 
  covered by this PR.

- The responses for Javascript snippets expected to have headers, do not have these headers except for GET queries. 
  This query for instance: 
      **GET https://graph.microsoft.com/v1.0/applications?$search="displayName:Web"&$count=true
      ConsistencyLevel: eventual**
   will return the correct Javascript snippet with the header included. 
   
   However, Javascript snippet response for queries with methods other than GET, do not include headers.
   For instance:
         **POST https://graph.microsoft.com/v1.0/users
         Content-type: application/json
         DummyHeader: dummyValue**
         (https://developer.microsoft.com/en-us/graph/graph-explorer?request=users&method=POST&version=v1.0&GraphUrl=https://graph.microsoft.com&requestBody=eyJhY2NvdW50RW5hYmxlZCI6dHJ1ZSwiY2l0eSI6IlNlYXR0bGUiLCJjb3VudHJ5IjoiVW5pdGVkIFN0YXRlcyIsImRlcGFydG1lbnQiOiJTYWxlcyAmIE1hcmtldGluZyIsImRpc3BsYXlOYW1lIjoiTWVsaXNzYSBEYXJyb3ciLCJnaXZlbk5hbWUiOiJNZWxpc3NhIiwiam9iVGl0bGUiOiJNYXJrZXRpbmcgRGlyZWN0b3IiLCJtYWlsTmlja25hbWUiOiJNZWxpc3NhRCIsInBhc3N3b3JkUG9saWNpZXMiOiJEaXNhYmxlUGFzc3dvcmRFeHBpcmF0aW9uIiwicGFzc3dvcmRQcm9maWxlIjp7InBhc3N3b3JkIjoiMTJmNTU1ODUtNjJjYy0zZThkLWE0NmMtOTM4ZDA4OWI1YTJkIiwiZm9yY2VDaGFuZ2VQYXNzd29yZE5leHRTaWduSW4iOmZhbHNlfSwib2ZmaWNlTG9jYXRpb24iOiIxMzEvMTEwNSIsInBvc3RhbENvZGUiOiI5ODA1MiIsInByZWZlcnJlZExhbmd1YWdlIjoiZW4tVVMiLCJzdGF0ZSI6IldBIiwic3RyZWV0QWRkcmVzcyI6IjkyNTYgVG93bmUgQ2VudGVyIERyLiwgU3VpdGUgNDAwIiwic3VybmFtZSI6IkRhcnJvdyIsIm1vYmlsZVBob25lIjoiKzEgMjA2IDU1NSAwMTEwIiwidXNhZ2VMb2NhdGlvbiI6IlVTIiwidXNlclByaW5jaXBhbE5hbWUiOiJNZWxpc3NhREBjb250b3NvLmNvbSJ9&headers=W3sibmFtZSI6IkR1bW15SGVhZGVyIiwidmFsdWUiOiJkdW1teVZhbHVlIn0seyJuYW1lIjoiQ29udGVudC10eXBlIiwidmFsdWUiOiJhcHBsaWNhdGlvbi9qc29uIn1d)
   The response for all other snippets have the dummy header:
   
image
    The Javascript snippet does not have this header:
    
image

## Testing Instructions

* How to test this PR
Check out this branch and run it locally
Add request headers on the request headers tab
Click on code snippets tab
Notice that the snippet response has all headers


src/app/services/actions/snippet-action-creator.ts
```diff
@@ -86,3 +87,13 @@ export function getSnippet(language: string): Function {
...
   };
 }
+
+export function constructHeaderString(sampleHeaders: Header[]): string {
+  let headers = '';
+  if (sampleHeaders && sampleHeaders.length > 0) {
+    sampleHeaders.forEach(header => {
+      headers += `${header.name}: ${header.value}\r\n`;
+    });
+  }
```
thewahome, line 97: We should add 'Content-Type: application/json' header if the content-type header doesn't already exist

Onokaev: Okay. Let me add that check

src/app/services/actions/snippet-action-creator.ts
```diff
@@ -86,3 +92,32 @@ export function getSnippet(language: string): Function {
...
+
+  headersString += !isContentTypeInHeaders && selectedVerb !== 'GET' ? 'Content-Type: application/json\r\n' : '';
+  return headersString;
+}
+
+function getHeaderStringProperties(sampleHeaders: Header[]): IHeaderStringProperties {
+  let constructedHeader = ''
+  let isContentTypeAvailable: boolean = false;
+  sampleHeaders.forEach((header: Header) => {
+    isContentTypeAvailable = isContentTypeAvailable ? isContentTypeAvailable : checkIfContentTypeAvailable(header);
```
thewahome, line 114: Not sure that this will evaluate. 
It looks like it has already set isContentTypeAvailable to false so line 114 will always run the function for checkIfContentTupeAvailable(header)
Is running the check the intention inside the foreach?

Onokaev: Yes. The intention is to run the check for each header until we find a content-type header. I think I can make it simpler with a .find() or .includes() array methods

thewahome (approved):

PR #1837, opened 2022-06-14
Task: Remove .spec.tsx tests add tests on App.spec.tsx file

### LINKED ISSUE #1763: Speed up tests
**Describe the bug**
We are currently not running tests for .tsx files because they are slow. The .tsx files recursively render components.

For instance, let's say component A has inner components B and C.
- When testing component A, it will render component B and C. To fix this, we can mock component B and C so as to test component A in isolation.

However, we do not need to test all the tsx components individually. We can just test the root App component that renders all other components. From this single test, we can make all necessary assertions to ensure all components are rendered as expected.

This will be an extension of this effort: https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/1513

- [ ] Remove .spec.tsx files and add assertions on App.spec.tsx
- [ ] Add config changes for speeding up tests

Onokaev:
## Overview
Fixes #1763 

### Demo
Sample of the tests on App.spec.tsx file
image

### Notes
Since we are testing what the end user sees and not implementation details, all necessary assertions have been made on the App.spec.tsx file. 
The unnecessary tests on the other .tsx files have been removed.
The tests confirm that all major sections of the app have been rendered, and that what we expect our users to see are available. 

## Testing Instructions

* How to test this PR
Check out this branch
Run 'npm run test'
Notice that all tests run in under 2 minutes


src/app/views/App.spec.tsx
```diff
@@ -62,52 +64,193 @@ const renderApp = (args?: any) : any => {
...
-jest.mock('@ms-ofb/officebrowserfeedbacknpm/scripts/app/Window/Window', () => ({
-  OfficeBrowserFeedback: Object
-}))
+// eslint-disable-next-line no-console
+console.warn = jest.fn();
 
-jest.mock('@ms-ofb/officebrowserfeedbacknpm/Floodgate', () => ({
-  makeFloodgate: Object
-}))
+describe('It should render the main GE site', () => {
```
thewahome, line 85: There are conventions for naming tests so that they are easy to understand. Like:` It/Should` or `Given/When/Then`

In this case it looks like we are using It/Should

So that the tests are easy to understand and also to make them read like logical prose, we should look at `Describe` as the title/ first part of the sentence of the group of tests

Like `​Describe('App rendering')​`

then `​it` sections​ should talk more about what was described like: `​it('should return x when y')​`

thewahome (changes requested):

src/app/views/App.spec.tsx
Onokaev: I noticed that most of the tests did not follow this convention. I have fixed them.
It's good you raised it!

PR #1956, opened 2022-07-26
Fix: Use resources as a source of truth

### LINKED ISSUE #1954: Use OpenApiUrlTreeNode as source of truth for Autocomplete
**Describe the bug**
DevX API is receiving a lot of error 404s for invalid urls not found in Graph from GE autocomplete.
Upon closer inspection we've noticed that a call is made to the autocomplete endpoint every time a user types `/` after a segment that's unavailable from the suggestions and for every subsequent character typed after this (before one runs the query).

**To Reproduce**
1. Go to [Graph Explorer](https://developer.microsoft.com/en-us/graph/graph-explorer)
2. Go to the autocomplete tab
3. Open the developer tools tab (Ctrl+Shift+I) and select the `Network` tab
4. Select a query segment available from the suggestions provided  e.g `/me/drive/bundles`. 
5. Add another `/` and check the Network tab. Observe that requests are sent to autocomplete after typing the `/` and for every subsequent character typed after this.

**Expected behavior**
Multiple 404s should not be recorded for just one query. Also, if the query isn't present in the suggestions (derived from the `OpenApiUrlTreeNode` in Resource Explorer) don't make a call to the autocomplete endpoint. This will help reduce the failures coming into DevX API.

**Screenshots**
1.  image

3.  image

thewahome:
## Overview

Fixes #1954

It prevents making calls to the API unless when fetching for query parameters

### DIFF
--- a/src/modules/suggestions/suggestions.ts
+++ b/src/modules/suggestions/suggestions.ts
@@ -26,6 +26,12 @@ class Suggestions implements ISuggestions {
       return localOptions;
     }
 
+    /*
+    * blocks calls to the API for path segments
+    * uses resource explorer and cache as source of truth
+    */
+    if (context === 'paths') { return null; }
+
     return this.fetchSuggestionsFromNetwork(url, api, version);
   }
 


Onokaev (approved):
LGTM!

ElinorW (approved):

PR #1984, opened 2022-08-02
Fix: Add exception provision for spaces in query url

### LINKED ISSUE #1893: Invalid whitespace in URL error in Graph Explorer with properly encoded URL
**Describe the bug**
When providing a properly formatted URL using %20 to represent a 'space', the Graph Explorer GUI indicates there is a invalid whitespace in the URL.

**To Reproduce**
1. Create folder in personal OneDrive called 'Test Folder'
2. Open Graph explorer, login, and grant proper permissions
3. In URL field use : https://graph.microsoft.com/v1.0/me/drive/root:/Test%20Folder
4. Error below URL field will say 'Invalid whitespace in URL'. However you can still click 'Run Query' and get a 200 response with data.

I provided a screenshot below to showcase the error and it still returning a response with data.

**Expected behavior**
with a properly encoded URL using %20 for spaces, I don't expect the error 'Invalid whitespace in URL'

**Screenshots**

image

 - OS: Windows 11
 - Browser: Edge
 - Version: 103.0.1264.44

Onokaev:
## Overview
Closes #1893 

### Demo
image

### Notes

## Testing Instructions

* How to test this PR
On the query textbox, type a query like /me/drive/root:/New Folder
Notice that a whitespace error is not raised
Add a space between /me and /. Notice that a whitespace error is now displayed


ElinorW (commented):
Why can't we do the whitespace validation for only the graph url + version part? 'https://graph.microsoft.com/(v1.0/beta)/'. Like instead of splitting at the '?' we could split after the 4th '/' and validate what's before that slash instead...
It seems queries after the version could have spaces without necessarily having a '?' so instead of checking all these how about reducing the validation to only spaces between the url and the verson part.

Onokaev:
> url and the verson part

That's a good idea. However, we still do not expect spaces in some areas after the version information. Here are some examples:
v1.0/root:/ SomeName -> We do not want the space before SomeName
v1.0/root:/SomeName /SomeOtherName -> We do not also want the space after SomeName and before the slash. This is only allowed when there is no slash at the end of SomeName

thewahome:
Closing this as it is taken care of by #2044

Issue #1990, opened 2022-08-02
New issues should automatically be added to the project boards.


ddyett:
Use a Github task to address this.

PR #1992, opened 2022-08-02
chore: automate adding issues to project board

gavinbarron:
## Overview

Closes: #1990 

Adds a github action to push newly created issues to the project board.

### Notes

This workflow uses secrets that are already configured in organization.


gavinbarron:
Having taken a look at some of our other repositories, I'm going to update this PR to use the same action that we already use on the SDKs e.g. https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/dev/.github/workflows/projectbot.yml which has the advantage of using secrets that are already configured from via the microsoftgraph organization.

ElinorW (approved):

PR #2041, opened 2022-08-16
Fix: Speed up autocomplete

### LINKED ISSUE #2005: Unfriendly Auto-complete editing experience
With the new work done on Auto-complete , it now tried to complete on every key press not just the slash making the editing experience challenging. 
Feedback received from a user:" entering a query is now broken. for each character typed into the query box, a suggestion is fetch and that locks and hangs the UI. We cannot type a query anymore." 

Also note that when fetching suggestions, the substring search is done while it should be a 'starts with'

Onokaev:
## Overview
Fixes #2005 

### Demo

### Notes
A debounce function has been added that allows you to type, and fetches suggestions 200 milliseconds after you've stopped typing

## Testing Instructions

* How to test this PR
Check out this branch
Type queries on the query textbox
Notice that you can type and delete queries without the UI hanging

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8664,11 +8664,6 @@
       "resolved": "https://registry.npmjs.org/functions-have-names/-/functions-have-names-1.2.2.tgz",
       "integrity": "sha512-bLgc3asbWdwPbx2mNk2S49kmJCuQeu0nfmaOgbs8WIyzzkw3r4htszdIi9Q9EMezDPTYuJx2wvjZ/EwgAthpnA=="
     },
-    "fuzzysort": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/fuzzysort/-/fuzzysort-2.0.1.tgz",
-      "integrity": "sha512-SlgbPAq0eQ6JQ1h3l4MNeGH/t9DHKH8GGM0RD/6RhmJrNnSoWt3oIVaiQm9g9BPB+wAhRMeMqlUTbhbd7+Ufcg=="
-    },
     "gensync": {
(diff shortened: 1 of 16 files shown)


ElinorW:
The autocomplete suggestions are definitely faster and the suggestions feel more intuitive. 

However in the case where I type in a string such as 'cale' 
image

I notice suggestions that don't exactly match my string but rather substrings of my string. 
As you scroll further down the list the suggestions are anything that have a 'c', 'a', 'l', 'e'
Not sure if this is the intended design. I would think that by the time I have typed out a 4-letter string... the suggestions list would only include items that include the 4-letter string in that order...
Please share you thoughts on this with me

Onokaev:
> The autocomplete suggestions are definitely faster and the suggestions feel more intuitive.
> 
> However in the case where I type in a string such as 'cale' image
> 
> I notice suggestions that don't exactly match my string but rather substrings of my string. As you scroll further down the list the suggestions are anything that have a 'c', 'a', 'l', 'e' Not sure if this is the intended design. I would think that by the time I have typed out a 4-letter string... the suggestions list would only include items that include the 4-letter string in that order...
> 
> I would appreciate your thoughts on this

Thank you for this @ElinorW. We have had received a lot of feedback, and most people prefer seeing suggestions related to their search text. So we have updated this to only show you `startsWith` and `includdes` suggestions

ElinorW (approved):

thewahome:
@Onokaev  Let's remember to remove the check for whether the search text matches the suggestions provided so that we give the user a chance to dismiss the suggestions themselves

PR #2056, opened 2022-08-23
Task: Add device characteristics telemetry

### LINKED ISSUE #2001: Add telemetry on device characteristics
In order to get responsiveness right and provide a good experience for our customers, we first need to understand what devices and screen resolutions are most used. 
The goal here is to add telemetry to capture screen resolution and other device characteristics that will help us understand usage patterns and build appropriately.

MaryannGitonga:
## Overview

Closes #2001

## Demo
image

### How to test this PR
* Checkout to this branch.
* Load your local GE instance with the network section of the DevTools open.
* Run any query on GE.
* On the network section, filter the track requests and under Request Payload, find a data object with the baseType of "EventData".
* Under it, there is baseData with the name "DEVICE_CHARACTERISTICS_EVENT" with the device characteristics telemetry
* Resize the browser or change the device scale ratio and/or the browser zoom levels and observe telemetry.
* Run a query on GE and observe telemetry on the network section.


src/app/views/App.tsx
```diff
@@ -84,10 +84,56 @@ class App extends Component<IAppProps, IAppState> {
...
+      deviceWidth: 0,
+      screenWidth: ''
+    };
+
+    const deviceScreenSize = this.getDeviceScreenSize(deviceWidth);
+
+    deviceProperties.deviceHeight = deviceHeight;
+    deviceProperties.deviceWidth = deviceWidth;
+    deviceProperties.screenWidth = deviceScreenSize;
+
```
thewahome, line 122:
```suggestion
     const deviceProperties = {
      deviceHeight: screen.height,
      deviceWidth: screen.width,
      screenWidth: this.getDeviceScreenSize(screen.width)
    };
```

thewahome (approved):

src/app/views/App.tsx
```diff
@@ -84,10 +85,27 @@ class App extends Component<IAppProps, IAppState> {
...
+    telemetry.trackDeviceCharacteristics(deviceProperties);
+  }
+
   public componentDidMount = async () => {
     this.displayToggleButton(this.mediaQueryList);
     this.mediaQueryList.addListener(this.displayToggleButton);
 
+    const hasTelemetryBeenCollected = localStorage.getItem('hasTelemetryBeenCollected');
+
+    if(!hasTelemetryBeenCollected){
```
gavinbarron, line 104: This results in a single piece of telemetry collected per browser, we don't find out what dimensions are being used if the user resizes the screen, which is a very real scenario.

My suggestion is that we capture this information in a way that appears redundant so that we can also look at how many graph requests are being made by screenWidth bucket.

src/app/views/App.tsx
```diff
@@ -84,10 +85,27 @@ class App extends Component<IAppProps, IAppState> {
     });
   }
 
+  private collectDeviceCharacteristicsTelemetry() {
+    const deviceProperties = {
+      deviceHeight: screen.height,
```
gavinbarron, line 90: unfortunately, these dimensions are the device dimensions, not the dimensions that is available to the browser to use.
Here's the telemetry from my browser:
image

But using window.innerHeight and window.innerWidth I get 859 and 1553 because I have resized the window to better use my screen.

During a call we also discussed capturing window.devicePixelRatio to help identify where users are being bumped down to lower effective resolutions due to device settings.

src/app/utils/device-screen-size.ts
```diff
@@ -0,0 +1,19 @@
+export function getDeviceScreenSize(deviceWidth: number){
+    if(deviceWidth >= 1920){
+      return 'xxxl';
+    }
+    else if(deviceWidth >= 1366){
+      return 'xxl';
+    }
+    else if(deviceWidth >= 1365){
```
gavinbarron, line 8: this conditions is incorrect
```suggestion
    else if(deviceWidth >= 1024){
```

src/app/utils/device-screen-size.ts
```diff
@@ -0,0 +1,19 @@
+export function getDeviceScreenSize(deviceWidth: number){
```
gavinbarron, line 1: Can we please have tests to validate that this produces the expected outcome?

gavinbarron (changes requested):
I think this needs a little more work to ensure that we get the data we want and are able to understand usage patterns by screen sizes.

src/app/views/App.tsx
```diff
@@ -84,10 +87,27 @@ class App extends Component<IAppProps, IAppState> {
...
+      deviceWidth: screen.width,
+      deviceScreenSize: getDeviceScreenSize(screen.width),
```
gavinbarron, line 94: This needs to be browserSize, device stats are interesting, browser stats are the ones that impact how we render the page

src/app/utils/device-characteristics-telemetry.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    addListener: jest.fn(), // Deprecated
+    removeListener: jest.fn(), // Deprecated
+    addEventListener: jest.fn(),
+    removeEventListener: jest.fn(),
+    dispatchEvent: jest.fn()
+  }))
+});
+describe('Device Telemetry', () => {
+  it('should get device screen scale', () => {
+    expect(getDeviceScreenScale()).toBeTruthy();
```
gavinbarron, line 18: There's formatting happening inside this function, I'd like to see the test assert that the result is formatted correctly.
Think of the test as a catch for a future developer who might come along and remove the formatting and break reporting as a result, a truthy check would still pass even if the value returned changed from `'100%'` to `1`

src/telemetry/telemetry.ts
```diff
@@ -119,9 +120,19 @@ class Telemetry implements ITelemetry {
...
+      deviceWidth: screen.width,
+      browserScreenSize: getBrowserScreenSize(window.innerWidth),
+      browserHeight: window.innerHeight,
+      browserWidth: window.innerWidth,
+      scale: getDeviceScreenScale()
+    };
+    properties = properties || deviceProperties;
+
     telemetry.trackEvent(DEVICE_CHARACTERISTICS_EVENT, properties);
+    console.log('telemetry collected');
```
gavinbarron, line 135: Let's not ship this console log call

src/telemetry/telemetry.ts
```diff
@@ -119,9 +120,19 @@ class Telemetry implements ITelemetry {
...
+  public trackDeviceCharacteristicsTelemetry(properties?: any){
+    const deviceProperties = {
+      deviceHeight: screen.height,
+      deviceWidth: screen.width,
+      browserScreenSize: getBrowserScreenSize(window.innerWidth),
+      browserHeight: window.innerHeight,
+      browserWidth: window.innerWidth,
+      scale: getDeviceScreenScale()
+    };
+    properties = properties || deviceProperties;
```
gavinbarron, line 132: This should probably be the union of the two objects:
```suggestion
    properties = Object.assign(properties || {}, deviceProperties);
```
This will cause any property on properties that is not on deviceProperties to be preserved while using the value from deviceProperties for any common properties. cf: https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/object/assign

gavinbarron (approved):
🚀

Onokaev (approved):

PR #2062, opened 2022-08-29
Fix: Schema validation for Adaptive cards 

### LINKED ISSUE #2020: Yellow border on Adaptive cards tab
There is a little yellow rectangle that looks like a border? 
IT's just next the copy button and can be confusing to users as it might look like a button
border

ElinorW:
## Overview
Fix #2020 

### Demo

image

### Notes

## Testing Instructions

* Checkout branch and load GE
* Run the v1.0/me query
* Navigate to Adaptive Cards, JSON Schema tab
* Observe yellow line is no longer there

### DIFF
--- a/src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
+++ b/src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
@@ -163,7 +163,7 @@ class AdaptiveCard extends Component<IAdaptiveCardProps> {
                 <Monaco
                   language='json'
                   body={data.template}
-                  height={'800px'}
+                  height={'35vh'}
                 />
               </div>
             </PivotItem>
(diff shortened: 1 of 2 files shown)


src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
```diff
@@ -163,7 +163,7 @@ class AdaptiveCard extends Component<IAdaptiveCardProps> {
                 <Monaco
                   language='json'
                   body={data.template}
-                  height={'800px'}
+                  height={'35vh'}
```
gavinbarron, line 166: Is this intentional?
This could result in some strange heights on this control on smaller screens

ElinorW: Yeah.. but I've realized the height issue was broken because of a recent PR that adjusted the response area heights...
So on localhost, all the different Pivots (i.e Adaptive Cards, Code Snippets etc) had different heights and the page was behaving terribly...
I'll actually leave this and as it was and create a separate issue to track the height problem

Onokaev (approved):

thewahome (approved):

PR #2109, opened 2022-09-09
Fix: Update wordSeparators in Monaco

### LINKED ISSUE #553: Guid interaction
In an original Graph Explorer build, in the response that came back. It was super easy to click on the GUIDs and get some options:

**Copy ID**
Being able to just copy the GUID. Many times in code I want to grab GUIDs quickly. Selecting them in the response is hard in Chrome/Edge. This would be much smoother an experience.

**Run query**
Imagine the scenario where you run `/me/events` and then click on the event-id in the json response. This immediately changes the graph explorer address bar to `/me/events/{event-id}` and returns that response. It makes the graph more "graphy" navigating around.

We would want some obvious way that either on hover or clicking on GUID has an interaction. 
[AB#7504](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7504)

Onokaev:
## Overview
Fixes #553 

### Demo
image
By double-clicking on the ID field, we get the whole ID

### Notes
Be default, word separators string on the Monaco editor is: '`~!@#$%^&*()-=+[{]}\\|;:\'",.<>/?. We only need " as a word separator. This makes it possible to select all fields within ""

## Testing Instructions

* How to test this PR
Run a query like /v1.0/me
Double click on the ID field
Notice that the whole ID field is selected

### DIFF
--- a/src/app/views/common/monaco/Monaco.tsx
+++ b/src/app/views/common/monaco/Monaco.tsx
@@ -63,7 +63,7 @@ export function Monaco(props: IMonaco) {
               scrollBeyondLastLine: true,
               overviewRulerBorder: false,
               // eslint-disable-next-line no-useless-escape
-              wordSeparators:'`~!#$%^&*()=+[{]}\|;:\'",<>/?'
+              wordSeparators:'"'
             }}
             onChange={onChange}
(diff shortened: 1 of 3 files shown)


thewahome (approved):
Beautiful!

PR #2182, opened 2022-10-27
Task: Add Adaptive card templates to local repo

### LINKED ISSUE #2170: The Adaptive Cards service is down
**Describe the bug**
We are currently following with the Adaptive Cards Template service team to identify if the service is something we should be relying on as today its certificate is expired. https://github.com/microsoft/adaptivecards-templates/issues/302

image

**Additional context**
It might be possible that we have to rely on a "local" version of these templates as the service is still in **pre-alpha** phase and doesn't offer any support.

ElinorW:
## Overview

Fixes #2170 
Having them locally is best as the site isn't always maintained.

## Notes
- The templates are from the [Adaptive cards templates repo](https://github.com/microsoft/adaptivecards-templates).
- Commit history is preserved.

## Testing Instructions

* Launch GE
* Run GET /me query
* Click in Adaptive cards Pivot
* Observe card and JSON template are available


thewahome (approved):
LGTM

PR #2210, opened 2022-11-02
Task: Add playwright accessibility test

### LINKED ISSUE #2209: Move accessibility test to use playwright
With playwright as the preferred tool for e2e tests. Accessibility can leverage this instead of using selenium tests.

Benefits:
- Faster
- less dependency on other packages
- easier to set up since it requires less configurations

ElinorW:
## Overview

Fixes #2209 
Playwright supports accessibility testing. 
This PR uses playwright to conduct the accessibility test.


thewahome (approved):
LGTM

PR #2219, opened 2022-11-09
Feature: Surface documentation links

### LINKED ISSUE #1854: Surface documentation about a URL in the query bar
Sometimes I paste/type in/ accept suggestions for a URL and I need to know where to find the documentation about the query. With a minimized sidebar or an already slimmed-down sidebar due to responsiveness, I can't easily find the documentation since the icon is hidden.

Sometimes I do not even know where to start checking so that I find the documentation.

image

The more info button has been surfacing this but we may need to expand the source of documentation links so that it can support urls that are outside of the set of sample queries.

We could change it to a documentation icon so that it only gives us information about documentation instead

**Question**
Since the documentation on docs already has the http requests in them, is there a way to reverse search the docs so that we can get the links that are related to the URL?

thewahome:
## Overview

Walks through the sample queries and resource explorer to get documentation links for the URL on the Query Input.

When available, changes the more info icon to be a documentation icon that allows a user to open the related link in a new tab

Closes #1854

## Testing Instructions

* Type in a URL
* Notice that the icon changes from disabled to enabled
* Click on the documentation icon
* A tooltip opens with the statement that documentation is available for the URL
* Click learn more and see the documentation on a new tab


Onokaev (approved):

ElinorW (approved):

ElinorW:
Seems like the methods on the collection panel changed to lowercase
image

ElinorW (approved):

Onokaev (approved):

PR #2226, opened 2022-11-09
Chore: remove class components

### LINKED ISSUE #2225: Use functional components instead of class components
Change the components from using class to function components. There are some files that still have class definitions. Changing them to functional components enables us to 

Using function components makes writing easier and reduces the amount of boilerplate code we need to write

thewahome:
## Overview

Fixes #2225 

### Notes

App.tsx may need another iteration


Onokaev (approved):
Looks good!

Issue #2228, opened 2022-11-10
Implement lazy loading on components


Onokaev:
**Is your feature request related to a problem? Please describe.**
GE page load time is a bit high for it's size. To reduce this, we can lazily load components that users don't see at first glance.
We can lazy load: Resources, History tab, Modify Permissions, Access tokens, Request headers, Response headers, Code snippets, Toolkit component and Adaptive cards tabs. 

Lazy loading is a technique in React that allows you to defer the loading of certain parts of your application until they are actually needed. With lazy loading, you can split your code into smaller chunks that can be loaded on demand, rather than loading everything at once.

When you use lazy loading in React, **the entire bundle is not downloaded from the server initially**. Instead, only the code that is required to render the initial view of your application is downloaded. Any components or modules that are marked for lazy loading will not be downloaded until they are actually needed.

When the user interacts with your application and triggers the loading of a lazily loaded component or module, a new HTTP request is sent to the server to download the required code. This code is then dynamically added to the running application, allowing the component or module to be rendered on the screen.

So, in summary, lazy loading in React allows you to split your code into smaller, more manageable chunks that can be loaded on demand, without requiring the entire bundle to be downloaded from the server at once.

**Describe the solution you'd like**
Graph explorer that loads faster

**Describe alternatives you've considered**

**Additional context**

adhiambovivian:
@Onokaev  what's the current page load avg time? And with this change, what's our projected page load time?

Onokaev:
Our current average is 12-15 seconds according to the load time stats. We can slash this to 5-7 seconds.

PR #2229, opened 2022-11-10
Task: Implement lazy loading

Onokaev:
## Overview
Implements #2228 

### Demo

image
Chunks before lazy loading

image

Chunks after lazy loading

### Notes
Currently, Graph Explorer is loaded in a single bundle. With the app growing, this bundle is also increasing in size. The current bundle has a size a little over 1MB. This has an impact on the app load time as the whole bundle has to be loaded first for the app to become usable

Lazy loading splits the original bundle into smaller chunks so that what is needed on first-load is what is loaded initially. The lazily loaded components will be in their own chunks and will be loaded on demand

## Testing Instructions

* How to test this PR
Load the app on your browser
Click 'Request headers' tab
Notice that it is loaded on demand
Click on other components (Modify Permissions, Access token, Response headers, Resources etc.)


src/app/views/query-response/headers/index.ts
```diff
@@ -1,2 +1,2 @@
 import ResponseHeaders from './ResponseHeaders';
-export { ResponseHeaders };
+export default ResponseHeaders;
```
thewahome, line 2: could `export { ResponseHeaders as default };`  also work?

src/app/views/query-response/pivot-items/pivot-items.tsx
```diff
@@ -1,5 +1,5 @@
-import { getTheme, IPivotItemProps, ITheme, PivotItem } from '@fluentui/react';
-import React from 'react';
+import { getTheme, IPivotItemProps, ITheme, PivotItem, Spinner, SpinnerSize } from '@fluentui/react';
```
thewahome, line 1: These are repeated heavily, we can have a `SuspenseLoader` component that we can use to wrap all the suspense content

thewahome (commented):
Let's have the PR body explain Please also mention in the PR description the gains expected. 
Are we aiming to reduce bundle size? How much does the work in this PR save us?

IMHO it could be that the network calls we are making on the first load are expensive. And that it could be unrelated to lazy loading the different tabs.

We can also analyse whether GE with the Dev portal chrome around it gives us the same load time as GE without the portal chrome. Then see whether it's the extra html added during production that affects the initial load time

config/webpack.config.js
```diff
@@ -414,7 +414,7 @@ module.exports = function (webpackEnv) {
     },
     plugins: [
       new webpack.optimize.LimitChunkCountPlugin({
-        maxChunks: 1
+        maxChunks: 100
```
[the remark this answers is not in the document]
Onokaev: Components that are lazy loaded are in their own chunks. The benefit of this is that the initial chunk that is loaded (with the components that a user sees for the first time) is small. So, it loads much faster. 
When you click on the Modify Permissions tab, it's chunk is loaded.

src/app/views/query-response/headers/index.ts
Onokaev: Yes. Let me change this

thewahome (changes requested):
Since some components are not immediately with the client when the page loads, they may fail to get downloaded eventually when needed due to various reasons. It would make sense to add an `ErrorBoundary` wrapped around the `SuspenseLoader`

thewahome (commented):
We are going to add a lot of SuspenseLoaders everywhere. 

It's probably possible for us to have a component registry where we register the suspended components and import them in the required files without adding the suspense and error boundaries.

If successful, it could mean we would be changing only where the components are imported from and not how they are imported. 

e.g.
`import { Permission } from 'component-registry';` instead of `const Permission = lazy(() => import('../../query-runner/request/permissions'))`

calebkiage:
You could also explore only lazy loading non-essential modules, then chunking and loading all essential modules. Multiple smaller files might load faster than a single large file due to parallel network downloads. However, the chunk size also plays a part. If the chunks are too small, you waste time in creating the connection before downloading.

src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
```diff
@@ -126,7 +126,7 @@ const AdaptiveCard = (props: any) => {
       }
       return (
         <Pivot className='adaptive-pivot'
-          onLinkClick={(pivotItem) => onPivotItemClick(sampleQuery, pivotItem)}
+          onLinkClick={(pivotItem: any) => onPivotItemClick(sampleQuery, pivotItem)}
```
gavinbarron, line 129: We shouldn't use any unless we have no other choice
```suggestion
          onLinkClick={(pivotItem: PivotItem) => onPivotItemClick(sampleQuery, pivotItem)}
```

gavinbarron (changes requested):

src/app/views/common/lazy-loader/component-registry/index.tsx
```diff
@@ -0,0 +1,133 @@
+/* eslint-disable max-len */
+import { lazy } from 'react';
+import { IPermissionProps } from '../../../../../types/permissions';
+import { SuspenseLoader } from '../suspense-loader/SuspenseLoader';
+
+const LazyPermission = lazy(() =>
+/* webpackChunkName: "permissions" */ import('../../../query-runner/request/permissions'));
```
gavinbarron, line 7: The webpackChunkName magic comment needs to be inside the import() function call to ensure that the chunk actually gets the name you want it to

src/app/views/common/lazy-loader/component-registry/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+const LazyStatusMessages = lazy( () =>
+/* webpackChunkName: "statusMessages" */ import('../../../app-sections/StatusMessages'));
```
gavinbarron, line 9:
```suggestion
    import(/* webpackChunkName: "statusMessages" */ '../../../app-sections/StatusMessages'));
```

gavinbarron (approved):
LGTM

src/app/views/query-runner/request/Request.tsx
```diff
@@ -74,7 +76,7 @@ const Request = (props: any) => {
         }}
       >
         <div style={containerStyle} id={'request-header-tab'}>
-          <RequestHeaders />
+          <RequstHeaders />
```
thewahome, line 79:
```suggestion
          <RequestHeaders />
```

src/app/views/query-runner/request/Request.tsx
```diff
@@ -88,7 +90,7 @@ const Request = (props: any) => {
         }}
       >
         <div style={containerStyle} id={'permission-tab'}>
-          <Permissions />
+          <Permission />
```
thewahome, line 93:
```suggestion
          <Permissions />
```

src/app/views/query-runner/request/Request.tsx
```diff
@@ -10,11 +14,9 @@ import { Mode } from '../../../../types/enums';
 import { setDimensions } from '../../../services/actions/dimensions-action-creator';
 import { translateMessage } from '../../../utils/translate-messages';
 import { convertPxToVh, convertVhToPx } from '../../common/dimensions/dimensions-adjustment';
-import { Auth } from './auth';
 import { RequestBody } from './body';
-import { RequestHeaders } from './headers';
-import { Permissions } from './permissions';
 import './request.scss';
+import { Permission, Auth, RequstHeaders } from '../../common/lazy-loader/component-registry';
```
thewahome, line 19:
```suggestion
import { Permission, Auth, RequestHeaders } from '../../common/lazy-loader/component-registry';
```

src/app/views/common/lazy-loader/component-registry/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+export const Auth = (props?: any) => {
+  return (
+    <SuspenseLoader>
+      <LazyAuth {...props} />
+    </SuspenseLoader>
+  )
+}
+
+export const RequstHeaders = (props?: any) => {
```
thewahome, line 101:
```suggestion
export const RequestHeaders = (props?: any) => {
```

thewahome (commented):
Noticed a few typos

thewahome (approved):

gavinbarron:
> Noticed a few typos

Nice spotting @thewahome 
I'd suggest that everyone uses the CSpell extension for VSCode, it really helps catch those kind of typos early.

thewahome:
@gavinbarron I'll add it as a recommended extension

PR #2239, opened 2022-11-16
Enhancement: sample queries cache

### LINKED ISSUE #2119: Refresh cached sample queries
In #2036 it was reported that the links weren't working. The cause was that the links in the cached sample queries that are built with the application had not been updated with the latest set.

These are shown when the DevX-API is unavailble for any reason.

Routinely hand over the samples so that the latest ones are built with the app for deployment.

**Additional context**
This could mean moving the sample queries from the DevX-API service since there would essentially no need to make a round trip to get the latest if the latest has already been built with the application.

thewahome:
## Overview

Fixes #2119 

### Notes

When a successful call is made to get samples, they are cached for use in case their next run experiences a failure.
This ensures the cache has the last known good set of samples.

The older version of cached sample queries is used as a fallback of the fallback. This will eventually be phased out so that we can remove it from the build.

## Testing Instructions

* Load Graph Explorer in the local environment
* Change the base url for the samples url so that it once the app loads, it receives an error 404
* Notice that a cached version exists, with all the previously received samples


src/app/services/actions/samples-action-creators.ts
```diff
@@ -55,7 +57,11 @@ export function fetchSamples() {
       const res = await response.json();
       return dispatch(fetchSamplesSuccess(res.sampleQueries));
     } catch (error) {
-      return dispatch(fetchSamplesError({ error }));
+      let cachedSamples = await samplesCache.readSamples();
+      if (cachedSamples.length === 0) {
```
Onokaev, line 61: Can we check if cachedSamples are available before accessing their .length property?
if(cachedSamples && cachedSamples.length === 0)

Onokaev (approved):

src/app/services/actions/samples-action-creators.ts
thewahome: It always returns a list. An empty list when there's nothing to show and an actual list when content exists

ElinorW (approved):

PR #2251, opened 2022-11-22
Chore: Code cleanup

### LINKED ISSUE #2224: Improve code quality from the Sonar Cloud stats
**Describe the bug**
There are currently multiple code smells, bugs, and security hotspots. Some of these can be solved to improve the code quality

**Screenshots**
image

ElinorW:
## Overview

Reduce code smells and bugs from sonar cloud analysis
Fixes #2224 

### Demo

### Notes

List of code smells addressed in this PR

- [x] Remove TODO comment
- [x] "Switch" statements should have at least 3 cases
- [x] Switch case should end with an unconditional break, continue, return or throw statement
- [x] Cases in switch statements have repeated code blocks 
- [x] Nested ternary operations
- [x] Loop counters being modified in the loop body
- [x] Use of deprecated methods such as substr, addListener, removeListener
- [x] React components must start with an uppercase.
- [x] Redundant assignments
- [x] A shorthand property defined after a longhand property will completely override the value defined in the longhand property making the longhand one useless.
- [x] Unused imports
- [x] Using `?` and `undefined` together as they both communicate the same thing

## Testing Instructions

* How to test this PR
* Prefer bulleted description


thewahome:
Not really important, but as you go along, which code smells have been closed? It would be a good learning experience for us to see what the problem was and how we've fixed it

ElinorW:
> Not really important, but as you go along, which code smells have been closed? It would be a good learning experience for us to see what the problem was and how you've fixed it. Maybe indexing them in the Notes section of the overview

Ah yes, this is great feedback.
I shall reference them in the notes section as requested

src/app/views/App.tsx
```diff
@@ -86,7 +86,7 @@ class App extends Component<IAppProps, IAppState> {
 
   public componentDidMount = async () => {
     this.displayToggleButton(this.mediaQueryList);
-    this.mediaQueryList.addListener(this.displayToggleButton);
+    this.mediaQueryList.addEventListener('click',this.displayToggleButton);
```
thewahome, line 89: Using devtools I can set browser to different phones. 
What I have noticed is that the app doesn't know that it's in mobile view so it does not respond appropriately.

I think the culprit could be setting the listener to listen to a 'click' event. I am not sure which one it should be.

thewahome (changes requested):

src/app/views/App.tsx
ElinorW: Ah, this is a great observation. 
Okay, let me look this up... 
There's probably a more appropriate listener

thewahome (approved):

Onokaev (approved):

PR #2254, opened 2022-11-23
Fix: MSA account jwt token

### LINKED ISSUE #1936: Getting token details for MSA accounts fails since it's not a valid jwt token
**Describe the bug**
When you are signed in with an MSA Account, the JWT token redirection to jwt.ms is not able to open up the token and provide the details.

**To Reproduce**
1. Sign in with an MSA account
2. Click on the Auth tab and confirm the presence of an access token
3. Click on the json icon
5. See that there is no decoded information

**Expected behavior**
1. Hidden button when the user has an MSA account
2. Messaging that the details will not be visible and why

thewahome:
## Overview

Fixes #1936

The access token for an MSA is not a JWT token. 

If they were to try and decode it using aka.ms/jwt they'd be be disappointed. 

Disabling it for these types of accounts


Onokaev (commented):
Nice work @thewahome. Can we add a toolkit that says why the button is disabled?
Maybe we can tell the users that the token they have is not a jwt token and cannot be decoded by jwt.ms

Onokaev (approved):
Ship it :) !

PR #2267, opened 2022-12-01
Fix: % symbol crash

### LINKED ISSUE #2265: Typing % in the query fields blanks out the screen
**Describe the bug**
If you type % in the query field the screen instantly blanks out. The console shows "Uncaught URIError: URI malformed" errors.

**To Reproduce**
1. Open Graph Explorer
2. Query field is by default set to "https://graph.microsoft.com/v1.0/me"
3. Type % after "me"

 - OS: Windows
 - Browser Edge, Pale Moon
 - Swedish Input keyboard

thewahome:
## Overview

Fixes #2265 

### Notes

Adding a % in a random location fails the decodeUrl function in the abnf validator.

### DIFF
--- a/src/modules/validation/abnf.spec.ts
+++ b/src/modules/validation/abnf.spec.ts
@@ -20,7 +20,8 @@ const validUrls = [
 
 const invalidUrls = [
   'https://graph.microsoft.com/me+you',
-  'https://graph.microsoft.com/v1.0/me/messages?$$select=id'
+  'https://graph.microsoft.com/v1.0/me/messages?$$select=id',
+  'https://graph.microsoft.com/v1.0/me/drive/root:/Encoded%'
 ];
 
 /*
--- a/src/modules/validation/abnf.spec.ts
+++ b/src/modules/validation/abnf.spec.ts
@@ -21,7 +21,8 @@ const validUrls = [
 const invalidUrls = [
   'https://graph.microsoft.com/me+you',
   'https://graph.microsoft.com/v1.0/me/messages?$$select=id',
-  'https://graph.microsoft.com/v1.0/me/drive/root:/Encoded%'
+  'https://graph.microsoft.com/v1.0/me/drive/root:/Encoded%',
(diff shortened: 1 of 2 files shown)


Onokaev (approved):

src/modules/validation/abnf.ts
```diff
@@ -38,10 +38,12 @@ export class ValidatedUrl {
   }
 
   public validate(graphUrl: string): ValidationResult {
+    let decodedGraphUrl = graphUrl;
```
gavinbarron, line 41: Can we please have some tests to validate that we're fixing the issues and not introducing new ones? Or at least one that validates that this fixes an issue?

Why is decodeURI being called? are there values in the encoded version of the string which might cause issues when fed into the parser here?

gavinbarron (changes requested):

src/modules/validation/abnf.ts
thewahome: Yes. The parser requires that the URL be passed in without encoding. Passing a lone % to the parser would cause it to fail the decode step. 

Let me add more tests

thewahome: @gavinbarron take a look

src/modules/validation/abnf.spec.ts
```diff
@@ -20,7 +20,8 @@ const validUrls = [
 
 const invalidUrls = [
   'https://graph.microsoft.com/me+you',
-  'https://graph.microsoft.com/v1.0/me/messages?$$select=id'
+  'https://graph.microsoft.com/v1.0/me/messages?$$select=id',
+  'https://graph.microsoft.com/v1.0/me/drive/root:/Encoded%'
```
gavinbarron, line 24: Encoded characters are a % followed by 2 hex digits
```suggestion
  'https://graph.microsoft.com/v1.0/me/drive/root:/Encoded%',
  'https://graph.microsoft.com/v1.0/me/drive/root:/Encoded%2'
```

PR #2288, opened 2022-12-05
Fix: Adaptive cards color on dark mode

### LINKED ISSUE #2018: Adaptive card colors on dark theme are hard to see
Some of the adaptive cards on dark theme are really hard to see or read. 
Example: https://graph.microsoft.com/v1.0/sites/root 
adaptive cards on dark theme

Onokaev:
## Overview
Fixes #2018 

### Demo
image

### Notes
This problem is only on production. We'll have to deploy to the staging environment with the header and footer injected to confirm that the fix works

## Testing Instructions

* How to test this PR
Run the query https://graph.microsoft.com/v1.0/sites/root
Click on `Adaptive Cards` tab
See fix

### DIFF
--- a/src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
+++ b/src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
@@ -93,6 +93,7 @@ const AdaptiveCard = (props: any) => {
       const renderedCard = adaptiveCard.render();
       if(renderedCard){
         renderedCard.style.backgroundColor = currentTheme.palette.whiteTranslucent40;
+        renderedCard.style.color = 'black';
       }
       const handleCopy = async () => {
         trackedGenericCopy(JSON.stringify(data.template, null, 4),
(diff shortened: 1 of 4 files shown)


src/app/views/query-response/adaptive-cards/AdaptiveCard.tsx
```diff
@@ -91,6 +91,10 @@ const AdaptiveCard = (props: any) => {
     try {
       adaptiveCard.parse(data.card);
       const renderedCard = adaptiveCard.render();
+      if(renderedCard){
+        renderedCard.style.backgroundColor = currentTheme.palette.whiteTranslucent40;
+        renderedCard.style.color = 'black';
```
thewahome, line 96: Shouldn't this too get it's colour from the currentTheme palette?

Onokaev: Yes. @ElinorW pointed this out and I already have a fix out

thewahome (approved):

PR #2348, opened 2023-01-16
Fix: enter key runs previous query

### LINKED ISSUE #2338: ENTER keyboard shortcut to run the current query is broken
**Describe the bug**
Pressing ENTER instead of clicking Run Query will run the previous query instead of the current one.

**To Reproduce**
1. Go to https://aka.ms/ge
2. change `/me` in the url to `/users`
3. remain with the cursor focus on the textbox, and press enter

**Expected behavior**
`/users` response should execute

**Actual behavior**
`/me` response is executed (previous request)

Note: pressing the "Run Query" button will execute the current request.

thewahome:
## Overview

Fixes #2338 

### Notes

It fails for the use case where a user types in something and it is the only thing in the suggestions. Since we didn't assume this becomes the selected suggestion, the query doesn't get changed.

It is fixed by appending the suggestion to the URL

## Testing Instructions

* Run `/me`
* Change `me` to `users`
* Hit enter
* Notice the results for `/users` in the response


src/app/views/query-runner/query-input/auto-complete/AutoComplete.tsx
```diff
@@ -58,6 +58,7 @@ const AutoComplete = (props: IAutoCompleteProps) => {
     const targetValue = e.target.value;
     setQueryUrl(targetValue);
     props.contentChanged(targetValue);
+    console.log(targetValue)
```
Onokaev, line 61: Must have forgotten this :)

thewahome: Nice catch

Onokaev (approved):

ElinorW (approved):

Issue #2344, opened 2023-01-13
Investigate the cause of script error reported on PowerBI


Onokaev:
We have an influx of script error, possibly coming from the Graph Explorer Try-It experience.

PR #2351, opened 2023-01-18
Task: Add logging for script errors

Onokaev:
## Overview
Adds telemetry for #2344 

### Notes
This PR adds telemetry for all script errors. We will be able to tell where exactly the errors are coming from. 

## Testing Instructions

* How to test this PR
1. On the .tsx file, add:
image
2. Now open the network tab and see the errors tracked. Notice that the exception data contains a source object that shows where the error is originating from:
3. 
image


millicentachieng (changes requested):
@Onokaev Could you please rename the properties to match what we are already processing in the pipelines?   
image

`lineno` -> `lineNumber`
`colno` -> `columnNumber`
`source` -> `errorSrc`

millicentachieng (approved):

PR #2364, opened 2023-01-30
Fix: Display beta resources subgroups for admin group

### LINKED ISSUE #2356: Resource Explorer and auto complete is not showing some beta endpoints
image

The admin node should show "edge", "windows", "sharepoint" as child nodes.

MaryannGitonga:
## Overview

Closes #2356 

### Demo

image

### Notes

## Testing Instructions

* Checkout to this branch.
* Load local instance of GE.
* Navigate to the Resources Explorer tab
* Toggle the **Switch to beta** toggle button
* Observe groups under the **admin** group

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,6 @@
                 "test-ui": "npx playwright test src/tests/ui/",
                 "test-playwright-accessibility": "start-server-and-test start-server http://localhost:3000 test-accessibility",
                 "test-accessibility": "npx playwright test src/tests/accessibility"
-
         },
         "eslintConfig": {
                 "extends": "react-app"
--- a/src/app/views/sidebar/resource-explorer/ResourceExplorer.tsx
+++ b/src/app/views/sidebar/resource-explorer/ResourceExplorer.tsx
@@ -36,17 +36,19 @@ const UnstyledResourceExplorer = (props: any) => {
   const { resources } = useAppSelector(
     (state) => state
   );
+
   const classes = classNames(props);
   const { data, pending, paths: selectedLinks } = resources;
-
   const versions: any[] = [
     { key: 'v1.0', text: 'v1.0' },
     { key: 'beta', text: 'beta' }
   ];
 
+  const resourcesToUse = JSON.parse(JSON.stringify(data.children));
+
   const [version, setVersion] = useState(versions[0].key);
   const [searchText, setSearchText] = useState<string>('');
-  const filteredPayload = getResourcesSupportedByVersion(data.children, version, searchText);
+  const filteredPayload = getResourcesSupportedByVersion(resourcesToUse, version, searchText);
   const navigationGroup = createResourcesList(filteredPayload, version, searchText);
 
   const [resourceItems, setResourceItems] = useState<IResource[]>(filteredPayload);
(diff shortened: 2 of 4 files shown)


Onokaev (changes requested):
Nice catch. Could we also update the fallback resources.json file. I noticed that it doesn't have the new beta nodes that prompted for this fix.
Get a new copy here:https://graphexplorerapi.azurewebsites.net/swagger/index.html on the openapi/tree endpoint

Onokaev (dismissed):
Looks good!

thewahome (dismissed):

ElinorW (dismissed):

darrelmiller:
@thewahome @gavinbarron Why is the resources tree being populated from a static JSON file?  This information should come from DevX API.  Microsoft Graph is updated weekly.  We cannot wait until the next Graph Explorer release to get the latest version of the available resources.  We also can't add an 8MB payload to every user of Graph Explorer regardless of whether they look at Resource Explorer or not.

MaryannGitonga:
@darrelmiller, the update in the resources.json file is for the cached resources, in case of no connection and GE cannot fetch the resources from DevX API. The resources in the file were outdated hence why I did an update. In normal cases, the resources should be fetched from the API. And that works fine now. 

I am currently looking into a periodic comparison that can be done on the resources.json to check if it's outdated and if that is the case, it is updated with the latest resources. It could be resource-intensive hence why I think it would be a better idea if it were after a significant period of time (eg. a few days). Feel free to share your thoughts on this.

I hope I have answered your question.

darrelmiller:
Hi @MaryannGitonga Thanks for the clarification.  A few things.  I don't believe the caching is working as intended.  I just went to Graph Explorer, called Microsoft Graph, showed the permissions tab, which returned results from DevX API and still the resource explorer does not show the additional endpoints under admin.

image

Secondly, I don't believe that packaging a cached copy of the resources.json as part of the GE deployment is an effective way to handle DevX API outages.  If a customer has no connection, then they will not be able to use Graph Explorer to make calls, so the value of having the resources tab is severely limited.  If DevX API is down then I would be perfectly fine in degrading the experience to indicate that the resources tree is not currently available.  If we choose to use caching to provide some kind of offline mode, then we should leverage one of the many caches that exist in the browser, we should not be inventing our own.  

Adding resources.json to the GE component would significantly increase its download size for every customer, regardless of whether they used the resources tab or not.  That doesn't seem customer friendly to me.

I believe this is a similar solution to what we do for samples today.  For the same reasons, I don't think we should continue with that approach.

MaryannGitonga:
@darrelmiller, this is well noted. Thanks.

thewahome:
> @thewahome @gavinbarron Why is the resources tree being populated from a static JSON file? This information should come from DevX API. Microsoft Graph is updated weekly. We cannot wait until the next Graph Explorer release to get the latest version of the available resources. We also can't add an 8MB payload to every user of Graph Explorer regardless of whether they look at Resource Explorer or not.

@darrelmiller, GE fully depends on the DevX API to get the resources. The JSON was initially used to have a local copy of it so that we can have a cached version to show in case of inconsistencies with the DevX API. 

I understand what you are saying about caching. We have a better mechanism when handling the sample queries which we will use here and remove the local JSON files.

gavinbarron:
@darrelmiller I agree that we should be using DevX API as a primary source, anything else is unacceptable.

Having a fallback to use should we not get a response from DevX API is a good idea in my opinion to prevent providing a poor user experience. We can and should be loading that resource using an awaited `import()` function, thus the weight is only placed on the browsing experience when DevX API has failed to respond. 

Now if we were to leverage some of the browser storage APIs and cache the response from DevX API indefinitely but keep a refresh period to update the cache I'd be totally comfortable with dropping the static JSON file altogether and providing a degraded experience if the user has never loaded GE on that browser and DevX API fails to respond.

PR #2367, opened 2023-01-31
Fix: Inform user if URL is incomplete

### LINKED ISSUE #2359: Graph explorer is prepending https://graph.microsoft.com/ for queries that do not include https://graph.microsoft.com/
**Describe the bug**
Graph explorer is prepending https://graph.microsoft.com/v1.0/ for queries that do not include https://graph.microsoft.com/v1.0/
image
Typing /me runs and returns a result

**To Reproduce**
2. Delete the entire query on the query textbox and type `me`
3. Click `Run query' button
4. Notice that the query runs and returns a result

**Expected behavior**
Graph explorer should raise an error in case of invalid URLs.

**Screenshots**

**Additional context**

Onokaev:
## Overview
Closes #2359 

### Demo
image

### Notes
The ABNF validator only validates URLs that have the http/https protocol appended. We therefore need extra validation when the abnf validator is not returning an error. 

## Testing Instructions

* How to test this PR
Backspace on the query URL until there's only http left
See fix

### DIFF
--- a/src/app/views/query-runner/query-input/auto-complete/auto-complete.util.ts
+++ b/src/app/views/query-runner/query-input/auto-complete/auto-complete.util.ts
@@ -51,6 +51,9 @@ function getErrorMessage(queryUrl: string) {
   if (error) {
     return `${translateMessage('Possible error found in URL near')}: ${error}`;
   }
+  if (queryUrl.indexOf('https://graph.microsoft.com/') === -1){
+    return translateMessage('The URL must start with https://graph.microsoft.com/');
+  }
   return '';
 }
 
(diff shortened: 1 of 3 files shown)


src/app/views/query-runner/query-input/auto-complete/auto-complete.util.ts
```diff
@@ -51,6 +51,9 @@ function getErrorMessage(queryUrl: string) {
   if (error) {
     return `${translateMessage('Possible error found in URL near')}: ${error}`;
   }
+  if (queryUrl.indexOf('https://graph.microsoft.com/') === -1){
+    return translateMessage('The URL must start with https://graph.microsoft.com/');
```
ElinorW, line 55: This error isn't appropriate because canary endpoint users use a modified URL that doesn't necessarily look like this

ElinorW (changes requested):

src/app/views/query-runner/query-input/auto-complete/auto-complete.util.ts
Onokaev: Yes. That's right. We should only validate the protocol

ElinorW (approved):

PR #2373, opened 2023-02-01
Task: Add browser storage for resources

### LINKED ISSUE #2369: Add in-browser caching for resources

In #2364 there is a suggestion to shift from having a cached copy of the resources as a way of handling DevX API outages. This approach significantly increases the download size of the GE component for every customer, regardless of whether they used the resources tab or not. This demotes the overall offline experience.

**Describe the solution you'd like:**

- If DevX API is down and the first fetch is unsuccessful, GE should indicate that the resources tree is currently unavailable.
- If the first fetch is successful, the resources should be stored the browser storage, and this can be used as a fallback when fetches in DevX API are unsuccessful.

MaryannGitonga:
## Overview

Closes #2369 
### Demo
- When the first fetch is unsuccessful, and the storage instance is empty, this is displayed to the user:
image

- Otherwise, the resources are displayed (served from DevX API or the browser storage when DevX API is down).

### Notes

## Testing Instructions

* Checkout to this branch.
* Run the local instance of GE.
* Navigate to the Resources tab.
* Observe changes.


thewahome:
We should remember to remove the local json files

src/app/services/actions/resource-explorer-action-creators.ts
```diff
@@ -44,25 +45,29 @@ export function removeResourcePaths(response: object): AppAction {
...
 
+    try {
       const response = await fetch(resourcesUrl, options);
       if (response.ok) {
         const resources = await response.json() as IResource;
         return dispatch(fetchResourcesSuccess(resources));
       }
       throw response;
     } catch (error) {
+      const cachedResources = await resourcesCache.readResources();
```
gavinbarron, line 67: IMO we should be using the cached data in preference to hitting the DevX API every time, it will be much quicker.

src/app/views/sidebar/resource-explorer/ResourceExplorer.tsx
```diff
@@ -183,7 +184,7 @@ const UnstyledResourceExplorer = (props: any) => {
 
   return (
     <section style={{ marginTop: '8px' }}>
-      {!isolated && <>
+      {!isolated && items[0].links.length > 0 && <>
```
gavinbarron, line 187: Is items always populated?

src/app/views/sidebar/resource-explorer/panels/postman.util.spec.ts
```diff
@@ -1,9 +1,9 @@
 import { IResource } from '../../../../../types/resources';
-import content from '../../../../utils/resources/resources.json';
 import { createResourcesList, getResourcePaths } from '../resource-explorer.utils';
 import { generatePostmanCollection } from './postman.util';
+import content from '../resources.json'
 
-const resource = JSON.parse(JSON.stringify(content)) as IResource;
+const resource = JSON.parse(JSON.stringify(content)) as IResource
```
gavinbarron, line 6:
```suggestion
const resource = JSON.parse(JSON.stringify(content)) as IResource;
```

gavinbarron (commented):
Using a cached data ONLY as a fallback is a missed opportunity to reduce COGS and improve speed.

src/app/services/actions/resource-explorer-action-creators.ts
thewahome: We only call the api once when the page loads. And there are no use cases where for the app to work a reload is required. 
I'm not sure that there'd be a use case for the preference

src/app/views/sidebar/resource-explorer/ResourceExplorer.tsx
```diff
@@ -226,11 +227,13 @@ const UnstyledResourceExplorer = (props: any) => {
         </>
       }
 
-      <Label styles={{ root: { position: 'relative', left: '10px' } }}>
+      { items[0].links.length > 0 && <Label styles={{ root: { position: 'relative', left: '10px' } }}>
```
Onokaev, line 230: I think we should check if items and their content are defined before accessing them
`items && items.length > 0 && items[0].links  && items[0].links.length > 0`?

src/app/views/sidebar/resource-explorer/ResourceExplorer.tsx
MaryannGitonga: image
This is what the condition is trying to preempt from rendering: the search box and the switch to beta toggle button.

src/app/services/actions/resource-explorer-action-creators.ts
gavinbarron: Our users never hit F5? or visit multiple times a day?

thewahome: that could happen. 
To prefer cache, we'd need to make it able to expire so that we can sometimes get the latest.

gavinbarron: I've been thinking on this a bit and chatting with @darrelmiller 
We should be looking to ensure that the DevX API provides appropriate cache control headers and not worrying about adding a complex caching mechanism here.

Honestly the less code we write the better in terms of cost of ownership, and it allows us to focus on other areas. 

Do we know how frequently the data in the DevX API is updated?

thewahome: We had a chat with @darrelmiller and we uncovered something I consider a large spanner in the works. The cache control headers  do not work in instances where the client and the api are not in the same domain. 
An alternative exists that we use service workers to try this. However I am not sure whether the code changes will be any less than the minimal work we have done now.

thewahome: @gavinbarron, @darrelmiller  I am more inclined to :
- following your suggesion to prefer cache.
- coming up with a mechanism to expire cache. For resources, we could take advantage of knowledge that the openapi document is refreshed every week. Or simply set cache to expire after 7 days 

Tell me what you think.

gavinbarron: >The cache control headers do not work in instances where the client and the api are not in the same domain.

O_o

I'm really surprised to hear this because this is literally the situation for resources served from a CDN, those are always cross domain and cache in the browser based on cache control headers.

Is there a way I can take a look at what you're seeing to validate this?

thewahome: @darrelmiller found this when searching. It was a shared screen where we were able to see the results of the query together.

I've been looking this up and it looks like it could be possible even when the client and the API are not in the same domain as long as the client origin is allowed in the API.

gavinbarron: Yeah, we should be able to set the allowed headers in the response to the options calls from the server:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers

Onokaev (dismissed):

thewahome (approved):

PR #2419, opened 2023-02-20
Task: Add copy telemetry

### LINKED ISSUE #2411: [Telemetry] capture copy events
details to be shared with the team

Onokaev:
## Overview
Closes #2411 

### Demo
image

### Notes

## Testing Instructions
Use the keyboard to copy any content on Access tokens, Response preview, Response headers, Code snippets, Adaptive cards and Response body tabs.
Open the network tab and inspect the track call
Notice that the keyboard event is available with the property `trgger` to differentiate it from other copy events


thewahome (dismissed):

src/app/views/common/copy/KeyboardCopyEvent.ts
```diff
@@ -0,0 +1,40 @@
+import { eventTypes, telemetry } from '../../../../telemetry';
+import { KEYBOARD_COPY_TABS } from '../../../../telemetry/component-names';
+
+interface IComponentList {
+  [key: string]: string;
+}
+export const KeyboardCopyEvent = () => {
+  const componentList: IComponentList = KEYBOARD_COPY_TABS;
+  document.addEventListener('keydown', (event: KeyboardEvent) => {
+    if (event && (event.ctrlKey || event.metaKey) && event.key === 'c') {
```
ElinorW, line 10: Assigning 'c' here limits it to cases where a user has their Caps Lock off... 
We could use the key code/key value of 'c', which is 67, to include all cases...

src/app/views/common/copy/KeyboardCopyEvent.ts
```diff
@@ -0,0 +1,40 @@
...
+  const getComponentName = (targets: EventTarget[]): string => {
+    const targetIds = targets.map((target: EventTarget) => {
+      return getTargetId(target);
+    });
+    const filteredTargetIds = targetIds.filter((value) => value !== null)!;
+    const componentName = Object.keys(componentList).find(key => filteredTargetIds.includes(key));
+    return componentName || '';
+  }
+
+  const getTargetId = (target: EventTarget) => {
```
ElinorW, line 26: Instead of id's could we use the itemKey property? 
The other telemetry events use itemKey and I think it's better to maintain consistency... 
I can see a potential issue where one telemetry has the id as `Javascript tab` and another will have `JavaScript tab` which will appear as two different components, yet they are the same component

ElinorW (changes requested):

src/app/views/common/copy/KeyboardCopyEvent.ts
Onokaev: We can't use the itemKey property because this is specific to clicks on pivot items. The global event does not know about this property, so the only other option is to use the target IDs. 

Nice catch. Let me change the spelling

src/app/views/common/copy/KeyboardCopyEvent.ts
Onokaev: The event.keyCode property was deprecated. We can use event.key.toLowerCase() to cover both cases where Caps Lock is on or off
image

ElinorW: Aah okay perfect!

src/app/views/common/copy/KeyboardCopyEvent.ts
ElinorW: Alright awesome

ElinorW (approved):

PR #2428, opened 2023-02-28
Feature: anonymous experience search queries

### LINKED ISSUE #2230: When using Graph Explorer without a logged in user, I should be able to use the search endpoints
**Describe the bug**
When using Graph Explorer without a logged in user, I should be able to use the search endpoints. I understand these are POST requests but don't impact the tenant, so I would believe they should allow developers to test these queries out.

**To Reproduce**
https://developer.microsoft.com/en-us/graph/graph-explorer?request=search%2Fquery&method=POST&version=v1.0&GraphUrl=https://graph.microsoft.com&requestBody=eyJyZXF1ZXN0cyI6W3siZW50aXR5VHlwZXMiOlsibWVzc2FnZSJdLCJxdWVyeSI6eyJxdWVyeVN0cmluZyI6ImNvbnRvc28ifX1dfQ==&headers=W3sibmFtZSI6IkNvbnRlbnQtVHlwZSIsInZhbHVlIjoiYXBwbGljYXRpb24vanNvbiJ9XQ==

**Expected behavior**
I can use the sample queries and my own on the search endpoints.

**Screenshots**
image

thewahome:
## Overview

Closes #2230

Allows a user in the anonymous experience to be able to make a POST request tp endpoints in the sample queries `search` category

### DIFF
--- a/src/app/views/sidebar/sample-queries/sample-query-utils.ts
+++ b/src/app/views/sidebar/sample-queries/sample-query-utils.ts
@@ -4,7 +4,7 @@ import { GRAPH_URL } from '../../../services/graph-constants';
 import { validateExternalLink } from '../../../utils/external-link-validation';
 import { sanitizeQueryUrl } from '../../../utils/query-url-sanitization';
 
-export function isJsonString(str: string): boolean {
+function isJsonString(str: string): boolean {
   try {
     JSON.parse(str);
     return true;
@@ -13,7 +13,7 @@ export function isJsonString(str: string): boolean {
   }
 }
 
-export function performSearch(queries: ISampleQuery[], value: string): ISampleQuery[] {
+function performSearch(queries: ISampleQuery[], value: string): ISampleQuery[] {
   const keyword = value.toLowerCase();
   return queries.filter((sample: any) => {
     const name = sample.humanName.toLowerCase();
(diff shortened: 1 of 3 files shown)


src/app/views/sidebar/sample-queries/sample-query-utils.ts
```diff
@@ -48,3 +48,23 @@ export const trackDocumentLinkClickedEvent = async (item: ISampleQuery): Promise
...
   validateExternalLink(item.docLink || '', componentNames.DOCUMENTATION_LINK, item.id);
 }
+
+interface ShouldRunQueryArgs {
+  method: string;
+  url: string;
+  authenticated: boolean;
+}
+
+const shouldRunQuery = ({ method, url, authenticated }: ShouldRunQueryArgs): boolean => {
```
Onokaev, line 58: This method should return true if we should run a query. Making it return false will be confusing for future maintenance.
In QueryInput.tsx, we should have something like this `const showError = !shouldRunQuery({
    method: sampleQuery.selectedVerb, authenticated,
    url: sampleQuery.sampleUrl
  });`
This means we should not show the error when shouldRunQuery returns true. 

Also, in SampleQueries.tsx:
` if (!shouldRunQuery({ method: query.method, authenticated: tokenPresent, url: query.requestUrl })) {
        selectionDisabled = true;
      }`
This means the selection is disabled only when we should not run the query.
What do you think?

thewahome: makes sense...

thewahome: This makes sense. Thanks for catching it @Onokaev and @ElinorW

ElinorW (approved):

PR #2445, opened 2023-03-08
Fix: Permissions styling

Onokaev:
## Overview
Fixes #2440 

### Demo

[screen-recorder-wed-mar-08-2023-12-27-47.webm](https://user-images.githubusercontent.com/45680252/223727344-e1325244-96c5-4769-adf9-1c62dcdb4d97.webm)

[screen-recorder-wed-mar-08-2023-16-32-28.webm](https://user-images.githubusercontent.com/45680252/223727359-3a7750de-33f1-497c-acbd-bb1db60acd87.webm)

### Notes

Optional. Ancillary topics, caveats, alternative strategies that didn't work out, anything else.

## Testing Instructions

* How to test this PR
Click on the Modify Permissions tab
Adjust your browser size and confirm that all items are aligned as per requirement
Sign in and out while interacting witht the tab

PR #2453, opened 2023-03-13
Fix: Consent type flipping

### LINKED ISSUE #2438: Permissions - consenting to a permission on the list would switch the consent button on the previous permission
On the modify permissions tab, sometimes when I consent to a permission , the consent button turns gray to unconsent. 
If I consent to the permission below on the list, the button on the previous permission flips back to blue (consent)

Onokaev:
## Overview
Closes #2438

### Demo

### Notes
Propagation of consented permissions takes time. As a result, subsequent calls to the token endpoint may not immediately return an already consented scope. This issue is however intermittent. 

## Testing Instructions

* How to test this PR
Sign in with the staging link
Consent to permissions
Notice that the consent button no longer flips


src/app/views/query-runner/request/permissions/Permission.tsx
```diff
@@ -181,29 +185,29 @@ export const Permission = (permissionProps?: IPermissionProps): JSX.Element => {
...
+    return <div />
   }
 
-  const permissionConsentTypeLabel = (permissionInAllPrincipal : boolean) : JSX.Element => {
-    if(permissionInAllPrincipal){
+  const permissionConsentTypeLabel = (permissionInAllPrincipal: boolean): JSX.Element => {
+    if (permissionInAllPrincipal) {
       return (
-        <div style={{textAlign: 'left', paddingLeft: '10px'}}>
+        <div style={{ textAlign: 'left', paddingLeft: '10px' }}>
```
calebkiage, line 203: Maybe use `textAlign: 'start` in case the user is working with RTL languages like Arabic.

calebkiage: It may also be better to have the padding in a style sheet so changing it is easier.

calebkiage (commented):
Left some CSS related comments.

src/app/services/actions/permissions-action-creator.ts
```diff
@@ -193,6 +194,7 @@ export function consentToScopes(scopes: string[]) {
           }))
       }
     } catch (error: any) {
+      console.log('Here is the error', error);
```
thewahome, line 197: Nit: This was forgotten

src/app/services/actions/permissions-action-creator.ts
```diff
@@ -207,6 +209,18 @@ export function consentToScopes(scopes: string[]) {
   };
 }
 
+const correctConsentedScopes = (scopeToBeConsented: string[], consentedScopes: string[],
```
thewahome, line 212: Naming: the function name is unclear. Does it correct the consented scopes or return a set of consented scopes?

src/app/views/query-runner/request/permissions/PanelList.tsx
```diff
@@ -49,7 +49,9 @@ const PanelList = ({ messages,
   const { permissionPanelStyles } = profileStyles(theme);
 
   useEffect(() => {
-    setPermissions(sortPermissions(fullPermissions));
+    if(searchValue === '' && groups.length === 0){
```
thewahome, line 52: This could also be null checked like so
```suggestion
    if(!searchValue && groups.length === 0){
```

src/app/views/query-runner/request/permissions/Permission.tsx
```diff
@@ -181,29 +185,29 @@ export const Permission = (permissionProps?: IPermissionProps): JSX.Element => {
   }
 
   const consentTypeProperty = (consented: boolean, item: any): JSX.Element => {
```
thewahome, line 187: We could probably be clearer by saying what the function does. Does it get/create the consent type property?

src/app/services/actions/permissions-action-creator.ts
Onokaev: It validates the consented scopes. How about 
const validatedScopes = validateConsentedScopes()
The function checks whether all previously consented scopes are available.

src/app/views/query-runner/request/permissions/Permission.tsx
Onokaev: It should be getConsentTypeProperty

src/app/views/query-runner/request/permissions/Permission.tsx
Onokaev: The styles have been addressed in a separate PR that was addressing styling on the Modify Permissions tab: https://github.com/microsoftgraph/microsoft-graph-explorer-v4/pull/2445

src/app/views/query-runner/request/permissions/Permission.tsx
Onokaev: I've also renamed the other column item headers with proper action verbs

thewahome (approved):

PR #2465, opened 2023-03-23
Fix: Set consent-type property

### LINKED ISSUE #2439: Permissions - consent type not showing up
Step 1: On the modify permissions tab, consent type is blank 
consent type blank

Expected behavior: I'd expect to be able to see the consent type. 

Step2: open permissions under the profile. Some permissions have the consent type but not all 
consent type profile

Step3: Close the permissions modal and suddenly I can see consent type on the modify tab screen 
consent type appearing

Onokaev:
## Overview
Fixes #2439 

### Demo
image
A spinner becomes available when the consent-type property is being fetched

When the fetch fails, a retry button is available to re-fetch the consent-type property
image

### Notes
The API call that returns the grants of the tenant takes time to return an updated grant after granting consent. For this reason, we have to poll the endpoint until we have an update. 

**The identity endpoints read data from eventually consistent datastores. This means that after updating a field, it takes time for all nodes to achieve the same state(eventual consistency). This means that calling the endpoint for updated data might still return stale data even after a write. 
For this reason, we have to keep polling the endpoint until we get the response that we expect. This is why we have a loop that continuously calls the grant endpoint until we have an updated grant.**

## Testing Instructions

* How to test this PR
Sign in with your account
Notice that all consent-type properties are available for consented permissions
Consent to a permission and notice the spinner just before the consent-type property is updated
Click on the retry button to re-fetch the consent-type property


src/app/views/query-runner/request/permissions/ConsentType.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    tenantGrantFetchPending ?
+      (<Spinner></Spinner>) :
+      (
+        <TooltipHost
+          content={translateMessage('Reload consent-type')}
+          directionalHint={DirectionalHint.leftCenter}
+        >
+          <IconButton onClick={handleOnClick}
```
ElinorW, line 56: Let's have an aria-label added to this.

thewahome (approved):

PR #2471, opened 2023-03-27
Update azure-pipelines.yml for Azure Pipelines

Onokaev:


Onokaev:
Closing this as the playwright tests we have already check for WCAG violations.

PR #2500, opened 2023-04-05
Fix dollar sign addition on autocomplete for chained query parameters

### LINKED ISSUE #2477: Autocomplete doesn't add `$` character
**Describe the bug**
Autocomplete doesn't add `$` character.

**To Reproduce**
1. Open Graph Explorer
2. Type URL `https://graph.microsoft.com/v1.0/users`
3. Type `?` after `https://graph.microsoft.com/v1.0/users` and when combobox is displayed then click on `$select`
4. `$select=` is added to URL
5. select `id`, so the URL is `https://graph.microsoft.com/v1.0/users?$select=id`
6. Type `&` and when combobox is displayed then click on any item from combobox, for example on `$filter`
7. The item from the combobox is added to URL but without `$` => `https://graph.microsoft.com/v1.0/users?$select=id&filter=`

**Expected behavior**
The query parameter selected from the combobox is added with `$` character `https://graph.microsoft.com/v1.0/users?$select=id&$filter=`

**Screenshots**

image

 - OS: Windows 11
 - Browser Edge
 - Version 111.0.1661.62 (Official build) (64-bit)

Onokaev:
## Overview
Fixes #2477 

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/src/tests/ui/anonymous-experiences/request.spec.ts
+++ b/src/tests/ui/anonymous-experiences/request.spec.ts
@@ -129,6 +129,39 @@ test.describe('Run query', () => {
     expect(page.getByText('"Microsoft Viva"')).toBeDefined();
   })
 
+  test('Tests query parameter addition for chained query parameters', async () => {
+    const queryInputField = page.locator('[aria-label="Query sample input"]');
+    await queryInputField.click();
+    await page.evaluate(() => document.fonts.ready);
+    await queryInputField.fill('');
+    await queryInputField.fill('https://graph.microsoft.com/v1.0/user');
+    await queryInputField.press('Tab');
+    await queryInputField.fill('https://graph.microsoft.com/v1.0/users?sel');
+    await queryInputField.press('Tab');
+    await queryInputField.press('Tab');
+    await queryInputField.fill('https://graph.microsoft.com/v1.0/users?$select=id&sel');
+    await queryInputField.press('Tab');
+    await queryInputField.fill('https://graph.microsoft.com/v1.0/users?$select=id&$select=d');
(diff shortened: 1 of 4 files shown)


src/tests/ui/anonymous-experiences/request.spec.ts
```diff
@@ -129,6 +129,39 @@ test.describe('Run query', () => {
...
+    await queryInputField.fill('https://graph.microsoft.com/v1.0/users?sel');
+    await queryInputField.press('Tab');
+    await queryInputField.press('Tab');
+    await queryInputField.fill('https://graph.microsoft.com/v1.0/users?$select=id&sel');
+    await queryInputField.press('Tab');
+    await queryInputField.fill('https://graph.microsoft.com/v1.0/users?$select=id&$select=d');
+    await queryInputField.press('Tab');
+    await queryInputField.press('Tab');
+    const queryContent = await page.locator('[aria-label="Query sample input"]').inputValue();
+    expect(queryContent).toBe('https://graph.microsoft.com/v1.0/users?$select=id&$select=deletedDateTime');
```
thewahome, line 148: We can improve this here by making sure we:
- can do a chained $ to different parameters e.g $select=x&$filter=y
- can not do a chained $ to the same parameter e.g $select=x&$select=u

Onokaev: True. Chaining the same parameter is wrong. Let me make adjustments to cover this case

ElinorW (approved):

PR #2514, opened 2023-04-17
Chore: popups manager

### LINKED ISSUE #2452: Create a popup manager to power dialogs and modals
Managing popup state from different locations is a problem that has been faced by contributors.

> One aspect that surfaced is that graph explorer is mixed between state/props and redux which sometimes makes inter-components communication cumbersome or leads to race conditions. I had to work my way around to get the panel to open from the header bar. (in case you're considering refactoring things in the future). @baywet [in this comment](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/pull/2189#issuecomment-1314259975) 

Create a centralized popup state manager that can be called from any component and has the power to show a popup (dialog/modal) at will.
It should have a simple to use contract like:
1. initialisation
`const { open: openChangeTheme } = usePopup("change-theme", 'modal');`
2. usage
```
openChangeTheme({
  settings: { title: 'Change theme' }
});
```

thewahome:
## Overview

Closes #2452

### Demo

In the component that you want to load another component as a dialog / modal / panel, simply:

import it using the usePopups hook
  `const { open: openPermissions } = usePopups('full-permissions', 'panel');`

Then call it by passing in parameters
```
const openPermissionsPanel = () => {
    openPermissions({
      settings: {
        title: translateMessage('Permissions'),
        width: 'lg'
      }
    })
  }
  ```

### Notes

Parameters in the functions are autocompleted e.g the component name and the type of display. The width in the settings

### CODE REVIEW

**Review by ElinorW — COMMENTED** [review]
Seems like the "Download Postman Collection" button has moved to the left. 
image

The rest all work as expected 👍🏾

**Comment by thewahome** [comment]
Yes @ElinorW ... the button moved to where Fluent Defines Dialog Footer buttons to go. I didn't see there being a need to move it

File: src/app/services/context/popups-context/PopupsContext.tsx [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+
+import { PopupsComponent, PopupsProps, PopupsStatus, PopupsType } from '.';
+
+export interface Popup<Data = {}> {
+  component: React.ElementType<PopupsComponent<Data>>,
+  popupsProps: PopupsProps<Data>,
+  type: PopupsType,
+  id: string;
+  result?: any;
+  open?: boolean;
```
ElinorW, line 11: What's the difference between open and status?

thewahome: Open can be true / false
Status can be used to tell the caller what the popup is doing. 
Has it been closed? (should be used when an action in the popup was successfully completed)
Has it been dismissed? (should be used when no action was done / action was unsuccessful)

ElinorW:
The permissions panel is occupying a lot of space on the screen, can you adjust it to how it was before
_before_
image

_now_
image

ElinorW:
> Yes @ElinorW ... the button moved to where Fluent Defines Dialog Footer buttons to go. I didn't see there being a need to move it

So we are the ones who changed the default Fluent behaviour?

thewahome:
> > Yes @ElinorW ... the button moved to where Fluent Defines Dialog Footer buttons to go. I didn't see there being a need to move it
> 
> So previously we are the ones who changed the default Fluent behavior?

Not exactly. I chose to use the DialogFooter on the Panel and the Dialog components. And it sat right

Onokaev (approved):
Looks good!

src/app/views/common/popups/DialogWrapper.tsx
```diff
@@ -0,0 +1,45 @@
...
+      onDismiss={() => dismissPopup()}
+      dialogContentProps={{
+        type,
+        title: title?.toString(),
+        isMultiline: false,
+        subText: subtitle?.toString()
+      }}
+    >
+      {
+        <Suspense fallback={<Spinner />}>
```
Onokaev, line 34: We need an error boundary since we are using Suspense

Onokaev:
The Modify permissions tab and permissions panel reload every time a consent/unconsent action is performed. The same behavior is not in dev

ElinorW (dismissed):

PR #2523, opened 2023-04-24
Fix: Modify permissions tab resize

### LINKED ISSUE #2516: Enhance permissions revoking experience to be consistent
Revoking permissions is inconsistent due to delays from the eventually consistent data stores. 
The experience should be enhanced in a way that users are presented with accurate information from the permission grants API.

- [x] Consenting and error messages
- [x] #2524

Onokaev:
## Overview
Closes #2516 

### Demo

### Notes
Test on monitors above 25inches

## Testing Instructions

* How to test this PR
1. Resize the browser window
2. Click on Modify Permissions tab
3. Drag the resize handler at the bottom of the permissions tab
4. Resize the browser window again to full screen
5. Notice that the tab resizes well into the screen

### DIFF
--- a/src/app/views/query-runner/request/Request.tsx
+++ b/src/app/views/query-runner/request/Request.tsx
@@ -134,6 +134,18 @@ const Request = (props: any) => {
     dispatch(setDimensions(dimen));
   };
 
+  // Resizable element does not update it's size when the browser window is resized.
+  // This is a workaround to reset the height
+  window.addEventListener('resize', () => {
+    const resizable = document.getElementsByClassName('request-resizable');
+    if (resizable && resizable.length > 0) {
+      const resizableElement = resizable[0] as HTMLElement;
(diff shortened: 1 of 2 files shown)


src/app/views/query-runner/request/Request.tsx
```diff
@@ -134,6 +134,18 @@ const Request = (props: any) => {
     dispatch(setDimensions(dimen));
   };
 
+  // Resizable element does not update it's size when the browser window is resized.
+  // This is a workaround to reset the height
+  window.addEventListener('resize', () => {
```
thewahome, line 139: We need to remove the event listener on unload

Onokaev: Done. I am adding and disposing of it when the size of the screen changes between mobile and full screen. This takes care of the case where a user changes the browser size multiple times without reloading the page

PR #2561, opened 2023-05-04
Hot Fix: display icons in UHF environments

thewahome:
## Overview

Currently in production and production-like environments, icons are not properly displayed. This is likely due to a package that is causing a conflict. This revert action should remove the package changes


thewahome:
CSP work was able to fix the need for this PR. Thanks @ElinorW for the approval

PR #2576, opened 2023-05-16
Fix: Permissions scrollbar

### LINKED ISSUE #2563: Scrollbar appears at the wrong position on the Modify Permissions tab
**Describe the bug**
The scrollbar appears at the wrong position for cases where the permissions items do not fill the entire tab and a horizontal scrollbar is rendered

**To Reproduce**
2. Type the query `https://graph.microsoft.com/v1.0/me/photo`
3. Click on the Modify Permissions tab
4. Resize the screen towards the Mobile view until a scrollbar appears on the tab
5. Notice the positioning of the scrollbar

**Expected behavior**
The scrollbar should appear at the bottom of the tab

**Screenshots**
image

**Additional context**

Onokaev:
## Overview
Closes #2563 

### Demo
Screenshot 2023-05-16 at 12 18 54 PM

### Notes
The DetailsList component has a wrapper div that is added at runtime with a classname of ms-ViewPort. So, we have to add css rules to the div dynamically

## Testing Instructions

* How to test this PR
1. Resize the screen so that the app is in mobile view. 
2. Type the query: https://graph.microsoft.com/v1.0/me/photo on the query textbox
3. Click on the Modify Permissions tab
4. Notice that the scrollbar that appears is at the bottom of the tab


src/app/views/query-runner/request/util.ts/CssHelpers.ts
```diff
@@ -0,0 +1,14 @@
+// DetailsList has a wrapper div that is added at runtime with the classname
```
ElinorW, line 1: You could try adding the height property on the .scss file to achieve the same without having to add this function

ElinorW (changes requested):

src/app/views/query-runner/request/util.ts/CssHelpers.ts
ElinorW: Also, I think it's worth investigating the parent heights... There may be some wrong configurations set

ElinorW (approved):

PR #2585, opened 2023-05-23
Task: Add new permissions scopeType

### LINKED ISSUE #2527: Add `scopeType=DelegatedWork` query parameter on permissions  endpoint
We are updating the DevX API permissions endpoint to allow fetching all permissions scope types if none is specified. the current implementation defaults to `DelegatedWork`.  

Graph Explorer will be required to explicitly specify the scope type when fetching all graph permissions or for a single request URL.

Onokaev:
## Overview
Closes #2527 

### Demo

### Notes

## Testing Instructions

* How to test this PR
Click on Modify Permissions tab. Notice that specific permissions are still rendered
Open the permissions panel. Notice that all available permissions are still rendered

### DIFF
--- a/src/app/services/actions/permissions-action-creator.spec.ts
+++ b/src/app/services/actions/permissions-action-creator.spec.ts
@@ -5,7 +5,7 @@ import {
 } from '../../../app/services/redux-constants';
 
 import {
-  fetchFullScopesSuccess, fetchScopesError, getPermissionsScopeType, fetchScopes,
+  fetchFullScopesSuccess, fetchScopesError, fetchScopes,
   consentToScopes,
   fetchUrlScopesPending,
   fetchFullScopesPending,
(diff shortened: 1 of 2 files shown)


src/app/services/actions/permissions-action-creator.ts
```diff
@@ -173,13 +170,6 @@ export function fetchScopes(scopesFetchType: ScopesFetchType = 'full') {
   };
 }
 
-export function getPermissionsScopeType(profile: IUser | null | undefined) {
-  if (profile?.profileType === ACCOUNT_TYPE.MSA) {
-    return PERMS_SCOPE.PERSONAL;
```
ElinorW, line 178: I think an MSA account has a different scope... `DelegatedPersonal` 
Shouldn't we still pass this scope?

Onokaev: I have seen it. This change is unnecessary. Let me close the PR and the issue with these comments.

Onokaev:
Closing this as we are passing the correct scope type depending on the signed-in user

src/app/services/actions/permissions-action-creator.spec.ts
```diff
@@ -198,18 +198,6 @@ describe('Permissions action creators', () => {
     expect(urlScopesAction).toEqual(expectedUrlScopesAction)
   });
```
fey101, line 200: What's the corresponding expected test with the new changes? Feels like we have introduced new functionality without test coverage.

ElinorW: Actually, there was no need to remove this test. I will restore it.

PR #2615, opened 2023-06-07
Chore: Clear network error message

### LINKED ISSUE #2613: Network error from fetching monaco workers
**Describe the bug**
Some playwright tests are failing because of this warning message that appears on some screenshots. This prevents any page interaction and in turn causes the tests to timeout

**To Reproduce**
Error is mostly caught in playwright snapshots

**Expected behavior**
The should be no errors so that all tests can run uninterrupted.

**Screenshots**
image

ElinorW:
## Overview

This PR removes the Monaco webpack plugin and the getWorkersUrl call as they aren't since we use the `@monaco-editor/react` package. This package does not need us to set up any plugins or modify configuration files for Webpack

Closes #2613


Onokaev (approved):

ElinorW:
Opted to fix the failing snapshots in this PR as well. I noticed some were outdated.

PR #2621, opened 2023-06-12
Feature: Add flighting support

### LINKED ISSUE #1569: Add support for feature flags
To have a chance to preview new features to a limited set of users, or have the chance to remove a feature from public use, we should use feature flags.

For instance, we can release the header in preview so that we receive feedback about it. Or have two headers and show each to different users. Similar case for resource explorer and other major features that are in development.

Connect the configurations to azure so that we can turn things on and off. 

**Additional context**

**Tasks include**
Breaking this into smaller tasks

- [x] Read the Azure Experimentation Platform documentation and gain knowledge about A/B testing, control groups, experiment groups, success metrics, and statistical analysis.
- [x] Set up an experimentation service and create a workspace
- [x] Design and define a sample experiment (create a hypothesis and define the variations, success metrics to measure, and traffic parameters)
- [x] Integrate Azure Experimentation Platform on GE
- [ ] Go through ExP [data contract](https://eng.ms/docs/products/experimentation-platform/docs/datasources/datacontract) and set up necessary telemetry collection on GE  
- [ ] Set up data source on the Azure Experimentation Platform
- [ ] Metrics authoring ([create new metric sets](https://eng.ms/docs/products/experimentation-platform/docs/metricauth/createnewmetricset), [define metrics](https://eng.ms/docs/products/experimentation-platform/docs/metricauth/definenewmetrics), [define segments](https://eng.ms/docs/products/experimentation-platform/docs/metricauth/definesegment) in metric sets and [publish changes to production](https://eng.ms/docs/products/experimentation-platform/docs/metricauth/createeditdeletepublishdrafts))
- [ ] Generate [scorecards](https://eng.ms/docs/products/experimentation-platform/docs/analysis/scorecard)

ElinorW:
## Overview

Closes #1569 

### Demo

### Notes

### Option 1
* Open the staging link provided on different browsers. 
* Click on 'Resources' tab
* Observe if icons are present or only display on hover
* If you can produce both results on two different browsers then it's working
 ### Option 2
* Open the staging link
* Open your browser 'Network' tab in developer tools 
* look for the 'tas' network call
* Confirm it has a 200 status code and the response looks sth like this 
image


src/app/services/variantService.ts
```diff
@@ -0,0 +1,63 @@
...
+    const settings: VariantAssignmentClientSettings = { endpoint: this.endpoint };
+    this.createUser();
+    const request: VariantAssignmentRequest =
+        {
+          parameters: this.getParameters()
+        };
+
+    const client = new VariantAssignmentServiceClient(settings);
+    const response = await client.getVariantAssignments(request);
+    Promise.resolve(response).then((r) => {
```
Onokaev, line 34: We should check whether r is defined before accessing its properties

Onokaev (changes requested):

src/app/utils/local-storage.ts
```diff
@@ -0,0 +1,17 @@
+export const saveToLocalStorage = (key: string, value: Object|string) => {
+  if (typeof value === 'string') {
+    localStorage.setItem(key, value);
+  } else {
+    localStorage.setItem(key, JSON.stringify(value));
+  }
```
thewahome, line 6:
```suggestion
  let itemToStore = value;
  if (typeof value != 'string') {
    itemToStore = JSON.stringify(value)
  } 
  localStorage.setItem(key, itemToStore);
  ```

File: src/app/views/sidebar/resource-explorer/ResourceLink.tsx [hunk 3]
```diff
@@ -27,6 +28,16 @@ const ResourceLink = (props: IResourceLinkProps) => {
   const paths = collections?.find(k => k.isDefault)?.paths || [];
   const resourceLink = { ...link };
 
+  const [alwaysShowButtons, setAlwaysShowButtons] = useState(false);
+
+  useEffect(() => {
+    variantService.getFeatureVariables('default', 'alwaysShowButtons').then((value) => {
```
thewahome, line 34: If the known variants could be stored in a constants file, it would be easy to add / remove / change them and reference them in the code where they are used

src/telemetry/telemetry.ts
```diff
@@ -140,6 +140,15 @@ class Telemetry implements ITelemetry {
       ''
     );
   }
+
+  public getUserId(){
```
thewahome, line 144: This function hasn't been defined in the interface

src/app/views/sidebar/resource-explorer/ResourceLink.tsx
```diff
@@ -27,6 +28,16 @@ const ResourceLink = (props: IResourceLinkProps) => {
...
+      if(value !== undefined) {
```
thewahome, line 35: This condition can be simplified to `if (value)`, which will check for any falsy value, not just undefined. This will make the code more concise and readable.

src/telemetry/telemetry.ts
ElinorW: Ah nice catch let me add that

src/app/views/sidebar/resource-explorer/ResourceLink.tsx
ElinorW: hmm you are saying we should have type variants? where all variants will be stored? or we should be saving all variants as constants in the constants file?

thewahome: All variants in a constants file

ElinorW: Oh got it. This is a great idea, it would help us know what variants have been declared

src/app/views/sidebar/resource-explorer/ResourceLink.tsx
ElinorW: This value is allowed to be false though.... the experiment randomizes the response, some people will get true and some will get false...
In cases where initialization was slow or there is a problem somehwere, it would return undefined and use the default state

thewahome (dismissed):

Onokaev (dismissed):
Looks good!

src/app/services/variant-service.ts
```diff
@@ -0,0 +1,69 @@
...
+interface TasResponse {
+  Id: string;
+  Parameters: Parameters;
+}
+interface Parameters {
+  [key: string]: string | boolean | number;
+}
+class VariantService {
+
+  static myInstance: VariantService = null as any;
```
Onokaev, line 20: This variable appears unused. Do we need it?

ElinorW: Ah, good catch. We don't need it

Onokaev (approved):

PR #2625, opened 2023-06-15
Fix: Show documentation URL for supported Resource Explorer items

thewahome:
## Overview

Some nodes like chats in the Resource Explorer, which have documentation links, didn't have their links showing. This changes now with a difference in implementation to make it similar to the Query bar.

### Demo

Before: 
image

After:
image

### Notes

## Testing Instructions

* Spin up GE locally
* Append ?devx-api=https://graphexplorerapi-staging.azurewebsites.net to change the destination
* open the chats node
* Notice children in the chats node have their documentation showing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "graph-explorer-v2",
-  "version": "9.0.0",
+  "version": "9.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


millicentachieng (commented):
@thewahome Why show the documentation icon if there is no link for the resource?

thewahome:
It helps with the user experience. If we don't have the icon visible the user will not have a consistent experience

ElinorW (approved):

PR #2664, opened 2023-07-03
Fix: Snippet error messaging

### LINKED ISSUE #2429: More accurate error messages
There are scenarios where the call fails but the error presented to the user doesn't reflect the issue clearly. 
Example : 
* code snippets showing snippet not available when it should say retry again or fetching.
* permissions tab showing permission not available when it should say retry again or fetching.

Expected behavior: the message should guide the user to the best action they should take.

Onokaev:
## Overview
Closes #2429 

### Demo

## Testing Instructions

* How to test this PR
Add any url to the query URL textbox then click on the code snippets tab
Notice that the correct error message is displayed


src/app/views/query-response/snippets/snippets-helper.tsx
```diff
@@ -123,6 +128,36 @@ function Snippet(props: ISnippetProps) {
     );
   }
 
+  const displayError = (): JSX.Element | null => {
+    if((!loadingState && snippet) || (!loadingState && !snippetError)){
+      return null;
+    }
+    // OpenAPI generated and C# snippets are returning 404 for snippets not available
```
ElinorW, line 135: don't forget to remove the comments

src/app/views/query-response/snippets/snippets-helper.tsx
```diff
@@ -123,6 +128,36 @@ function Snippet(props: ISnippetProps) {
...
+          <FormattedMessage id='Snippet not available' />
+        </Label>
+      )
+    }
+    else{
+      return (
+        <>
+          {!loadingState &&
+            <Label style={{ padding: 10 }}>
+              <FormattedMessage id='Fetching code snippet failing' />
```
ElinorW, line 153: This part tells the user's to retry... is this what we want?

Onokaev: This uncovered a very hidden edge case that I missed. Thank you @ElinorW

ElinorW (approved):

PR #2680, opened 2023-07-11
Task: Make resources tab consistent with kiota vs code openapi explorer

### LINKED ISSUE #2641: Resources tree consistency
Currently on the resources tree, the nodes look like this 
tree nodes

it should be 
tree vs

Methods (GET, POST, etc) shouldn't have the name of the resource
Children nodes should be places before methods
Children nodes should have the same structure as parent node ie all the rules above

Onokaev:
## Overview
Closes #2641 

### Notes

## Testing Instructions

* How to test this PR
Click on the Resources tab
Notice the changes in icons and links


ElinorW (changes requested):
Hey @Onokaev, 
So we need the methods to come after the children. For example the GET and PATCH should come after the edge, serviceAnnouncement and sharepoint nodes
Also, I'm not sure the icons were supposed to change 😅

ElinorW:
> @ElinorW The methods should come before the child path segments. This is because sometimes there are many child path segments and it is hard to find the methods. Originally the VS Code extension didn't work like this but we asked Vincent to change it. And that icon is going away in VS Code, so we definitely shouldn't copy it.

Oh I see... this hadn't been updated on the issue. Thank you for clarifying this 🙂

ElinorW (approved):

PR #2726, opened 2023-08-03
Fix: Add required header for tip annotations display

### LINKED ISSUE #2636: Encourage callers to use $select and inform callers when there's default properties in the response
Today API consumers regularly call API endpoints without using $select. This impacts performance of client apps (and impacts service COGS too).  Additionally, APIs endpoints that return a default set of properties in a response don't inform callers that there are more properties available, and that they'll need to use $select - which is often a support call driver.  

The API Gateway Service team is trying to address these related issues by including a new "tips" instance annotation in the response to alert developers and encourage them to use $select, if $select is not being used.  This annotation is also only added if the client application provides an opt-in "ms-graph-dev-mode" Prefer request header

As one the the premium tools for API discoverability and testing, we'd like Graph Explorer to send the `prefer: ms-graph-dev-mode` request header, so that the tips instance annotation shows up.

NA - although we vetted a range of ideas with Identity Developer Advisors.

**Additional context**
[Feature spec](https://microsoft-my.sharepoint-df.com/:w:/p/dkershaw/EU1SpMyyyeBJn-eJ00J1_8wBO_HvzOFHt3f6QRksN8fqgg?e=C8oZSD)

ElinorW:
## Overview

Fixes #2636 

### Demo

image

### Notes

- Works for signed-in users 
- Doesn't affect queries with no tip message 

## Testing Instructions

* Check out this branch and load Graph Explorer locally
* Sign in with the preferred account
* Run `/users` query
* Observe tip message


Onokaev:
Remember to ask the portals team to add the new header to their API calls on the proxy service.

Onokaev (approved):

PR #2735, opened 2023-08-09
Fix: Add comments as part of the snippets for consistency

### LINKED ISSUE #2689: Consistent comments for the SDK snippets tab
**Today:**

image

**For Java tab**
image

**Ask:**
From the above screenshot, only C# tab provides additional information on the SDK version that can be used with the code snippets. For consistent experience, we should avail this information for other SDK tabs as well and ensure that the SDK download links text is also consistent with the SDK version text

ElinorW:
## Overview

Fixes #2689 

### Demo
**Before**
image

**After**
image

### Notes

## Testing Instructions

* Check out this branch and run GE locally
* Navigate to the code snippets tab
* Select each snippet 
* Observe comments are now part of the code and not separate


Onokaev:
Nit:
To fix the php tag appearing twice on the editor, we can do this:
const constructBody = () => {
    if(language === 'php'){
      return snippet.replace(/<\?php/g, `<?php\n${addExtraSnippetInformation()}`)
    }
    return addExtraSnippetInformation() + snippet;
  }
 
 Then pass this function to the body property of the editor like so:
 <Monaco
            body={constructBody()}
            language={language}
            readOnly={true}
            height={height}
  />

Then you can remove the check for the php language in the addExtraSnippetInformation function

Onokaev (approved):

PR #2786, opened 2023-09-07
Chore: remove all any

### LINKED ISSUE #2785: Strongly type the code (avoid using any as a type)

I am proposing to strongly type all parts of the code and especially avoid using any as a type. This would be a valuable improvement as it will bring the following benefits:

- **Increased robustness:** Strong typing can help to prevent errors caused by type mismatches. This is because the compiler will be able to catch errors at compile time, rather than runtime.
- **Easier to maintain:** Strong typing can make the code easier to maintain by making it easier to understand and reason about. This is because the types of the variables and expressions are explicit, rather than implicit.
- **Improved readability:** Strong typing can improve the readability of the code by making it easier to see the types of the variables and expressions. This is because the types are not inferred by the compiler, but are explicitly specified by the programmer.
- **No type ambiguity:** The use of any as a type can lead to type ambiguity, which can make the code more difficult to understand and maintain.
- **Improved code safety:** Avoiding the use of any as a type can help to improve the code safety by making it more difficult to introduce errors into the code.

thewahome:
### Notes

Some files will need closely monitored refactoring so they have been omitted entirely from the `any` check

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,7 +51,7 @@ module.exports = {
     '@typescript-eslint/member-ordering': 'off',
     '@typescript-eslint/no-empty-function': 'error',
     '@typescript-eslint/no-empty-interface': 'error',
-    '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-explicit-any': 'error',
     '@typescript-eslint/no-misused-new': 'error',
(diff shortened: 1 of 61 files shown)


thewahome:
Closing this as it's  going to be too large.
It makes sense to incrementally improve the code by making the use of implicit/explicit any produce a warning

PR #2811, opened 2023-10-02
Fix: Autocomplete validation

### LINKED ISSUE #2461: Valid URLs marked as potential errors
**Describe the bug**
When using the textbox to work with a lot of URLs is misleading when using a valid URL. For instance, I'm using this URL and it tells me I _might_ have an error. I don't know how other users feel about this but it's misleading. `https://graph.microsoft.com/v1.0/sites/m365x55726300.sharepoint.com,f2102bda-0e14-4fc5-bbe1-e65a13cd91a4,6be9dd5c-0640-4465-b955-1528279a0227/lists/7ec9d64f-732f-483b-b1d0-5abb149869f1?$expand=drive,items`

image

 - OS: Windows
 - Browser: Edge
 - Version: 111.0.1661.44 (Official build) (64-bit)

### LINKED ISSUE #2688: validate graph queries before making calls to devx API
GE to validate user queries before making API calls to DevX API for both permissions and code snippets and display correct error message to the user.

**Today:**
image

**Expected:**
GE to display error message on the permissions tab & code snippets that the query is incorrect instead of 'snippets not available' & 'check permissions panel'. Snippets not available & check permissions panel message should only be used in scenarios where a user has run a valid query, but we do not have support for snippets or permissions..

thewahome:
## Overview

Fixes #2688 
Fixes #2461

### Demo
These functionalities are disabled as long as the validation flags the URL as invalid
- running a query
- fetching permissions for a query
- showing snippets

image

### Notes

Walking through the resources ensures that we have a working model for the validator to go through. This guarantees that we are working off a known set of resources . 
All queries that do not match a known path created by combining segments in the resource tree are marked as invalid which we can validate with certainity

[The ABNF rules in the ODATA ABNF repository are documentation intended for human readers. They are not intended for generating a ready-to-use URL parser..](https://github.com/oasis-tcs/odata-abnf/issues/99#issuecomment-1600323468)

## Testing Instructions

* Spin up GE
* Write something crazy on the URL
* Watch for Permissions / snippets / run query


src/modules/validation/validation-service.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  private static getResourceValidationError(queryUrl: string, resources: IResource[]): string | null {
+    if (resources.length === 0) {
+      return null;
+    }
+    const sanitizedUrl = sanitizeQueryUrl(queryUrl);
+    const { requestUrl } = parseSampleUrl(sanitizedUrl);
+    const matchingResource = getMatchingResourceForUrl(requestUrl, resources)!;
+    if (!matchingResource) {
+      return 'No resource found matching this query';
```
ElinorW, line 20: Should we localize this?

calebkiage:
I don't think it's wise to disable features if validation fails. There will be false positives that might make the application unusable.

src/modules/validation/validation-service.ts
thewahome: It's localised now. Nice catch

darrelmiller:
@calebkiage I don't think this is going to be a concern.  The vast majority of imprecision in our metadata is due to us missing restrictions not because of incorrect restrictions.

calebkiage:
@darrelmiller, my suggestion was to disable the submit button only if:
- The request is not HTTPS
- The request is not to a known graph host

Other validations (e.g. unknown path segment) would show a warning and disable permissions and snippets requests but still allow sending the request. The server would return a 404 anyway.

ElinorW (dismissed):

ElinorW (approved):

ElinorW (approved):

PR #2821, opened 2023-10-11
Fix: missing paths available in resource tree

### LINKED ISSUE #2820: Missing paths from Resource Explorer
**Describe the bug**
There are some paths that are not available on resource explorer, these paths are present in the resource tree.
GET v1.0/places and it's children 
GET v1.0/ sites
GET beta/me/profile
GET me/drive/root/children

**To Reproduce**
1. Go to Resource explorer
2. Scroll down and look for the paths above
3. Notice that they do not exist on resource explorer
4. Check in https://graphexplorerapi.azurewebsites.net/openapi/tree to confirm them as part of the list

**Expected behavior**
Known paths should be visible on Resource explorer

thewahome:
## Overview

Fixes #2820

### Demo

image

### Notes

`/places` for example, does not have indicators of which version it belongs to. Hence it was not visible in the list of tree nodes. This PR has changed this to make `places` and similar ones that have no version to appear on the list.

## Testing Instructions

* Check for sites as one of the root nodes
* Confirm GET sites as one of the children

### DIFF
--- a/src/app/views/sidebar/resource-explorer/resource-explorer.utils.ts
+++ b/src/app/views/sidebar/resource-explorer/resource-explorer.utils.ts
@@ -19,7 +19,7 @@ export function createResourcesList(
 ): INavLinkGroup[] {
   function getLinkType({ segment, links }: any): ResourceLinkType {
     const isGraphFunction = segment.startsWith('microsoft.graph');
-    const hasChildren = links && links.length > 0;
+    const hasChildren = links && links.length > 1;
     if (hasChildren) {
       return ResourceLinkType.NODE;
     }
@@ -34,11 +34,8 @@ export function createResourcesList(
     const { segment, children, labels } = parent;
     const links: IResourceLink[] = [];
     const childPaths = Array.from(new Set([...paths, segment]));
-    if (methods.length > 1) {
-      if (
-        !searchText ||
-        (searchText && childPaths.some((path) => path.contains(searchText)))
-      ) {
+    if (methods.length >= 1) {
+      if (!searchText || (searchText && childPaths.some((path) => path.contains(searchText)))) {
         methods.forEach((method) => {
           links.push(
             createNavLink(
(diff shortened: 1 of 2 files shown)


ElinorW:
Also clicking on the search items results in this error 
image

src/app/views/sidebar/resource-explorer/resource-explorer.utils.ts
```diff
@@ -94,14 +91,14 @@ export function createResourcesList(
...
-    if (availableMethods.length === 1 && versionedChildren.length === 0) {
+    // if segment has one method only and at most 1 child, do not make segment a node
+    if (availableMethods.length === 1 && versionedChildren.length <= 1 ) {
       paths = [...paths, segment];
       method = availableMethods[0];
     }
     const type = getLinkType({ ...info, links: versionedChildren });
     const enclosedCounter =
-      versionedChildren && versionedChildren.length > 0
+      versionedChildren && versionedChildren.length > 1
```
ElinorW, line 101: This line hinders the counter displaying when a user searches for a resource and the parent node has only 1 child that matches the search term. 
Leaving it as it was ensures the counter always displays
```suggestion
      versionedChildren && versionedChildren.length > 0
```

src/app/views/sidebar/resource-explorer/resource-explorer.utils.ts
```diff
@@ -126,7 +123,7 @@ export function createResourcesList(
       paths: pathItems,
       method: method?.toUpperCase(),
       type,
-      links: versionedChildren,
+      links: versionedChildren.length > 1 ? versionedChildren : [],
```
ElinorW, line 126: This also disrupts the search as parent nodes with one child can't expand...
leaving it as `links: versionedChildren`  fixes the issue
```suggestion
      links: versionedChildren,
```

ElinorW (changes requested):

thewahome:
Closing this as the work handled here has also been taken care of by https://github.com/microsoftgraph/microsoft-graph-explorer-v4/pull/2680

PR #2827, opened 2023-10-17
Fix: Update CLI tab to use uppercase

### LINKED ISSUE #2826: The 'Cli' tab should be 'CLI' and should be written the same way in the comments

ElinorW:
## Overview

Fixes #2826 

### Demo
image

## Testing Instructions

* Check out branch
* Navigate to Code snippets tab
* Select CLI tab
* Observe changes

### DIFF
--- a/src/app/views/query-response/snippets/snippets-helper.tsx
+++ b/src/app/views/query-response/snippets/snippets-helper.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
 import { ITheme, Label, Link, PivotItem, getTheme } from '@fluentui/react';
 import { useEffect, useState } from 'react';
 import { useDispatch } from 'react-redux';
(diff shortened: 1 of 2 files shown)


src/app/views/query-response/snippets/snippets-helper.tsx
```diff
@@ -30,10 +31,12 @@ interface ISupportedLanguages {
 
 export function renderSnippets(supportedLanguages: ISupportedLanguages) {
   const sortedSupportedLanguages: ISupportedLanguages = {};
-  Object.keys(supportedLanguages).sort().forEach(key => {
+  sortedSupportedLanguages.CSharp = supportedLanguages.CSharp;
+  Object.keys(supportedLanguages).filter((language) => language !== 'CSharp').sort().forEach(key => {
```
calebkiage, line 35: It might be easier to use the [`sort(compareFn)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) version of the sort which will save having to filter out csharp first.

calebkiage (changes requested):

src/app/views/query-response/snippets/snippets-helper.tsx
ElinorW: I don't see how the `compareFn` makes it easier.. I see the alternative as just splicing then unshifting...

calebkiage: I was thinking:

```javascript
Object.keys(supportedLanguages).sort((l1, l2) => {
  if (l1 == 'CSharp') {
    return -1;
  }
  if (l2 == 'CSharp') {
    return 1;
  }
  if (l1 < l2) {
    return -1;
  } else if (l1 > l2) {
    return 1;
  }
  return 0;
})
```

thewahome (approved):

PR #3009, opened 2024-02-16
Feature: Add a notification banner

ElinorW:
## Overview

Adds a notification banner to show announcement messages.

## Notes
Follows markdown


src/app/views/main-header/MainHeader.tsx
```diff
@@ -42,6 +47,14 @@ export const MainHeader: React.FunctionComponent<MainHeaderProps> = (props: Main
     feedbackIconAdjustmentStyles, tenantIconStyles, moreInformationStyles,
     tenantLabelStyle, tenantContainerStyle } = mainHeaderStyles(currentTheme, mobileScreen);
 
+  useEffect(() => {
+    variantService.getFeatureVariables('default', BANNERMESSAGE).then((value) => {
+      if (value) {
+        setBannerMessage(value as string);
+      }
```
musale: This method `getFeatureVariables` is declared and used in an _async_ manner but it does nothing async. Removing `async` from the method declaration should get rid of the `then` block.

```typescript
const value = varianceService.getFeatureVariables('default, BANNERMESSAGE) as string;
if(value){setBannerMessage(value)};
```

ElinorW:
Closing this as it's no longer needed

PR #3020, opened 2024-02-28
Fix: show all resources

### LINKED ISSUE #3011: Unhidden APIs not showing up in Graph explorer
**Describe the bug**
The permissions for `beta/security/auditLog/queries` are not showing up and neither can the URL be recognised as valid

**Expected behavior**
Permissions and autocomplete options for `beta/security/auditLog/queries` to be visible

**Screenshots**
stackoverflow unhidden apis not visible on GE

thewahome:
## Overview

Fixes #3011 

GE has been calling the DevX API `openapi/tree` endpoint to get the consolidated open api tree document with both versions. This leads to a huge file. 

Currently the built in Javacript `response.json()` function truncates the content as it parses the response string to JSON. This leads to some of the children not being passed from the browser to the code hence losing some nodes along the way.

This new implementation 
- makes parallel calls to get the content of each version. 
- changes the schema so that the resources in state are grouped by the version
- removes the function that was used to filter nodes based on their version as the items are already versioned

### Demo
`security/auditLog`

image

`security/auditLog/queries`

image

### Notes

## Testing Instructions

* Switch to beta
* In the query input, type in `security/` and confirm you get `auditLog` as one of the options
* Select `auditLog` and add a `/` to trigger a new autocomplete action and confirm you get `queries` as one of the options

### DIFF
--- a/src/app/services/actions/autocomplete-action-creators.spec.ts
+++ b/src/app/services/actions/autocomplete-action-creators.spec.ts
@@ -90,11 +90,7 @@ const mockState: ApplicationState = {
   },
   resources: {
     pending: false,
-    data: {
-      segment: '',
-      labels: [],
-      children: []
-    },
+    data: {},
     error: null
   }
 }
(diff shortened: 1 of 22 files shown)


ElinorW (approved):

adhiambovivian:
@thewahome  what changed to trigger this bug ?

thewahome:
It was an unknown issue that had always existed

PR #3107, opened 2024-04-15
Fix: malformed json error

### LINKED ISSUE #3091: 400 Bad Request is shown if the response is a 200 with a malformed JSON
**Describe the bug**
If the server response is 200 with a malformed JSON body, Graph Explorer UI shows `400 Bad Request`.

**Expected behavior**
4xx responses indicate client error. There is no client error in this case. https://datatracker.ietf.org/doc/html/rfc7231#section-6.5
 
We should either propagate the server response as is or if we want to signal that the post processing of the JSON fails, we need an alternative to `400 Bad Request` as it is misleading.

thewahome:
## Overview
Some API responses are returned as incomplete json objects. They've been treated as 400 bad requests.
This change creates a different response that is clearer since the response can't be displayed as is

Closes #3091 

### Demo

image

### DIFF
--- a/src/app/services/actions/query-action-creator-util.ts
+++ b/src/app/services/actions/query-action-creator-util.ts
@@ -221,6 +221,14 @@ export async function generateResponseDownloadUrl(
   }
 }
 
+async function tryParseJson(textValue: string) {
+  try {
+    return JSON.parse(textValue);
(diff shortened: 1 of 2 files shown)


musale (approved):
LGTM

PR #3123, opened 2024-04-26
Feature: support text csv

### LINKED ISSUE #2872: Support “edm.stream” format in Graph Explorer
We have recently introduced an API endpoint whose response is of type "text/csv."
https://graph.microsoft.com/beta/employeeexperience/goals/exportJobs/{jobId}/content

Here,  "/content" property is of type "Edm.Stream." on export job.
When we hit the above endpoint in graph explorer, we get an empty object; however, we can see the data in the console. It is happening because Graph Explorer currently only supports JSON and showing user photos.

Can we please start supporting response type "text/csv"?

Or, is there a way we can tooltip when there is an Edm.Stream?

Schema PR: https://dev.azure.com/msazure/One/_git/AD-AggregatorService-Workloads/pullrequest/8561460?_a=files&path=/Workloads/Microsoft.Viva.Goals/override

thewahome:
## Overview

Closes #2872 

The query mentioned returns a text/csv response that can be properly handled by the browser. This PR adds text/csv headers as supported headers and enables showing the response as plain text on the response viewer

### Demo

image

### DIFF
--- a/src/app/services/actions/query-action-creator-util.ts
+++ b/src/app/services/actions/query-action-creator-util.ts
@@ -236,6 +236,7 @@ export function parseResponse(
         return response.json();
       case ContentType.XML:
       case ContentType.HTML:
+      case ContentType.TextCsv:
       case ContentType.TextPlain:
         return response.text();
 
--- a/src/app/services/context/validation-context/ValidationProvider.tsx
+++ b/src/app/services/context/validation-context/ValidationProvider.tsx
@@ -34,6 +34,9 @@ export const ValidationProvider = ({ children }: ValidationProviderProps) => {
   }, [resources])
 
   useEffect(() => {
+    if (!queryVersion || !query || Object.keys(resources.data).length === 0) {
+      return;
+    }
     if (version !== queryVersion && GRAPH_API_VERSIONS.includes(queryVersion)
       && resources.data[queryVersion].children!.length > 0) {
       setVersionedResources(resources.data[queryVersion].children!);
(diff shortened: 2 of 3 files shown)


src/app/services/actions/query-action-creator-util.ts
```diff
@@ -236,6 +236,7 @@ export function parseResponse(
         return response.json();
       case ContentType.XML:
       case ContentType.HTML:
+      case ContentType.TextCsv:
```
musale, line 239: @thewahome this looks good to me and works well.

I have a suggestion on handling these content types differently. Starting with the overview on media types https://www.rfc-editor.org/rfc/rfc2046.html#section-3, we can make a few changes on how we handle content types:
1. check the top-level media type of the content type in `getContentType`. This means, we return the type as `application`, `text`, or `image`. Alternativley, we can just split the content type on `/` and get the value.
2. we make the switch case to group the content types by type, therefore, all `text` types are rendered as text, `image` types can have their own image render and `application` we can process as we want i.e. json/xml -> format.
3. since we have few content types to parse, we can improve UI display by having customized previews of the response i.e. csv in a table.

NB: this is purely a good to have, as it is, it works well.

musale (approved):

src/app/services/actions/query-action-creator-util.ts
thewahome: I would love to explore this further. If you could have it as a feature request that would help us triage it better

Issue #895, opened 2021-03-16
Graph Explorer become empty page when click  GET: all groups I belong to (direct 


wy193777:
**Describe the bug**

**To Reproduce**
1. Go to 'Graph Explorer'
2. Click on 'signin and sign in your coop account'
3. Scroll down to 'Groups' on the left
4. Click `GET: all groups I belong to (direct
5. page become empty, there are errors on developer console:
```
SyntaxError: Unexpected token e in JSON at position 16162
    at JSON.parse (<anonymous>)
    at Monaco.tsx:64
    at ub (Monaco.tsx:23)
    at qi (react-dom.production.min.js:3183)
    at Ao (react-dom.production.min.js:3795)
    at Fs (react-dom.production.min.js:5390)
    at Bs (react-dom.production.min.js:5409)
    at Ta (react-dom.production.min.js:5826)
    at Sa (react-dom.production.min.js:5794)
    at Ma (react-dom.production.min.js:5860)
    at Object.notify (Subscription.js:23)
    at e.t.notifyNestedSubs (Subscription.js:63)
    at r.i.notifySubscribers (Provider.js:57)
    at e.t.handleChangeWrapper (Subscription.js:68)
    at h (redux.js:213)

TypeError: Cannot read property 'getModeId' of null
    at languageFeatures.js:110
    at e.invoke (zone.min.js:1)
    at t.run (zone.min.js:1)
    at zone.min.js:1
    at e.invokeTask (zone.min.js:1)
    at t.runTask (zone.min.js:1)
    at i (zone.min.js:1)
```

**Expected behavior**
The window still showing up.

**Screenshots**
![image](https://user-images.githubusercontent.com/2449450/111377134-8b2d1480-865d-11eb-9735-9b08cb9aa40b.png)

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Browser [Chrome, Edge]
 - Version:
Microsoft Edge Version 89.0.774.54 (Official build) (64-bit)
Google Chrome Version 89.0.4389.82 (Official Build) (64-bit)

**Smartphone (please complete the following information):**

**Additional context**
 
[AB#8621](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8621)

PR #3163, opened 2024-05-29
fix: remove the replacement of escape characters in json response formatting

### LINKED ISSUE #3158: Graph Explorer does not return Json formatted output
**Describe the bug**

Graph Explorer is not returning Json formatted output. 
Endpoints being tested so far:
https://graph.microsoft.com/v1.0/appCatalogs/teamsApps?$expand=* 
https://graph.microsoft.com/v1.0/appCatalogs/teamsApps?$expand=appDefinitions
The same API works as expected if I pull Teams Apps information from other tenants.
My assumption is that this could be because the amount of information returned may varies. But I’m not sure.

**To Reproduce**
1. Go to (https://developer.microsoft.com/en-us/graph/graph-explorer)
GET 

https://graph.microsoft.com/v1.0/appCatalogs/teamsApps?$expand=* 
https://graph.microsoft.com/v1.0/appCatalogs/teamsApps?$expand=appDefinitions

4. I get a 200 OK response.
5. The output is not easy to read because it is not Json formated.

**Expected behavior**
The output being returned in Json format.

musale:
## Overview

Removes replacement of escaped characters in the stringified json.
closes #3158

## Testing Instructions

* run the query `https://graph.microsoft.com/v1.0/appCatalogs/teamsApps?$expand=*`
* Check the response preview tab to have well formatted json


SilasKenneth:
Tested in Safari on iOS 17.5.1
IMG_0444

musale:
Thank you @SilasKenneth! Also tested the paths in the #895 and they work as expected.

thewahome (approved):

thewahome:
@musale since the test works, the stringify -> parse is an unnecessary set of legacy steps. Is this something that we can confidently remove?

musale:
> @musale since the test works, the stringify -> parse is an unnecessary set of legacy steps. Is this something that we can confidently remove?

Sure! Unless that telemetry is important, I don't see the need for that function too.

PR #3212, opened 2024-06-12
fix: add query input slice

### LINKED ISSUE #3203: Replace query-input actions, action creators and reducers with RTK slices

### LINKED ISSUE #3204: Replace query-status actions, action creators and reducers with RTK slices

musale:
### Description

Adds query input slice reducer (samples query)
Updated actions in various files

Closes #3203


src/app/services/reducers/index.ts
```diff
@@ -12,7 +13,7 @@ import { graphExplorerMode } from './graph-explorer-mode-reducer';
 import { scopes } from './permissions-reducer';
 import { profile } from './profile-reducer';
 import { proxyUrl } from './proxy-url-reducer';
-import { sampleQuery } from './query-input-reducers';
+// import { sampleQuery } from './query-input-reducers';
```
thewahome, line 16: We can delete this

thewahome: The `query-input-reducers` and the associated test files can also be deleted since they are no longer in use

PR #3258, opened 2024-07-16
Task: create history slice

### LINKED ISSUE #3205: Replace request-history actions, action creators and reducers with RTK slices

thewahome:
## Overview

Closes #3205


musale (approved):

src/app/services/slices/history.slice.ts
```diff
@@ -0,0 +1,33 @@
...
+    },
+    bulkAddHistoryItemsSuccess(state, action: PayloadAction<IHistoryItem[]>) {
+      state.push(...action.payload);
+    },
+    removeHistoryItemSuccess(state, action: PayloadAction<IHistoryItem>) {
+      state = state.filter(item => item !== action.payload);
+    },
+    removeAllHistoryItemsSuccess(state, action: PayloadAction<string[]>) {
+      state = state.filter(item => !action.payload.includes(item.createdAt));
+    }
```
musale, line 22: I was reading this https://redux-toolkit.js.org/usage/immer-reducers#resetting-and-replacing-state in regards to changing the entire state. In our case, we are creating a new array with `filter`, so to mutate the state, we need to return this filtered array, and not assign it to the `state` variable.

PR #3263, opened 2024-07-25
Task: complete slices query runner status

thewahome:
## Overview
- enhance and use query status slice
- fix status related console errors
- remove redundant query reducers


src/app/services/slices/query-status.slice.ts
```diff
@@ -1,19 +1,28 @@
...
-const initialState: Status = {}
+import { IStatus } from '../../../types/status';
+import { LOGOUT_SUCCESS, QUERY_GRAPH_RUNNING } from '../redux-constants';
 
-const queryStatus = createSlice({
-  name: 'queryStatus',
-  initialState,
+const queryRunnerStatusSlice = createSlice({
+  name: 'queryRunnerStatus',
+  initialState: null as IStatus | null,
```
musale, line 8: Hmm... I wonder why you're casting null to `Istatus`? Is it because `initialState` can have fields with `null` values?

thewahome: yes.

thewahome: and it was the only way to keep the compiler from shouting at me

musale (approved):

PR #3286, opened 2024-08-21
feat: investigate fluentui v9 migration

Mnickii:
## Overview

This PR investigates possibility of migrating GE components to use fluentUI v9. 
Migrates some button usage and investigates theming changes

W.I.P

### Notes

Theming changes require more work
This PR showcases instances where some component counterparts may not work to the desired GE experience

## Testing Instructions

* How to test this PR : Test on the jolly link


adhiambovivian:
@Mnickii  some feedback;

1.Please ensure the  run query text area is responsive to the size of the Graph query. In this upgrade, the size of this text area is not responsive to the Graph query text string size

image

2. The docs icon, sample query text and method names seems to be 'floating' ie too much padding. Can the size of these be increased to reduce the padding?
image

3. The dividers here are overlapping, refer to the image
image

4. Vertical scrollbars for Samples is missing - refer to the image (taken from Graph Explorer pro)
image

5. The header is not aligned with the rest of the GE components mainly the share icon. Additionally, the share icon (on my screen) has be cut off and the rest of the components on the right side of the screen. When Graph Explorer view is minimized using the minimize side bar (at the top left), the problem does not exist
image

6. Any reason why the docs icons changed from blue to black & white?
image
7. The GE resources, history etc icons also changed from blue to black and white. When you hover the icons, only the sample queries icon changes color to blue.
image

8. Consent type on the permission tab is not visible even when I resize the browser tab window or zoom out
image

Mnickii:
@adhiambovivian as this is still a work in progress and a proof of concept, not all parts of the application have been migrated into fluentUI v9. This is meant to showcase how migration can be done, and in particular, how some experiences may need adapting to use different components. Theming also hasn't been migrated app-wide. Some of these changes are fixable in the scope of the POC but others would require actual project-wide migration.

config/webpack.config.js
```diff
@@ -419,6 +420,8 @@ module.exports = function (webpackEnv) {
       new webpack.optimize.LimitChunkCountPlugin({
         maxChunks: 1
       }),
+      // remove after comparing bundle sizes
```
musale, line 423: Any insights on bundle sizes?

src/app/views/Layout.styles.ts
```diff
@@ -0,0 +1,94 @@
+import { makeStyles, tokens } from '@fluentui/react-components';
```
musale, line 1: Given the new designs and the effort needed to make the component look as the design as possible, what is the effort you figure is needed to achieve the migration? Are there lots of CSS tinkering? Does this file also mean there will be a change in the way styling is done (theming)?

musale:
@Mnickii going forward, is it possible to break this out into smaller portions where we have scoped PRs?

PR #3337, opened 2024-09-25
fix: clone the resources and re-assign children to found values

### LINKED ISSUE #3335: Graph Explorer Resources not working properly
**Describe the bug**
When user goes to resources tab and search something in the resources search bar, the entire Graph Explorer web app disappears, and only the footer is visible. 

**To Reproduce**
2. Go to resources tab
3. Enter 'folder' or any other phrase in the search bar
4. See error --> 
image

**Expected behavior**
The correct resources and request should be returned. 

**Screenshots**
image

musale:
## Overview

Use `Object.assign` to clone the resources object and re-assign the `.children` to the found resources.

Closes #3335


src/app/utils/resources/resources-filter.ts
```diff
@@ -1,22 +1,23 @@
...
-        resource.children = foundChildResources;
-        foundResources.push(resource);
+        const res = Object.assign({}, resource)
+        res.children = foundChildResources;
+        foundResources.push(res);
       }
     }
   });
-  return foundResources;
+  return foundResources
```
Mnickii, line 20:
```suggestion
  return foundResources;
```

ElinorW (approved):

Issue #3050, opened 2024-03-14
GE showing error messages for sample queries.


jessicaaawu:
**Describe the bug**
GE is showing this "possible error found in URL near: $value" message is showing up for one of our default samples such as GET my photo.

Several samples are also displaying similar error messages such as GET list items in my drive. 

**To Reproduce**
1. Go to https://developer.microsoft.com/en-us/graph/graph-explorer 
2. Click on get my photo
4. See error message "possible error found in URL near: $value" 

Examples of other sample queries reproducing this error: 
GET list items in my drive, 

**Expected behavior**
No error message should show for sample queries. 

**Screenshots**
![image](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/assets/56090713/1ad0b523-85ab-472d-a623-6212b99922bd)
![image](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/assets/56090713/528a886a-37d4-4c6a-8d97-3a80026b3cb7)

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Browser [Microsoft Edge]

**Additional context**
Please investigate this and fix the error messages.

adhiambovivian:
Issue was not resolved, re-opening the issue. See latest screenshot

![image](https://github.com/user-attachments/assets/30478e3f-468b-43a7-8e4f-c677734fa9c1)

PR #3346, opened 2024-09-30
fix: update the odata abnf rules and check odata relative URIs

musale:
## Overview
#3050

Some paths like `/me/$value` fail. These checks fail when the rule is `odataUri`. This change adds:
- a check for the `odataRelativeUri` when the first check fails.
- Updated rules to the core v4.01 specification.
- Cache the rules and get them from https://github.com/oasis-tcs/odata-abnf/blob/main/abnf/odata-abnf-testcases.yaml

### Demo

`$value` and `$count` should pass.

### Notes

Local tests are failing due to caching so working on fix.


musale:
Superseded by #3422

PR #3381, opened 2024-10-04
Fix: Update playwright tests

ElinorW:
## Overview
Fixes failing resources test.
Updates test to ignore authenticated tests due to MFA restrictions

### Demo

### Notes

Authenticated tests will be disabled for now since MFA is required and there's no way to handle this when running headless tests on playwright

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -19,6 +19,7 @@ const config: PlaywrightTestConfig = {
     screenshot: 'only-on-failure'
   },
   testDir: './src/tests',
+  testIgnore: './src/tests/authenticated-experiences/**',
   reporter: [
     [
       'html',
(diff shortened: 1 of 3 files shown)


andrueastman (approved):

playwright.config.ts
```diff
@@ -19,6 +19,7 @@ const config: PlaywrightTestConfig = {
     screenshot: 'only-on-failure'
   },
   testDir: './src/tests',
+  testIgnore: './src/tests/authenticated-experiences/**',
```
andrueastman, line 22:
```suggestion
  testIgnore: './src/tests/ui/authenticated-experiences/**',
```

andrueastman (approved):

ElinorW:
Playwright tests passed 
image

andrueastman (approved):

PR #3398, opened 2024-10-31
Add fluent v9 components and react-icons

### LINKED ISSUE #3383: Migrate GE general theming from v8 to v9

musale:
Closes #3383 

Bootstraps fluent ui v9 to work alongside fluent ui v8 with theming set.


package.json
```diff
@@ -9,6 +9,8 @@
                 "@babel/core": "7.25.2",
                 "@babel/runtime": "7.25.6",
                 "@fluentui/react": "8.120.7",
+                "@fluentui/react-components": "^9.55.1",
```
ElinorW, line 12: Let's remove the carets (^) to avoid long pipeline runs

musale: Got it!

PR #3441, opened 2024-11-26
feat: update request body area

ElinorW:
## Overview

Update request body, request header, modify permissions and access token tabs

### Demo

image
image
image

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,6 @@
         "dependencies": {
                 "@augloop/types-core": "file:packages/types-core-2.16.189.tgz",
                 "@axe-core/webdriverjs": "4.10.0",
-                "@fluentui/react-components": "9.55.1",
-                "@fluentui/react-icons": "2.0.264",
                 "@azure/msal-browser": "3.26.1",
                 "@babel/core": "7.26.0",
                 "@babel/runtime": "7.26.0",
--- a/src/app/views/query-runner/request/auth/AuthV9.tsx
+++ b/src/app/views/query-runner/request/auth/AuthV9.tsx
@@ -54,7 +54,9 @@ const useStyles = makeStyles({
 });
 
 export function Auth() {
-  const { auth: { authToken }, profile, dimensions: {request:{height}}} = useAppSelector((state) => state);
+  const profile = useAppSelector((state) => state.profile);
+  const height: string = useAppSelector((state) => state.dimensions.request.height);
+  const authToken = useAppSelector((state) => state.auth.authToken);
   const { user } = profile;
   const requestHeight = convertVhToPx(height, 60);
   const [accessToken, setAccessToken] = useState<string | null>(null);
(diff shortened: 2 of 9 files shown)


src/app/views/query-runner/request/auth/AuthV9.tsx
```diff
@@ -0,0 +1,133 @@
...
+    display: 'flex',
+    width: '100%',
+    minHeight: '100%',
+    justifyContent: 'center',
+    alignItems: 'center'
+  }
+});
+
+export function Auth() {
+  const { auth: { authToken }, profile, dimensions: {request:{height}}} = useAppSelector((state) => state);
```
musale, line 57: Use state values individually

src/app/views/query-runner/request/auth/AuthV9.tsx
```diff
@@ -0,0 +1,133 @@
...
+    authenticationWrapper
+      .getToken()
+      .then((response: AuthenticationResult) => {
+        setAccessToken(response.accessToken);
+        setLoading(false);
+      })
+      .catch(() => {
+        setLoading(false);
+      });
+  }, [authToken]);
```
musale, line 80: `authToken` is never used in this `useEffect`. Remove it.

src/app/views/query-runner/request/headers/RequestHeadersV9.tsx
```diff
@@ -0,0 +1,120 @@
...
+    minWidth: '80px'
+  },
+  listContainer: {
+    flex: 1,
+    overflow: 'auto'
+  }
+});
+
+const RequestHeaders = () => {
+  const { sampleQuery, dimensions: { request: { height } } } = useAppSelector((state) => state);
```
musale, line 42: Get state variables individually.

ElinorW:
> This is coming along well! I tested locally and everything looks great, really good stuff. I experienced a slight issue with resizing though. In some cases, I can resize up/down and when I release, I can't resize again. In other cases, I could not trigger a resize. See this:
> 
>  resizing-request-region.mp4

Ahh this is a nice catch, I'm going to test the resizing some more

src/app/views/query-runner/request/auth/AuthV9.tsx
musale: I have understood why this is like this. To trigger the useEffect to set the authToken when it's available. Leave it as is.

package.json
```diff
@@ -5,8 +5,6 @@
         "dependencies": {
                 "@augloop/types-core": "file:packages/types-core-2.16.189.tgz",
                 "@axe-core/webdriverjs": "4.10.0",
-                "@fluentui/react-components": "9.55.1",
-                "@fluentui/react-icons": "2.0.264",
```
Mnickii, line 9: Oh we missed this, let's not remove the packages

Mnickii (changes requested):

package.json
ElinorW: they had been repeated

musale (approved):

Mnickii (approved):

PR #3442, opened 2024-11-26
feat: migrate theme popover to v9

### LINKED ISSUE #3439: Migrate components in theming popover

Mnickii:
## Overview

Closes #3439 

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description


src/app/views/main-header/settings/ThemeChooserV9.tsx
```diff
@@ -0,0 +1,68 @@
...
+    case 'high-contrast':
+      return dispatch(changeTheme('high-contrast'));
+    }
+    telemetry.trackEvent(eventTypes.BUTTON_CLICK_EVENT, {
+      ComponentName: componentNames.SELECT_THEME_BUTTON,
+      SelectedTheme: newTheme.replace('-', ' ').toSentenceCase()
+    });
+  };
+  return (
+    <Menu>
```
musale, line 48: I think a better component to use here is a [Radio Group](https://react.fluentui.dev/?path=/docs/components-radiogroup--docs#layout--horizontal).

Also, I noticed that the theming does not stick. If I choose web light, when I reload, it goes back to the initial one. The theme should be persisted on local storage and remain after reloads. (happens in current GE so it's a good bug to fix on the go)

Lastly, I've seen this on theme choosers where there's an option to "use system default". Here we already check the default system them and we can have it as the default selected. Also, we can add an event listener that captures changes to the default system theme and applies them appropriately.

src/app/views/main-header/settings/ThemeChooserV9.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const dispatch = useAppDispatch();
+  const appTheme = useAppSelector(state=> state.theme);
+
+
+  const handleChangeTheme = (selectedTheme: { key: string; displayName: string; icon: JSX.Element }) => {
+    const newTheme: string = selectedTheme?.key ?? '';
+    // Applies the theme to the Fluent UI components
+    switch (newTheme) {
+    case 'light':
+      return dispatch(changeTheme('light'));
```
musale, line 36: Using a `return` here means that telemetry will never be logged. Remove returns.

src/app/views/main-header/settings/ThemeChooserV9.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const ThemeChooserV9: React.FC<PopupsComponent<null>> = () => {
+  const dispatch = useAppDispatch();
+  const appTheme = useAppSelector(state=> state.theme);
+
+
+  const handleChangeTheme = (selectedTheme: { key: string; displayName: string; icon: JSX.Element }) => {
+    const newTheme: string = selectedTheme?.key ?? '';
+    // Applies the theme to the Fluent UI components
+    switch (newTheme) {
```
musale, line 34:
```suggestion
 dispatch(changeTheme(newTheme))
 loadGETheme(newTheme)   
```

Remove the switch statement because the chooser (in this case menu items) give us defined values.
Use `loadGETheme` to also cater for v8 components. Add a note/todo to remove it later.

src/app/views/main-header/settings/ThemeChooserV9.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+];
+
+const ThemeChooserV9: React.FC<PopupsComponent<null>> = () => {
+  const dispatch = useAppDispatch();
+  const appTheme = useAppSelector(state=> state.theme);
+
+
+  const handleChangeTheme = (selectedTheme: { key: string; displayName: string; icon: JSX.Element }) => {
+    const newTheme: string = selectedTheme?.key ?? '';
```
musale, line 32:
```suggestion
    const newTheme: string = selectedTheme.key;
```

src/app/views/main-header/settings/ThemeChooserV9.tsx
Mnickii: Nice catch! Yeah we can definitely use the default system theme here. Will make this change.

Mnickii: I could definitely use the Radio Group/ Radio component, customize the component to use slots for the icons.

src/app/views/main-header/settings/ThemeChooserV9.tsx
```diff
@@ -0,0 +1,88 @@
...
+    dispatch(changeTheme(newTheme));
+    loadGETheme(newTheme); //Remove when cleaning up
+    telemetry.trackEvent(eventTypes.BUTTON_CLICK_EVENT, {
+      ComponentName: componentNames.SELECT_THEME_BUTTON,
+      SelectedTheme: newTheme.replace('-', ' ').toSentenceCase()
+    });
+  };
+  return (
+    <RadioGroup layout="horizontal" aria-labelledby='theme-chooser'
+      defaultValue={availableThemes.find(theme => theme.key === appTheme)?.displayName}
```
musale, line 65: Fantastic work so far!

There's a console warning that draws attention to the implementation you have here
image

My understanding is that you can't use `defaultValue` and `checked` at the same time. My recomendation here is to:
1. use the `onChange` listener on `RadioGroup` and remove the `defaultValue` for `value`.
2. You can then create a state `value` which defaults to the `appTheme`, and use all the implementation of `handleChangeTheme` after `setValue(data.value)`. Something like this:
```typescript
 const [value, setValue] = useState(appTheme);
// some stuff
 <RadioGroup value={value} onChange={(_, data) =>{
        const newTheme: string = data.value
        setValue(newTheme)
        // Applies the theme to the Fluent UI components
        dispatch(changeTheme(newTheme));
        loadGETheme(newTheme); //Remove when cleaning up
        telemetry.trackEvent(eventTypes.BUTTON_CLICK_EVENT, {
          ComponentName: componentNames.SELECT_THEME_BUTTON,
          SelectedTheme: newTheme.replace('-', ' ').toSentenceCase()
        });
      }}>
// other stuff
```
3. Remove `checked` and `onClick` on `Radio`.

Borrowed this from https://react.fluentui.dev/?path=/docs/components-radiogroup--docs#controlled-value

ElinorW:
Noticed the close button is no longer there, shouldn't we still keep it? but just as a regular button?

Mnickii:
@ElinorW since clicking outside the popover closes it, I wonder if we should keep it.

src/app/views/main-header/settings/ThemeChooserV9.tsx
Mnickii: I'm currently unable to test this locally, but I'm curious if removing the Radio Checked on ln 71 in current implementation would have a similar warning. Per https://react.fluentui.dev/?path=/docs/components-radiogroup--docs#default-value, only either the default Value or the checked can be used

ElinorW:
> @ElinorW since clicking outside the popover closes it, I wonder if we should keep it.

Hmm good point, the same behaviour is there in the current version as well. 
Maybe we can ask design.

src/app/views/main-header/settings/ThemeChooserV9.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { makeStyles, RadioGroup, Radio} from '@fluentui/react-components';
+import { useAppDispatch, useAppSelector } from '../../../../store';
+import { componentNames, eventTypes, telemetry } from '../../../../telemetry';
+import { PopupsComponent } from '../../../services/context/popups-context';
+import { changeTheme } from '../../../services/slices/theme.slice';
+import { loadGETheme } from '../../../../themes';
+import { translateMessage } from '../../../utils/translate-messages';
+import { BrightnessHighRegular, WeatherMoonFilled, CircleHalfFillFilled } from '@fluentui/react-icons';
+
+const availableThemes = [
```
musale, line 11: Since you've added system theme support, make it an option and probably the default selected.

musale (approved):
In `App.tsx`, we want to do two things:
- Get the current theme or load the system theme and set it. Allows us to set the theme on app load.
- Create a listener to the theme changes. Allows us to change our theme if a user changes the theme from their system. Something like this:
   ```js
   const mediaQueryList = window.matchMedia('(prefers-color-scheme: dark)');
   // Add the listener for theme changes
   mediaQueryList.addEventListener('change', handleThemeChange);
   ```

What do you think?

Mnickii:
Yeah @musale, I noticed the theme was only picked when interacting with theme chooser, currently implementing this.

src/app/views/main-header/settings/ThemeChooserV9.tsx
Mnickii: It's the default selected ln70-75

musale: I meant on the popup 
image

PR #3619, opened 2025-02-14
Fix: Response and Request area heights and style

ElinorW:
## Overview

Fixes the issues with the request and response area overflows.
Adjusts styling to match what is on the design.

### Demo

image
image

### Notes

## Testing Instructions

* 
* Prefer bulleted description


musale (commented):
In request headers, could you -re-adjust the height of the headers container so that it doesn't overflow and the lower end of the scroll bar can be seen 
image

Mnickii:
The query runner and headers are not responsive in mobile view
screenshot-1739802106196

public/index.html
```diff
@@ -66,7 +66,7 @@
       margin: 0;
       padding: 0;
       border: 0;
-      overflow: hidden;
+      /* overflow: hidden; */
```
Mnickii, line 69: let's delete

src/app/views/query-response/adaptive-cards/AdaptiveCardV9.tsx
```diff
@@ -275,13 +275,12 @@ const RenderJSONSchema = (props: RenderCardJSONProps) => {
           </Link>
         </MessageBarBody>
       </MessageBar>
-      <div>
+      <div style={{ display: 'flex', flexDirection: 'column',height: '500px' }}>
```
Mnickii, line 278: let's use classNames for consistency

Mnickii (commented):
Mobile view is looking great! 
I've noticed though that I can't scroll to the response area.
screenshot-1739809205234

PR #3628, opened 2025-02-18
ci: updates outdated sonarcloud action

baywet:


baywet:
NOTE: the workflow is failing because I don't have write permissions on this repo and had to create the PR from a fork. Which then means we can't access secrets.

ElinorW (approved):

baywet:
@ElinorW can you also merge please? I don't have permissions

PR #3633, opened 2025-02-19
fix: regressions in component functionality

### LINKED ISSUE #3630: fix functional regressions in GE migration
- [ ] Whenever you select a query to run, the v1.0 version number gets truncated to v1. Same happens with beta --> bet
- [ ] Tip information does not 'dissapear' or close whenever you switch between a sample which has one to another which doesn't have. For example, I'm on my profile sample but I'm still seeing tip for a /drive sample
- [ ] Clicking on a history item doesn't show the response, only the status code
- [ ] Proper icon for the 'Edit collection' icon
- [ ] Clicking to another sample resource does not automatically clear the tooltip

Mnickii:
## Overview
Closes #3630 

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description


ElinorW:
A few comments

1. History tab is missing a scrollbar 
2. Also, idk if this is part of the history stuff, but when you're on the Adaptive Card tab and you select a history item that has no adaptive card It just puts some grey thing instead of displaying the 'no adaptive card...' message  
3. Missed Regression: previously history items were arranged with the latest at the top, I've noticed that now the latest goes to the bottom
4. Bug: when I select the "delete" icon on the history tab the list disappears  
image

Mnickii:
> A few comments
> 
> 1. History tab is missing a scrollbar
> 2. Also, idk if this is part of the history stuff, but when you're on the Adaptive Card tab and you select a history item that has no adaptive card It just puts some grey thing instead of displaying the 'no adaptive card...' message
>    e.g running GET v1.0/me/todo/lists as a history item
>    image
> 3. Missed Regression: previously history items were arranged with the latest at the top, I've noticed that now the latest goes to the bottom
> 4. Bug: when I select the "delete" icon on the history tab the list disappears then reappears when I choose to cancel
>    image

I've added these to the regressions, will be covered in the next PR

ElinorW (approved):

PR #3663, opened 2025-02-27
fix: remove duplicates in collections

### LINKED ISSUE #3662: Adding an entire resource to collection duplicates previous added resource items, if in the same resource category.
Steps to reproduce:
2. Add admin>people resource to collections
3. Add the whole admin resource
Notice duplication.

Notes: There should be no duplication. There should be a check on if items already exist in collection.

Mnickii:
## Overview

Closes #3662 

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description


ElinorW:
This works, but when I add one endpoint, then later I change my mind and decide to add the whole group, I don't get that option, now the group button just reads "remove from collection" which means I'll have to add everything in the group, one by one...
I think we should allow users to still be able to add a whole resource group to the collection, only that, something that was already there shouldn't appear again

musale (approved):

PR #3810, opened 2025-04-16
feat: add request/response area resizing

### LINKED ISSUE #3778: Can't reduce the size of the top pane (request body etc) anymore
**Describe the bug**
In previous version of Graph Explorer, it was possible to "grab" the separation between the top pane and bottom pane (response preview etc) to change the size on the fly. Very useful when only doing GETs.

**To Reproduce**
1. Go to Graph Explorer
2. Move the mouse between the top and bottom panes
3. Fail to see any space where resizing panes could happen

**Expected behavior**
We should be able to resize (or collapse) the top pane, at least

### LINKED ISSUE #3785: unable to resize the Request Body window in the new Graph Explorer UI
**Describe the bug**
We have an updated Microsoft Graph Explorer UI now (not sure I like it though) but its quite annoying I can no longer resizee the "Request Body" window part. It now takes up more then half of the screen now and makes it difficult to work with the actual results of the graph requests we do.

**To Reproduce**
Go to https://developer.microsoft.com/en-us/graph/graph-explorer
you land on "Request Body" which now takes more then half of the screen and cannot be resized.

**Expected behavior**
use a smaller size for the "Request Body" by default and/or allow to resize that part of the window (as it did before)

 - OS: Windows 11
 - Browser: Edge
 - Version: 135.0.3179.66 (Official build) (64-bit)

ElinorW:
## Overview

fixes #3778 
fixes #3785 

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -89,7 +89,6 @@ configuration:
           - isOpen
           - not:
               and:
-                - isAssignedToSomeone
                 - isLabeled
       then:
         - addLabel:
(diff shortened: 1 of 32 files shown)


Mnickii (approved):
Works great!

PR #3984, opened 2025-10-15
fix: upgrading axios version in tarbell file

lehrt:
## Overview

Local expvariantassignmentsdk-1.0.0.tgz package uses older version of axios. Upgraded to 0.30.0 for compliance purposes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,6 +71,7 @@
                         },
                         "devDependencies": {
                                 "@axe-core/playwright": "4.10.1",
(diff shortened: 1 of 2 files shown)


MIchaelMainer (changes requested):
Looks like plugin-proposal-private-property-in-object has been deprecated. 

https://www.npmjs.com/package/@babel/plugin-proposal-private-property-in-object states that you should use @babel/plugin-transform-private-property-in-object

PR #4008, opened 2026-01-13
fix: prevent createShareLink crash on invalid URL

### LINKED ISSUE #3983: The app crashes when I try to create a POST request
**Describe the bug**
After changing the request method to POST and adding a request URL, the app crashes when I select the Request Body tab.

**To Reproduce**
2. Click on GET and select POST
3. Add an invalid Request URL: htps://graph.microsoft.com/beta/external/connections
4. Select the Request Body tab and see the app crashing

**Expected behavior**
App should not crash

**Screenshots**

Image

Image

**Additional context**

xDiegoNunezx:
Fix #3983

### DIFF
--- a/src/app/views/common/share.ts
+++ b/src/app/views/common/share.ts
@@ -1,5 +1,5 @@
-import { authenticationWrapper } from '../../../modules/authentication';
 import { IQuery } from '../../../types/query-runner';
+import { authenticationWrapper } from '../../../modules/authentication';
 import { encodeHashCharacters } from '../../utils/query-url-sanitization';
 import { parseSampleUrl } from '../../utils/sample-url-generation';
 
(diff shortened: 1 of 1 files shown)


src/app/views/common/share.ts
```diff
@@ -17,8 +17,15 @@ export const createShareLink = (sampleQuery: IQuery, authenticated?: boolean): s
...
 
-  const url = new URL(sampleUrl);
-  const graphUrl = url.origin;
+  let graphUrl = 'https://graph.microsoft.com';
+
+  try {
+    const url = new URL(sampleUrl);
+    graphUrl = url.origin;
+  } catch {
+    // do nothing and use default graph url
```
lehrt: Can we add a dialog that informs the user that they have an invalid origin and recommend they double check for typos? A typo in the 'https' part of the origin could take an annoying moment to find if there's no prompt except 'cannot find a resource with that name'.

PR #4019, opened 2026-03-09
fix: remove blob storage deployment steps from build pipeline

xDiegoNunezx:


azure-pipelines.yml
Highbury1993: Could we remove it from this line?

xDiegoNunezx: Do you mean delete all the DeployProduction stage?

Highbury1993: Yes, seems unnecessary.

xDiegoNunezx: Removed it. That remaining part was in charge of creating a GitHub release with the built artifacts when changes are merged to the master branch.

